PR #2, opened 2023-01-26
Bootstrap apiserver repository

zugao:
This PR boostraps the API Server for this repository. The initial commit has the generic code for an api server from [apiserver-builder-alpha](https://github.com/kubernetes-sigs/apiserver-builder-alpha/).

To do a quick test that the API Server works, initiate a local installation `make local-install` as described in the docs.


glrf (changes requested):
`make local-install` fails with `Error: stat dev/config/aggregated-apiserver.yaml: no such file or directory`

I think you forgot to commit `dev/config`?

zugao:
> `make local-install` fails with `Error: stat dev/config/aggregated-apiserver.yaml: no such file or directory`
> 
> I think you forgot to commit `dev/config`?

I lost it somewhere between the commits :D

Makefile
```diff
@@ -55,76 +53,35 @@ fmt: ## Run go fmt against code.
...
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
+docker-build:
+	env GOOS=$(DOCKER_IMAGE_GOOS) GOARCH=$(DOCKER_IMAGE_GOARCH) \
+		go build -o ${BIN_FILENAME}
```
glrf, line 77: We need to statically link this otherwise the container won't work when built on a machine with glibc

```suggestion
	env CGO_ENABLED=0 GOOS=$(DOCKER_IMAGE_GOOS) GOARCH=$(DOCKER_IMAGE_GOARCH) \
		go build -o ${BIN_FILENAME}
```

dev/config/apiservice.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1.appcat.vshn.io
+  labels:
+    api: appcat
+    apiserver: "true"
+spec:
+  version: v1
+  group: appcat.vshn.io
```
glrf, line 11: So I had to double check this, but this will cause some weird issues when we deploy it.

This `APIService` definition basically tells the k8s API server that we as an aggregated API server are responsible for this api group. Than means if a user requests a resource `foo.appcat.io` the k8s API server will basically just blindly forward the request to us.

That's a problem for `objectbuckets.appcat.vshn.io `. If we install this, we are unable to interact with any objectbuckets. We need to have a unique API group for the types provided by the aggregated API server

pkg/apiserver/appcat/appcat.go
```diff
@@ -0,0 +1,38 @@
+package appcat
+
```
glrf, line 2: General nitpick: I prefer not to unnecessarily put all go code under `pkg`. But it's also fine if we keep that

glrf (changes requested):
Looks fairly good to me overall. But we need to be careful with the API group

dev/config/apiservice.yaml
zugao: so we need to have a different group for our API server. We might call it `appcat.service.vshn.io`.

glrf: Exactly. `appcat.apiservice.vshn.io` or `apiservice.appcat.vshn.io` or `api.appcat.vshn.io` or IDK naming is hard, but it needs to be unique

zugao: Maybe @Kidswiss or @tobru has some input.

pkg/apiserver/appcat/appcat.go
zugao: I know that generally the code should go in `pkg` but it's fine I guess to remove it.

glrf: Well AFAIK people started putting code under `pkg` and the main under `cmd` because they saw kubernetes doing it. But IMO there is no good reason to do that in general if you don't have a very large project with a lot of non Go files.

But again this very nipicky so :shrug:

dev/config/apiservice.yaml
Kidswiss: I'd go with `api.appcat.vshn.io`.

But let's also get tobru's input on that. 👍

zugao: Changed to `api.appcat.vshn.io`

tobru (approved):
I haven't checked the code, but I agree with the naming :+1:

glrf (approved):

PR #6, opened 2023-03-01
Add VSHNPostgresBackup resource

zugao:
## Summary

This PR adds a new managed resource in this API Server.

* Stackgres SGBackups are being sourced from postgres namespaces to customer namespace via the new resource VSHNPostgresBackup.
* Added tests
* Supported operation `get;list;watch`
* Small refactoring

Test locally: 
1) `make local-install`
2) wait until `kubectl get vshnpostgresql` is ready
3) `kubectl get vshnpostgresbackup` 

## Checklist

- [x] Update tests.


apis/vshn/postgres/v1/backup_types.go
```diff
@@ -0,0 +1,145 @@
...
+func init() {
+	apis.SchemeBuilder.Register(&VSHNPostgresBackup{}, &VSHNPostgresBackupList{})
+}
+
+func New() *VSHNPostgresBackup {
+	return &VSHNPostgresBackup{}
+}
+
+// BackupInfo holds necessary data for VSHNPostgresBackup
+type BackupInfo struct {
```
Kidswiss, line 112: The name here seems a bit generic. Or will this also be usable for other backup types in the future? Like Redis?

apiserver/vshn/postgres/multiwatch.go
```diff
@@ -0,0 +1,41 @@
...
+		watcher.Stop()
+	}
+}
+
+// ResultChan aggregates all channels from all watches of this MultiWatch
+func (m *MultiWatcher) ResultChan() <-chan watch.Event {
+	out := make(chan watch.Event)
+	for _, w := range m.watchers {
+		watcher := w
+		go func() {
```
Kidswiss, line 34: This go routine will never return. So we'll end up with a bunch of those stuck every time someone uses the watch command.

We probably want to implement some sort of cancel/stop mechanism here.

dev/local.mk
```diff
@@ -1,12 +1,82 @@
-.PHONY: local-install
+crossplane_sentinel = $(kind_dir)/crossplane_sentinel
```
Kidswiss, line 1: Just as a note, we probably want to move such things into a central kind dev repo in the future.

Kidswiss (changes requested):
Found something about the watcher that we probably should address. Rest LGTM.

apis/vshn/postgres/v1/backup_types.go
zugao: No, definitely not. We know naming is hard so I am open to suggestions

apiserver/vshn/postgres/multiwatch.go
zugao: The watches in `multiwatch` instance should not be stopped when calling the channels as they need to send events continuously. Only when the client stops the "watching" action then the watches underneath. `multiwatch` should be closed. I see it like that.

dev/local.mk
zugao: we need a ticket for that. I will not update the makefile for any new feature.

apiserver/vshn/postgres/multiwatch.go
```diff
@@ -0,0 +1,41 @@
...
+func (m *MultiWatcher) AddWatcher(w watch.Interface) {
+	m.watchers = append(m.watchers, w)
+}
+
+// Stop stops all watches of this MultiWatch
+func (m *MultiWatcher) Stop() {
+	for _, watcher := range m.watchers {
+		watcher.Stop()
+	}
+}
```
glrf, line 27: We should also close the channel of the multiWatcher

apiserver/vshn/postgres/vshnpostgresql.go
```diff
@@ -0,0 +1,48 @@
...
+	for _, v := range unstructuredList.Items {
+		instanceNamespace, exists, err := unstructured.NestedString(v.UnstructuredContent(), v1.Status, v1.InstanceNamespace)
+		if err != nil || !exists {
+			return nil, err
+		}
+		instances[v1.VSHNPostgreSQLName(v.GetName())] = v1.VSHNPostgreSQLNamespace(instanceNamespace)
+	}
+
+	return &instances, nil
+}
```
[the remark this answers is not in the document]
zugao: The list is not implemented for VSHNPostgreSQL hence the reason reusing the dynamic client.

apiserver/vshn/postgres/get.go
```diff
@@ -0,0 +1,48 @@
...
+		if err != nil {
+			resolvedErr := apiserver.ResolveError(sgbackupGroupVersionResource.GroupResource(), err)
+			if apierrors.IsNotFound(resolvedErr) {
+				continue
+			}
+			return nil, err
+		}
+
+		vshnBackup = v1.NewVSHNPostgresBackup(backupInfo, string(k), namespace)
+	}
```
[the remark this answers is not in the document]
zugao: The backup name is composted like that `<db-instance-name>-<timestamp-to-seconds>`. The database instance has already a dynamic part at the end of the chosen name so there's no chance of having two backups with the same name. Ex: `pgsql-app1-prod-sxbmc-2023-02-24-15-05-08`.

apiserver/vshn/postgres/vshnpostgresql.go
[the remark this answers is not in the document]
glrf: Hmm oke.

I'd probably just have added the List then. That's not much work and prevents possible bugs with having to work with unstructured, but :shrug:

apiserver/vshn/postgres/list.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return nil, apiserver.ResolveError(apis.GetGroupResource(v1.Resource), err)
+		}
+		for _, b := range *bis {
+			vb := v1.NewVSHNPostgresBackup(&b, string(k), namespace)
+			if vb != nil {
+				backups.Items = append(backups.Items, *vb)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
zugao: I will resolve this and will discuss that in the other comment.

apiserver/vshn/postgres/multiwatch.go
zugao: ah good catch, absolutely.

apiserver/vshn/postgres/get.go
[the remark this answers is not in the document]
glrf: Hmm I guess that's fine. I'm still not a huge fan as it could still be theoretically possible that there are two SGBackups with the same name and that would lead to a very weird bug, but I'm ok with it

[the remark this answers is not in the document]
zugao: Well, theoretically speaking even if we add the namespace to the sgbackup name we will not have eliminated 100% the chance of 2 backups having the same name no? It will just reduce further the possibility of having 2 with the same name :)

apiserver/vshn/postgres/vshnpostgresql.go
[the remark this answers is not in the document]
zugao: No it isn't but for some reason I agreed with @Kidswiss to move with unstructured here again. @Kidswiss do you think I should add the List implementation in the component?

apiserver/vshn/postgres/get.go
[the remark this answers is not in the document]
glrf: Well namespace-name is unique, so no. But yeah you'd have to consider some other things

apis/vshn/postgres/v1/backup_types.go
Kidswiss: maybe `PgBackupInfo`?

apiserver/vshn/postgres/vshnpostgresql.go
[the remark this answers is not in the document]
Kidswiss: No hard feelings with implementing the list.

apis/vshn/postgres/v1/backup_types.go
```diff
@@ -0,0 +1,145 @@
...
+type VSHNPostgreSQLNamespace string
+
+// +kubebuilder:object:root=true
+
+// VSHNPostgresBackup defines VSHN managed PostgreSQL backups
+type VSHNPostgresBackup struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec VSHNPostgresBackupSpec `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
zugao: Yeah makes sense.

apis/vshn/postgres/v1/backup_types.go
```diff
@@ -0,0 +1,145 @@
...
+// VSHNPostgresBackupList defines a list of VSHNPostgresBackup
+type VSHNPostgresBackupList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []VSHNPostgresBackup `json:"items"`
+}
+
+// VSHNPostgresBackupSpec defines the desired state of VSHNPostgresBackup
+type VSHNPostgresBackupSpec map[string]runtime.RawExtension
```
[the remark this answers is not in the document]
zugao: Moved to a struct. Indeed having some fields in there helps reducing complexity.

apiserver/vshn/postgres/multiwatch.go
zugao: As already discussed @glrf , a single routing and a select statement will not work in this case. I will leave it as is. Thanks for the suggestion.

apis/vshn/postgres/v1/backup_types.go
zugao: I changed to `SGBackupInfo`. I think that's the best way to move forward.

apiserver/vshn/postgres/multiwatch.go
```diff
@@ -0,0 +1,44 @@
...
+	m.watchers = append(m.watchers, w)
+}
+
+// Stop stops all watches of this MultiWatch
+func (m *MultiWatcher) Stop() {
+	for _, watcher := range m.watchers {
+		watcher.Stop()
+	}
+	close(m.eventChan)
+}
```
glrf, line 31: This is nitpicky but I think there is a small chance that this will cause a panic.

1. A proxy goroutine pulls an event form the upstream channel
2. `Stop` is called. All channels are closed
3. The proxy goroutine continues, tries to write to a closed channel, and panics

Not sure if this actually ever happens.

apis/appcat/v1/backup_types.go
```diff
@@ -0,0 +1,146 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
glrf, line 4: Nitpick: I'd rename these files to something like `vshn_postgres_backup_types.go` or something like that?

glrf (approved):

apiserver/vshn/postgres/vshnpostgresql.go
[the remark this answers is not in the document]
zugao: Let me take a bit more time and try to implement the list

apiserver/vshn/postgres/multiwatch.go
zugao: do you expect some locks here?

apis/appcat/v1/backup_types.go
zugao: yeah makes sense

apiserver/vshn/postgres/multiwatch.go
glrf: A WaitGroup might make sense so that we wait for all goroutines to finish before closing the channel?

apiserver/vshn/postgres/vshnpostgresql.go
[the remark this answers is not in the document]
zugao: Implemented the list.

apiserver/vshn/postgres/multiwatch.go
zugao: Just added it, can you check again?

glrf (approved):

PR #16, opened 2023-05-11
Override deletion protection on instance namespace deletion

Kidswiss:
## Summary

* If the instance namespace is deleted, then the controller will disable the deletion protection on the composite

## Checklist

- [x] Update tests.

### DIFF
--- a/config/controller/cluster-role.yaml
+++ b/config/controller/cluster-role.yaml
@@ -19,3 +19,12 @@ rules:
       - objects
     verbs:
       - delete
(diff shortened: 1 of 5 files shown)


Kidswiss:
Not sure if I should already rebase to the common repo branch. What do you think @zugao?

zugao:
@Kidswiss you could rebase. I don't expect to do bit changes

wejdross (approved):
TEst it properly, code looks clean, approving

PR #23, opened 2023-06-01
Add support for multiple buckets in the miniodev environment

Kidswiss:
## Summary

* This adds support to provision multiple buckets in the local minio
* A new comp-func will create a job that in turn creates a new bucket on a pre-provisioned minio
* This now supports multiple buckets in local dev and thus multiple postgresql instances
* The whole provisioning of an instance locally is much faster

There will be other PRs for the component and kindev accordingly.

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/miniodev/objectbucket.go
```diff
@@ -0,0 +1,76 @@
...
+	appcatv1 "github.com/vshn/appcat/apis/v1"
+	"github.com/vshn/appcat/pkg/comp-functions/runtime"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ProvisionMiniobucket will create a bucket in the
+func ProvisionMiniobucket(ctx context.Context, iof *runtime.Runtime) runtime.Result {
+	claim := &appcatv1.ObjectBucket{}
```
glrf, line 16: Maybe make it a bit more clear that this is a function only used on dev instance. Also please finish the docstring :sweat_smile:

glrf (approved):
I'd prefer if this would be behind a feature flag. Just so this won't run in production. But otherwise LOKTM

glrf (approved):

PR #24, opened 2023-06-12
Add SLA reporting tool

Kidswiss:
## Summary

This commit contains a simple SLA reporting tool. It connects to a
Prometheus/Thanos/Mimir instance and queries them to gather the SLA
information.
    
It then creates an ASCIIdoc template and sends it to VSHN's docgen.
There it will be converted to a PDF and sent to an S3 instance.

## Checklist

- [x] Update tests.


cmd/slareport.go
```diff
@@ -0,0 +1,129 @@
...
+		endpointURL,
+		bucket,
+		awsaccesskeyid,
+		awssecretaccesskey,
+	)
+	if err != nil {
+		return err
+	}
+
+	for customer, metrics := range metrics {
```
glrf, line 83: Nitpick: Shadowing `metrics` is IMO a bit confusing

cmd/slareport.go
```diff
@@ -0,0 +1,129 @@
...
+
+		l.Info("Rendering PDF for customer", "customer", customer)
+		parsedDate, err := time.Parse(time.RFC3339, date)
+		if err != nil {
+			return err
+		}
```
glrf, line 89: Nitpick: You could parse the date outside of the loop

pkg/slareport/prom.go
```diff
@@ -0,0 +1,183 @@
...
+
+	rangedQuery := strings.Replace(metricQuery, "{{DURATION}}", timeRange, 2)
+
+	value, warnings, err := client.QueryRange(ctx, rangedQuery, queryRange)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(warnings) != 0 {
+		fmt.Println(warnings)
```
glrf, line 150: Please use the logger and not `fmt`

pkg/slareport/prom.go
```diff
@@ -0,0 +1,183 @@
...
+func getTargetSLA(ctx context.Context, slothID string, client apiv1.API, endDate *time.Time) (float64, error) {
+	query := strings.Replace(slaQuery, "{{SLOTHID}}", slothID, 1)
+
+	res, warns, err := client.Query(ctx, query, *endDate)
+	if err != nil {
+		return 0, err
+	}
+
+	if len(warns) != 0 {
+		fmt.Println(warns)
```
glrf, line 170: Also please use the logger

pkg/slareport/prom.go
```diff
@@ -0,0 +1,183 @@
...
+	samples, ok := res.(model.Vector)
+	if !ok {
+		return 0, fmt.Errorf("expected prometheus query to return a model.Vector, got %T", res)
+	}
+
+	if len(samples) == 0 {
+		return 0, errors.New("no target SLA found in prometheus")
+	}
+
+	return float64(samples[len(samples)-1].Value) * 100, nil
```
glrf, line 182: Nitpick: It's a bit inconsistent that you multiply the measured SLI in the PromQL query, but multiply the SLA in code.

pkg/slareport/prom_test.go
```diff
@@ -0,0 +1,203 @@
...
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+
+		promClientFunc = getDummyPromClient
+		getTargetSLAFunc = getDummySLA
+		getMetricsFunc = func(ctx context.Context, startDate, endDate *time.Time, timeRange string, client apiv1.API) (model.Matrix, error) {
+			return tt.metrics, nil
+		}
```
glrf, line 183: Nitpick: You change global state in the test without reverting it. So this could potentially lead to weird behavior if someone adds an integration test and is confused why the actual `promClientFunc` isn't called

pkg/slareport/s3.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// PDFUploader uploads pdfs to the configured S3 endpoint.
+type PDFUploader struct {
+	client *minio.Client
+	bucket string
+	ctx    context.Context
```
glrf, line 20: Please don't save the context in the struct. You should pass it when calling `Upload`.

pkg/slareport/s3.go
```diff
@@ -0,0 +1,70 @@
...
+
+// PDF is contains all metainformation to upload a PDF to S3.
+type PDF struct {
+	Customer string
+	Date     time.Time
+	PDFData  io.ReadCloser
+}
+
+// Login initializes an S3 client.
+func (p *PDFUploader) Login(ctx context.Context, endpoint, bucket, keyID, secretKey string) error {
```
glrf, line 31: Nitpick: I'd probably make this a constructor function instead of a method, the same way the minio library does it. It's IMO a bit awkward to call `Login` on the Uploader.

glrf (approved):
LGTM overall. Just some nitpicks

cmd/slareport.go
Kidswiss: Naming is hard :D

pkg/slareport/prom.go
Kidswiss: Good catch.

PR #25, opened 2023-06-14
Add Redis Backup

Kidswiss:
## Summary

* Add redis backups based on K8up
* The backup can handle snapshot storage and aof storage or any mix of them
* Each new instance also deploys a k8up schedule as well as a config map that contains the actual backup script to be triggered within the redis container.

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnredis/script/backup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# This script works both with enabled and disabled AOF. It ensures that there
+# are no AOF rewritings happening at the time of backup. If there's a dump file
+# it will also backup that one.
+
+redis_cmd() {
+  redis-cli -a "${REDIS_PASSWORD}" --cacert "${REDIS_TLS_CA_FILE}" --cert "${REDIS_TLS_CERT_FILE}" --key "${REDIS_TLS_KEY_FILE}" --tls $@
```
glrf, line 8:
```suggestion
  redis-cli -a "${REDIS_PASSWORD}" --cacert "${REDIS_TLS_CA_FILE}" --cert "${REDIS_TLS_CERT_FILE}" --key "${REDIS_TLS_KEY_FILE}" --tls "$@"
```

Input from shellcheck (And I think it's right, but I'm terrible at writing shell scripts so maybe it's not :sweat_smile: ) `Double quote array expansions to avoid re-splitting elements`

pkg/comp-functions/functions/vshnredis/backup.go
```diff
@@ -0,0 +1,374 @@
...
+		return runtime.NewFatalErr(ctx, "failed to set composite", err)
+	}
+
+	for _, res := range backupResources {
+		l.Info("running backup step", "name", res.Name)
+		err := res.ResFunc(ctx, comp, iof)
+		if err != nil {
+			return runtime.NewFatalErr(ctx, "error during backup step: "+res.Name, err)
+		}
+	}
```
glrf, line 93: Nitpick/ personal preference (feel free to completely ignore): I never saw the point of such "pipelines". IMO makes it a bit harder to read for saving maybe a few lines of error handling code. I'd just call the functions one after the other.

pkg/comp-functions/functions/vshnredis/backup.go
```diff
@@ -0,0 +1,374 @@
...
+
+	err = createScriptCM(ctx, comp, iof)
+	if err != nil {
+		return runtime.NewFatalErr(ctx, "cannot add backup cm", err)
+	}
```
glrf, line 98: Why is that one function special and isn't part of the `backupResources` seems to be the same thing?

pkg/comp-functions/functions/vshnredis/backup.go
```diff
@@ -0,0 +1,374 @@
...
+	volumeMounts := []corev1.VolumeMount{
+		{
+			Name:      backupScriptCMName,
+			MountPath: "/scripts",
+		},
+	}
+
+	masterMap["extraVolumeMounts"] = volumeMounts
+
+	valueMap["master"] = masterMap
```
glrf, line 356: All of these unchecked type casts and map access have the potential to panic. So as far as I can tell (potentially harmless looking) changes to the component can result in a panic here.

I can see that these should be fine in practice, but I'd still prefer this to be a bit more defensive and to check if the typecasts actually work and that the fields aren't nil

pkg/comp-functions/functions/vshnredis/commong_test.go
glrf, line 1: Probably a typo or what does the `g` stand for? :sweat_smile:

pkg/comp-functions/functions/vshnredis/script/backup.sh
Kidswiss: Still seems to work as expected 👍

pkg/comp-functions/functions/vshnredis/backup.go
Kidswiss: Yeah, maybe I'll change it. That I already missed a function speaks for itself here 🙈

glrf (approved):

PR #26, opened 2023-06-20
Add Redis Backup Listing

Kidswiss:
## Summary

It's now possible to list the backups of Redis instances via the API server.

## Checklist

- [x] Update tests.


pkg/apiserver/vshn/redis/backup.go
```diff
@@ -0,0 +1,65 @@
...
+	return &appcatv1.VSHNRedisBackup{}
+}
+
+func (v vshnRedisBackupStorage) Destroy() {}
+
+func (v *vshnRedisBackupStorage) NamespaceScoped() bool {
+	return true
+}
+
+func trimStringLenght(in string) string {
```
glrf, line 53: Typo

```suggestion
func trimStringLength(in string) string {
```

pkg/apiserver/vshn/redis/get.go
```diff
@@ -0,0 +1,54 @@
...
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return nil, err
+		}
+
+		redisSnap = &appcatv1.VSHNRedisBackup{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: trimStringLenght(*snap.Spec.ID),
+			},
```
glrf, line 44: I see two issues with this:

* Why do you take the beginning of the ID instead of the name of the backup? They are the same right? If they aren't things will be weird. I.e. you list backups and then try to get one based on the name but you won't be able to as the name is different
* You should set more then just the name in the object meta. At least the namespace should be set. But generally we expect `uid`, `resourceVersion`, `generation`, and `creationTimestamp` to be set. Some tooling might expect this to be present and things might fail in very unexpected ways. I'd propose to just copy the complete metadata from the snapshot and update the namespace

Also more of a k8up issue: How likely is it that two snapshots have the same name? That could result in an issue...

wejdross (approved):
code is clean
accidently, but tested it on lab today while debugging other issue with You
approve from my side

pkg/apiserver/vshn/redis/list.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+
+		for _, snap := range snapshots.Items {
+
+			redisSnapshots.Items = append(redisSnapshots.Items, appcatv1.VSHNRedisBackup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: trimStringLenght(*snap.Spec.ID),
+				},
```
glrf, line 45: Same input as for the `Get`. We need to be careful with metadata.

pkg/apiserver/vshn/redis/get.go
```diff
@@ -0,0 +1,54 @@
...
+			Status: appcatv1.VSHNRedisBackupStatus{
+				ID:       *snap.Spec.ID,
+				Date:     *snap.Spec.Date,
```
glrf, line 47: Bit of a nitpick, but given that these two fields are pointers, they might be empty. And de-referencing them can lead to a panic. I'd prefer to be defensive and check for `nil`.

pkg/apiserver/vshn/redis/watch.go
```diff
@@ -0,0 +1,74 @@
...
+			if value.Status.InstanceNamespace == backupInfo.GetNamespace() {
+				db = value.GetName()
+			}
+		}
+
+		if db == "" {
+			return in, false
+		}
+
+		in.Object = &appcatv1.VSHNRedisBackup{
```
glrf, line 60: You're not setting any metadata at all. I'm impressed this works at all :sweat_smile: 

Same input as everywhere else: We should set metadata properly

pkg/apiserver/vshn/redis/watch.go
```diff
@@ -0,0 +1,74 @@
...
+			Status: v1.VSHNRedisBackupStatus{
+				ID:       *backupInfo.Spec.ID,
+				Date:     *backupInfo.Spec.Date,
+				Instance: db,
+			},
+		}
+
+		if in.Object.(*appcatv1.VSHNRedisBackup) == nil {
+			return in, false
+		}
```
glrf, line 70:
```suggestion

```

Don't need this

glrf (changes requested):

pkg/apiserver/vshn/redis/get.go
Kidswiss: > Also more of a k8up issue: How likely is it that two snapshots have the same name? That could result in an issue...

This should really not happen at all, as a clash in the snapshot IDs would completely break a restic repository.

Kidswiss: > You should set more then just the name in the object meta. At least the namespace should be set. But generally we expect uid, resourceVersion, generation, and creationTimestamp to be set. Some tooling might expect this to be present and things might fail in very unexpected ways. I'd propose to just copy the complete metadata from the snapshot and update the namespace

Shoulddn't I create a new UID in this case though? This feels like it could be problematic as well...

glrf: > Shoulddn't I create a new UID in this case though? This feels like it could be problematic as well...

I remember we had a longer discussion about this for the initial control API and found after longer digging that UIDs only need to be unique per Kind. So reusing the UID from the snapshots solves our problem of unique IDs without having to generate new once and having to track more state.

Also OpenShift does the exact same thing for `Projects` and that seemed to have worked fine until now :)

glrf: > This should really not happen at all, as a clash in the snapshot IDs would completely break a restic repository.

But as far as I can tell we're only using the first few characters of the snapshot ID as the name. So there could be clashes. Not likely so not a huge issue but still

Kidswiss: > > This should really not happen at all, as a clash in the snapshot IDs would completely break a restic repository.
> 
> But as far as I can tell we're only using the first few characters of the snapshot ID as the name. So there could be clashes. Not likely so not a huge issue but still

True, but printing the whole ID is waaay too long and we currently don't have another field we could use as a name. So for now let's stick with this.

Kidswiss: > > Shoulddn't I create a new UID in this case though? This feels like it could be problematic as well...
> 
> I remember we had a longer discussion about this for the initial control API and found after longer digging that UIDs only need to be unique per Kind. So reusing the UID from the snapshots solves our problem of unique IDs without having to generate new once and having to track more state.
> 
> Also OpenShift does the exact same thing for `Projects` and that seemed to have worked fine until now :)

Yeah makes it easier. I did a some test with generating a stable UID based on the name/id of the snapshot, but if re-using the UID is no issue I'll just do that.

glrf (approved):

PR #29, opened 2023-06-27
Add redis maintenance comp function

zugao:
## Summary

* This PR adds the cronjob for Redis maintenance

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/common/common.go
```diff
@@ -29,7 +29,7 @@ func loadRuntimeFromFile(t assert.TestingT, file string) *runtime.Runtime {
 	return funcIO
 }
 
-func getFunctionIo(funcIO *runtime.Runtime) xfnv1alpha1.FunctionIO {
+func GetFunctionIo(funcIO *runtime.Runtime) xfnv1alpha1.FunctionIO {
 	field := reflect.ValueOf(funcIO).Elem().FieldByName("io")
 	return reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem().Interface().(xfnv1alpha1.FunctionIO)
 }
```
Kidswiss, line 35: Shouldn't this file be `common_test.go`?

Kidswiss (changes requested):
Two smaller things, the rest LGTM

EDIT: Also fix the tests please :)

pkg/comp-functions/functions/common/maintenance.go
```diff
@@ -0,0 +1,206 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Maintenance contains data for maintenance k8s resource generation
+type Maintenance struct {
+	InstanceNamespace string
+	MainRole          string
+	Service           string
+	Resource          client.Object
+	Ctx               context.Context
```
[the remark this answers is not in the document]
glrf: I think the underlying issue here is. Why do you create a struct just to call a single method on it? 

Why isn't `Run` a function that takes the necessary input as parameters?

pkg/comp-functions/functions/common/common.go
Kidswiss: From what we've just discussed offline:

Maybe put it in a separate package containing the word test? I don't like test functions to clutter the public API...

cmd/grpc.go
```diff
@@ -100,6 +100,10 @@ var images = map[string][]runtime.Transform{
 			Name:          "backup",
 			TransformFunc: vshnredis.AddBackup,
 		},
+		{
+			Name:          "backup",
```
glrf, line 104:
```suggestion
			Name:          "maintenance",
```

pkg/comp-functions/functions/common/maintenance.go
```diff
@@ -0,0 +1,206 @@
...
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Maintenance contains data for maintenance k8s resource generation
+type Maintenance struct {
+	InstanceNamespace string
+	MainRole          string
+	Service           string
+	Resource          client.Object
```
glrf, line 23: Could you add some kind of docstring explaining what all these fields are? Right now I'd have to read the code to understand what all these mean.

pkg/comp-functions/functions/common/maintenance.go
```diff
@@ -0,0 +1,206 @@
...
+	Ctx               context.Context
+	Iof               *runtime.Runtime
+	Schedule          vshnv1.VSHNDBaaSMaintenanceScheduleSpec
+	PolicyRules       []rbacv1.PolicyRule
+	ExtraEnvs         []corev1.EnvVar
+	ExtraResFuncs     []ExtraResourceFunctions
```
glrf, line 29: Nitpick: Not sure if I'm a fan of how the `ExtraResFuncs` work currently, because they can easily be abused to do pretty much anything..

pkg/comp-functions/functions/common/common.go
zugao: Since test methods are not exported between packages I had to come up with a solution which I agree is not the best. I created another package `commontest` where I put the common functions for testing.

pkg/comp-functions/functions/common/maintenance.go
zugao: Yup could catch, I will add docs all over the `maintenance.go`.

pkg/comp-functions/functions/common/maintenance.go
zugao: Do you have an idea how I could create extra resources which I need for postgres? I could also leave it in the postgres function but I would prefer to pass it in here somehow.

glrf: Maybe instead of providing arbitrary functions it could only take a `[]client.Object`?

zugao: Check it now

pkg/comp-functions/functions/common/maintenance.go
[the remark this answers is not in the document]
Kidswiss: > Could you call it something other than run. common.Run doesn't really sound like maintenance..

Same from my side.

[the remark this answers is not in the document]
zugao: As already discussed I changed the whole thing to a builder.

Kidswiss (approved):

glrf (approved):
LGTM

PR #32, opened 2023-07-04
Update postres backup table

zugao:
## Summary

* After some research it turns out that the _end time_ of a postgres backup is the relevant timestamp that we have to show to the customer. All the data is saved and available up to this timestamp.

## Checklist


TheBigLee (approved):
LGTM

PR #33, opened 2023-07-04
Add support for restoring redis backups

TheBigLee:
## Summary

* This PR adds a new composite function to configure restores for redis.

## Checklist

- [x] Update tests.


crds/vshn.appcat.vshn.io_vshnredis.yaml
```diff
@@ -78,6 +78,16 @@ spec:
                           pattern: ^([0-1]?[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])$
                           type: string
                       type: object
+                    restore:
+                      description: Rewstore contains settings to control the restore of an instance.
```
glrf, line 82: Looks like a typo. But that should be auto generated and I don't see the typo in the comment. Please double check and regenerate the CRD.

zugao, line 82: There's a typo here. Regenerate the CRDs.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
```diff
@@ -0,0 +1,25 @@
...
+
+source_namespace=$(kubectl -n "${CLAIM_NAMESPACE}" get vshnredis "${CLAIM_NAME}" -ojson | jq -r '.status.instanceNamespace')
+
+echo "copy secret"
+
+access_key=$(kubectl -n "${source_namespace}" get secret backup-bucket-credentials -o template='{{ .data.AWS_ACCESS_KEY_ID | base64decode }}')
+secret_key=$(kubectl -n "${source_namespace}" get secret backup-bucket-credentials -o template='{{ .data.AWS_SECRET_ACCESS_KEY | base64decode }}')
+restic_password=$(kubectl -n "${source_namespace}" get secret k8up-repository-password -o template='{{ .data.password | base64decode }}')
+restic_repository=$(kubectl -n "${source_namespace}" get snapshots.k8up.io "${BACKUP_NAME}" -o jsonpath='{.spec.repository}')
+backup_path=$(kubectl -n "${source_namespace}" get snapshots.k8up.io "${BACKUP_NAME}" -o jsonpath='{.spec.paths[0]}')
```
zugao, line 13: What happens if the backup failed and the customer tries to use the backup to restore the instance?

pkg/comp-functions/functions/vshnredis/restore.go
```diff
@@ -0,0 +1,314 @@
...
+		return runtime.NewFatalErr(ctx, "Can't restore from backup", err)
+	}
+
+	log.Info("Cleanup restore")
+
+	err = cleanUpJob(ctx, comp, iof)
+	if err != nil {
+		return runtime.NewFatalErr(ctx, "Can't cleanup after restore", err)
+	}
+
```
glrf, line 72: Tiny Nitpick: Naming of these helper functions could be a bit better. Something like `addPrepareRestoreJob` etc.

Right now while reading this I thought: "Ok we're first preparing the restore job, then we're deploying the restore job and then were cleaning up the job", which didn't really make sense.

Also the error messages are a bit misleading too. It's not that we "Can't restore from backup", we can't deploy the restore job.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
```diff
@@ -0,0 +1,25 @@
...
+backup_name=$(kubectl -n "${source_namespace}" get snapshots.k8up.io "${BACKUP_NAME}" -o jsonpath='{.spec.id}')
+num_replicas=$(kubectl -n "${TARGET_NAMESPACE}" get statefulset redis-master -o jsonpath='{.spec.replicas}')
+kubectl -n "${TARGET_NAMESPACE}" create secret generic "restore-credentials-${BACKUP_NAME}" --from-literal AWS_ACCESS_KEY_ID="${access_key}" --from-literal AWS_SECRET_ACCESS_KEY="${secret_key}" --from-literal RESTIC_PASSWORD="${restic_password}" --from-literal RESTIC_REPOSITORY="${restic_repository}" --from-literal BACKUP_PATH="${backup_path}" --from-literal BACKUP_NAME="${backup_name}"
+kubectl create secret generic "statefulset-replicas-${CLAIM_NAME}-${BACKUP_NAME}" --from-literal NUM_REPLICAS="${num_replicas}"
+echo "scaling down redis"
+
+until kubectl -n "${TARGET_NAMESPACE}" get statefulset redis-master
```
zugao, line 20: why this loop exists?

glrf (approved):
LGTM overall

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euo pipefail
+
+echo "Wait for restore to complete"
+
+until [ "$(kubectl -n "${TARGET_NAMESPACE}" get job "${RESTORE_JOB_NAME}" -o jsonpath='{.status.succeeded}')" -eq 1 ];
```
zugao, line 7: This will not go into infinite loop in case the restore job failed right?

zugao (changes requested):
Some things to check.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
TheBigLee: Because the job might be deployed before the statefulset exists.

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
TheBigLee: It actually will. Not sure if we should handle this differently.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
Kidswiss: They probably either end up with a corrupted or an empty instance after the restore.

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
zugao: > Maybe it makes sense to use `kubectl wait --for=condition=succeeded --timeout=300s` or something?

This will work if the resource is already created by the time kubectl wait is issued. If that's the case then we should implement this solution.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
zugao: I would try it in a manual test and see if we handle this better than having something that we don't know what might happen.

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
TheBigLee: > This will work if the resource is already created by the time kubectl wait is issued. If that's the case then we should implement this solution.

We can't guarantee that. It really depends if the provider-helm is faster in deploying the release then the composite function is in deploying the job.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
TheBigLee: Does a failed backup actually create a usable snapshot? If yes, shouldn't this be handled by k8up?

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
zugao: I would just add a timeout with bash in this case.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
glrf: As far as I can tell k8up snapshots have no info whether they failed, so I'd agree and expect that all k8up snapshots are valid and can be restored.

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
glrf: It's honestly not clear to me how we should handle a restore failure.  Just scaling back up, even through we did not restore isn't really the correct behavior.

Maybe we'd need some way to bubble up the failure information to the user.

pkg/comp-functions/functions/vshnredis/script/prepRestore.sh
Kidswiss: There are two ways a backup could go wrong:

- the script as a whole fails and exits with != 0 -> then k8up notices that. however the underlying tool `restic` still creates a snapshot, as it doesn't know anything about the failure, it just takes everything it gets from stdin. It should be possible to teach K8up to remove the damaged snapshot.
- the script delivers k8up with garbage data and exits with 0 -> then k8up has no way to check if the data it received is actually usable

The first case should be easy to fix.

However the second case would probably be very hard to verify without actual restore tests.

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
Kidswiss: We could maybe add a "timeout" to this script with something like:
```
counter=0
until [ $counter -eq 300 ] ||  kubectl -n ...
do
  sleep 1
  $((counter++))
done
```

This should exit after 5 minutes regardless of the object's status.

TheBigLee: Good idea, added a counter variable and adjusted the `until` loop accordingly

TheBigLee: I have added the counter and adjusted the scripts a bit to not "leak" command outputs where it's not useful and made the loops more robust as well.

glrf (approved):

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
Kidswiss: Should be fine for now. But I guess we sooner or later run into instances that won't complete the restore within 5 minutes due to their size.

TheBigLee: Hm didn't consider that part. I'll think about this a little bit longer and try to come up with a more resilient approach, as the 5 minute limit for restore is problematic.

TheBigLee: I refactored that part a bit. Please check again if that solution is acceptable.

Kidswiss (changes requested):
The completed event should probably not be a warning :)

But it's a nice touch, we should also document this, so customers know if/when a restore failed or completed.

glrf (approved):
LGTM. Emitting events is a great idea. 

However, please make sure that you add the permissions to create events in the component. I think the current roles can't.

pkg/comp-functions/functions/vshnredis/script/cleanupRestore.sh
```diff
@@ -4,18 +4,61 @@ set -euo pipefail
...
+fi
+
+if [[ $(kubectl -n "${TARGET_NAMESPACE}" get job "${RESTORE_JOB_NAME}" -o jsonpath='{.status.succeeded}' 2> /dev/null) -eq 1 ]]; then
+  kubectl apply -f - <<EOF
+  apiVersion: v1
+  kind: Event
+  metadata:
+    name: "${DEST_CLAIM_NAME}-restore-completed"
+    namespace: "${CLAIM_NAMESPACE}"
+  type: Warning
```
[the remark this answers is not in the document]
TheBigLee: Fixed

PR #34, opened 2023-07-06
Update target SLA in SLA Reporter

zugao:
## Summary

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/slareport/prom.go
+++ b/pkg/slareport/prom.go
@@ -18,7 +18,7 @@ var (
 	metricQuery = `(1 - max without(prometheus_replica) (sum_over_time(slo:sli_error:ratio_rate5m{sloth_service=~"appcat-.+"}[{{DURATION}}])
 / ignoring (sloth_window) count_over_time(slo:sli_error:ratio_rate5m{sloth_service=~"appcat-.+"}[{{DURATION}}])
 ) >= 0) * 100`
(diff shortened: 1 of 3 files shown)


pkg/slareport/prom.go
```diff
@@ -18,7 +18,7 @@ var (
 	metricQuery = `(1 - max without(prometheus_replica) (sum_over_time(slo:sli_error:ratio_rate5m{sloth_service=~"appcat-.+"}[{{DURATION}}])
 / ignoring (sloth_window) count_over_time(slo:sli_error:ratio_rate5m{sloth_service=~"appcat-.+"}[{{DURATION}}])
 ) >= 0) * 100`
-	slaQuery         = `100*slo:objective:ratio{sloth_id="{{SLOTHID}}"}`
+	slaQuery         = `sla:objective:ratio{sloth_id="{{SLOTHID}}"}`
```
Kidswiss, line 21: You did not add the `sloth_id` to your recording rule. This metric will only have a `service` label.

Kidswiss (changes requested):
Have you tested this? It looks like the labels in the query don't match up with what you specified in the prometheus rule.

zugao:
I updated the label in the component. Now it should match.

Kidswiss (approved):

PR #35, opened 2023-07-07
Don't return error on deletion

Kidswiss:
## Summary

Deletion should not return an error. It looks like the kube-controller calls the deletion function when a namespace is deleted.

## Checklist

- [x] Update tests.


zugao (approved):
LGMT

PR #37, opened 2023-07-11
Add user configurable extensions

Kidswiss:
## Summary

* expose a list of extensions that can be configured by the user
* tested extensions:
  * postgis
  * timescaledb
  * ltree 

## Checklist

- [x] Update tests.


apis/stackgres/v1/sgpostresconfig.go
```diff
@@ -0,0 +1,49 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// This CRD isn't autogenerated, as the generator from the CRD somewhat falls apart with this one.
+// As it only has two fields it should not be too much trouble.
```
glrf, line 8: I think we should be able to make this work, but I also struggle to do so right now.  I think this is fine for now, but maybe we should think about investing some time to un-YOLO-fy this CRD generation in the future.

pkg/comp-functions/functions/vshn-postgres-func/extensions.go
```diff
@@ -0,0 +1,96 @@
...
+		return runtime.NewFatalErr(ctx, "not able to get cluster", err)
+	}
+
+	finalExtensions := []stackgresv1.SGClusterSpecPostgresExtensionsItem{}
+
+	for _, ext := range extensionsMap {
+		finalExtensions = append(finalExtensions, ext)
+	}
+
+	cluster.Spec.Postgres.Extensions = &finalExtensions
```
glrf, line 64: The order when iterating over a map is not fixed. So this `extensions` array will have a different order every time.  Did you check that we do not introduce unnecessary updates of the cluster spec every reconcile loop?

apis/stackgres/v1/sgpostresconfig.go
Kidswiss: Yeah we'll have to look into this in the future.

pkg/comp-functions/functions/vshn-postgres-func/extensions.go
Kidswiss: Good point, I did not check that. But I think with a sort before we apply it, it should be a non-issue.

Kidswiss:
Woops, just realized that I forgot about some tests 🙈 Will add them.

zugao (approved):
One thing rest looks good

glrf (approved):

PR #41, opened 2023-07-13
Improve logging

Kidswiss:
## Summary

* Remove all object prints from the debug logs.
* Add metatada to the logger before running the transformation
* Ability to print whole functionIO to the events

If the GRPC server is in devmode, it will save the funcIOs of each function to configMaps in the `default` namespace. It will always keep two configMaps per composite and function. One contains the previous and one the current state. For example for VSHNPostgreSQL this could result in up to four configMaps; two for the postgresql function and two for the miniodev function (which provides the local S3 bucket).

These configMaps can be used to see the state of the whole FuncIO and should help with local debugging. Additionally the GRPC server now prints a diff of those configMaps on each reconcile.

The configMaps and the diffs will only rotate on `.spec` changes on the claim/composite, it uses the generation to track current and previous states. This way we can track the "logical" diff of a change, even if it needs multiple reconcile loops to be applied fully.

## Checklist

- [x] Update tests.


zugao:
As already discussed, for `prepareFuncLogger()` is a great idea but we should only get the composite name as the other 2 somehow are redundant. Meanwhile `appendFuncIOToMessages()` does not add in my opinion much value, with this approach we move out the function io from logs to events. It helps cleaning out the logs but it will be more difficult to connect the logs to the function io later on. Anyway for debugging purposes it should be fine though we need to find a better approach like having only the diff?

Kidswiss:
I'm looking into the diff idea.

pkg/comp-functions/functions/helper/funcio.go
```diff
@@ -0,0 +1,236 @@
...
+	unstructured.RemoveNestedField(obj, "spec", "atProvider", "manifest", "metadata", "managedFields")
+
+	res.Raw, err = json.Marshal(&obj)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func createOrUpdateObject(ctx context.Context, configMap *corev1.ConfigMap, kubeClient client.Client) error {
```
glrf, line 169: Nitpick: Why don't you use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

zugao (approved):
Seems good so far haven't tested it though.

pkg/comp-functions/functions/helper/funcio.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	diff := difflib.UnifiedDiff{
+		A:        difflib.SplitLines(prevPretty.String()),
+		B:        difflib.SplitLines(currentPretty.String()),
+		FromFile: "Previous",
+		ToFile:   "Current",
+		Context:  3,
+	}
+	text, _ := difflib.GetUnifiedDiffString(diff)
```
glrf, line 232: Nitpick: You shouldn't ignore the error, even if you don't ever expect one

pkg/comp-functions/functions/helper/funcio.go
```diff
@@ -0,0 +1,236 @@
...
+// composite name as the name.
+func SaveFuncIOToConfigMap(funcName string) func(context.Context, *runtime.Runtime) runtime.Result {
+	return func(ctx context.Context, iof *runtime.Runtime) runtime.Result {
+
+		kubeClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{
+			Scheme: pkg.SetupScheme(),
+		})
+		if err != nil {
+			return runtime.NewFatalErr(ctx, "cannot create kube client", err)
+		}
```
glrf, line 41: Why do you create a new kube client in every reconciliation. Wouldn't it be smarter to pass it as an argument? That would also make it testable

glrf (approved):
LGTM overall

pkg/comp-functions/functions/helper/funcio.go
Kidswiss: Oh, I completely forgot, that this existed.

PR #42, opened 2023-07-18
Fix test race condition

zugao:
## Summary

* This small change should resolve the race condition in `GivenMaintenanceTooLong_ThenExpectNoRepack` test

## Checklist


Kidswiss:
There's a typo in the title.

Could you also please elaborate how this should fix the race condition? Did it run into a timeout because it waited too long?

zugao:
> There's a typo in the title.
> 
> Could you also please elaborate how this should fix the race condition? Did it run into a timeout because it waited too long?

The problem can occur only in a test. Apparently the updates to the fake client are so fast that 1 second is too much for the test to go into timeout therefore I set it to a millisecond which will trigger right away the timout.

Kidswiss (approved):

PR #43, opened 2023-07-19
Fix edge cases in PostgreSQL maintenance

glrf:
This commit fixes two separate issue with the PostreSQL maintenance script. Both are related to waiting for the upgrade to complete.

* Fixes a panic in case we receive an event with an SGDbOps without any conditions
* Fixes an issue where the script will run repack prematurely, in case there is an event on a different, completed, SGDbOps in the namespace.

The second point is fairly theoretical, but the second one can be reproduced locally, even though it does not seem to occur in production. Probably a timing issue.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/maintenance/postgresql.go
+++ b/pkg/maintenance/postgresql.go
@@ -87,13 +87,13 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
 	currentVersion := sgclusters.Items[0].Spec.Postgres.Version
 
 	p.log.Info("Checking for upgrades...")
-	err = p.upgradeVersion(currentVersion, clusterName)
+	op, err := p.upgradeVersion(currentVersion, clusterName)
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/maintenance/postgresql.go
```diff
@@ -168,9 +168,12 @@ func (p *PostgreSQL) waitForUpgrade(ctx context.Context) error {
 			switch event.Type {
 			case watch.Modified:
 				ops, _ := event.Object.(*stackgresv1.SGDbOps)
+				if ops.Status.Conditions == nil || ops.Spec.Op != string(op) {
```
zugao, line 171: I think there's no status whatsoever for non recycled CRD. Therefore you might want to check `status` for nil?

glrf: Yeah but the status isn't a pointer :)

zugao: Then no problem :D

zugao (approved):
LGTM

PR #44, opened 2023-07-19
Add option to configure multiple PostgreSQL instances

glrf:
## Summary

This PR enables service users to configure multi-instance Postgres clusters as well as allows them to choose the [replication mode](https://stackgres.io/doc/1.4/reference/crd/sgcluster/#replication). We don't expose  `role` or `groups` of stackgres as it introduces complexity and doesn't really provide any benefits for clusters up to three nodes that we support.

Further this PR removes two possible footguns

* We automatically set the replication mode to `async` if the user only requests a single instance. Stackgres will simply fail if one chooses `sync` or `strict-sync` with only one replica
* We automatically set the `syncInstances` to `instances - 1 `.  There is really no point in setting this to anything else for clusters up to three nodes and misconfiguring this can lead to fail-over issues.

## Checklist

- [x] Update tests.


Kidswiss (approved):
LGTM

One thing that just came to my mind: we're reducing the memory for the cluster-controller container. Did you notice any issues regarding that? I assume the cluster controller probably has a bit more load now that it has to manage more nodes.

glrf:
> One thing that just came to my mind: we're reducing the memory for the cluster-controller container. Did you notice any issues regarding that? I assume the cluster controller probably has a bit more load now that it has to manage more nodes.

I didn't notice any issues, but good input. I'll double check that it doesn't use too much memory before deploying this.

PR #46, opened 2023-07-26
Add external connection IPs to connection secret if service type is set to LoadBalancer for PostgreSQL Managed by VSHN

wejdross:
## Summary

* Added new field under spec.parameters.network called ServiceType, if it's set to LoadBalancer, while defaulting to ClusterIP, comp-function starts and add 2 ip addresses to connection secret allowing to access cluster externally without port-forwarding
* tested with metallb and kind cluster 
* it must spawn in-cluster k8s client and check for services in specific namespace to grab IP addresses

## Checklist

- [-] Update tests.


pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -0,0 +1,100 @@
...
+func AddLoadBalancerIPToConnectionDetails(ctx context.Context, iof *runtime.Runtime) runtime.Result {
+	log := controllerruntime.LoggerFrom(ctx)
+
+	comp := &vshnv1.VSHNPostgreSQL{}
+	err := iof.Desired.GetComposite(ctx, comp)
+	if err != nil {
+		return runtime.NewFatalErr(ctx, "Cannot get composite from function io", err)
+	}
+	// Wait for the next reconciliation in case instance namespace is missing
+	if comp.Status.InstanceNamespace == "" {
```
Kidswiss, line 28: You can use `getInstanceNamespace()` here instead.

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+
+	return runtime.NewNormal()
+}
+
+func getLoadBalancerIPs(instanceNamespace string) (primary string, replicas string, err error) {
+	// Build the Kubernetes config
+	// stackgres just creates services in the same namespace as the instance
+	// therefore I need to manually check IP addresses and then return them
+	config, err := rest.InClusterConfig()
```
Kidswiss, line 65: I'm a bit torn with this. On one hand it's easier to get the values you want with this approach, on the other hand it breaks crossplane's concepts how to interact with the kubernetes objects.

We can shoot ourselves in the foot really badly if we start to rely on the kubeclient directly. Crossplane themselves advise us not to do that:

> Don’t bypass providers. Composition Functions are designed to tell Crossplane how to orchestrate managed resources - not to directly orchestrate external systems.

https://docs.crossplane.io/master/concepts/composition-functions/

So even though it's a bit more complicated I would do it via `provider-kubernetes` to get these values here.

Kidswiss (changes requested):
We should be really careful about using the kubeclient directly. Crossplane themselves advice us to not do that. See my comment for more details.

wejdross:
@Kidswiss I converted it into draft, please take a look, I'll write some tests in meanwhile, so far I tested it manually and it works pretty well locally

Kidswiss (changes requested):
I've added a comment on how to properly instruct provider-kubernetes to observe an object.

I guess if we keep using this more and more we could add that as a function to the runtime library.

pkg/comp-functions/runtime/desired.go
```diff
@@ -100,6 +100,11 @@ func (d *DesiredResources) PutIntoObject(ctx context.Context, o client.Object, k
 			References: refs,
 		},
 	}
+
+	if annotations["appcat.io/observe-only"] == "true" {
+		ko.Spec.ManagementPolicy = "Observe"
```
Kidswiss, line 105:
```suggestion
		ko.Spec.ManagementPolicy = xkube.Observe
```

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -0,0 +1,93 @@
...
+	serviceToApply := &v1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      comp.GetName(),
+			Namespace: getInstanceNamespace(comp),
+			Annotations: map[string]string{
+				"appcat.io/observe-only": "true",
```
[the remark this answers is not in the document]
Kidswiss: I've been thinking a bit more about this and maybe it makes more sense to have a specific function for this?

Something like `iof.Desired.ObserveObject()`. This would make the API a lot clearer and not hide this functionality behind an arbitrary annotation.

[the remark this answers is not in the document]
wejdross: I'd even say ObserveOnlyObject, it makes sense to introduce such function, because we are going to use it more and more

pkg/comp-functions/runtime/desired.go
```diff
@@ -69,8 +69,16 @@ func (d *DesiredResources) GetFromObject(ctx context.Context, o client.Object, k
 	return d.fromKubeObject(ctx, ko, o)
 }
 
-// PutIntoObject adds or updates the desired resource into its kube object
```
Kidswiss, line 72: Please don't remove this comment.

pkg/comp-functions/runtime/desired.go
```diff
@@ -69,8 +69,16 @@ func (d *DesiredResources) GetFromObject(ctx context.Context, o client.Object, k
...
 func (d *DesiredResources) PutIntoObject(ctx context.Context, o client.Object, kon string, refs ...xkube.Reference) error {
+	return d.putIntoObject(ctx, false, o, kon, refs...)
+}
+
```
Kidswiss, line 75: This one should also have a comment.

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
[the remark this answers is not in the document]
Kidswiss: Looks good.

cmd/grpc.go
```diff
@@ -25,12 +25,14 @@ func init() {
...
 }
 
 var (
-	Network     = "unix"
-	AddressFlag = "@crossplane/fn/default.sock"
-	DevMode     = false
+	Network                   = "unix"
+	AddressFlag               = "@crossplane/fn/default.sock"
+	DevMode                   = false
+	ExternalConnectionEnabled = false
```
Kidswiss, line 35: I wouldn't pass this as a flag.

We usually have these configurations in the composition itself in the config: https://github.com/vshn/component-appcat/blob/master/component/vshn_postgres.jsonnet#L849

They are dynamic and can even be set per composition, this makes them a lot more flexible. You can access the configs via `iof.Config` in the compisition function.

cmd/grpc.go
```diff
@@ -137,6 +139,12 @@ func (s *server) RunFunction(ctx context.Context, in *pb.RunFunctionRequest) (*p
 	if err != nil {
 		return nil, status.Errorf(codes.Aborted, "cannot enable devMode: %s", err)
 	}
+	if ExternalConnectionEnabled {
```
Kidswiss, line 142: Instead of conditionally adding this, just add it to the list like all other functions.

Then the first thing you do in the function is checking if the loadbalancer config is enabled or not and either just return or run the function.

Kidswiss:
@wejdross please re-request reviews when I should have another look. Then it pops up in https://github.com/pulls/review-requested. Then I should hopefully not miss it :)

Having a look now.

README.md
```diff
@@ -205,17 +205,22 @@ To add a new function to PostgreSQL by VSHN:
 entrypoint to start working with gRPC server is to run:
 ```
 go run main.go --log-level 1 start grpc --network tcp --socket ':9547' --devmode
+
 ```
 
 This will start the GRPC server listening on a TCP port. Afterward you can configure the composition to use this connection:
 
 ```yaml
+### Macos:
```
Kidswiss, line 214: Maybe we can re-use the readme snipped in line 99 here again?
```
HOSTIP=$(docker inspect kindev-control-plane | jq '.[0].NetworkSettings.Networks.kind.Gateway')
# HOSTIP=host.docker.internal # On some docker distributions
# HOSTIP=host.lima.internal # On lima distributions
```

That seems to cover most cases.

pkg/comp-functions/runtime/desired.go
```diff
@@ -69,8 +69,24 @@ func (d *DesiredResources) GetFromObject(ctx context.Context, o client.Object, k
...
 }
 
-// PutIntoObject adds or updates the desired resource into its kube object
+// PutIntoObject adds or updates the desired resource into its kube object, from docs:
+//
+// The provider can fully manage the resource.
 func (d *DesiredResources) PutIntoObject(ctx context.Context, o client.Object, kon string, refs ...xkube.Reference) error {
+	return d.putIntoObject(ctx, false, o, kon, refs...)
+}
+
```
Kidswiss, line 78:
```suggestion
// PutIntoObserveOnlyObject adds or updates the desired resource into its OBSERVE-ONLY kube object, from docs:
```

Or it won't be a valid godoc line.

pkg/comp-functions/runtime/runtime.go
```diff
@@ -173,3 +174,16 @@ func parseCompFuncConfig(iof *Runtime) (*corev1.ConfigMap, error) {
 func (r *Runtime) GetRawFuncIO() *xfnv1alpha1.FunctionIO {
 	return &r.io
 }
+
+// Small function to help us retrieve bool values from configMap
```
Kidswiss, line 178: Valid godoc comments should always contain the function name as the first word.

Kidswiss (approved):
Some small things, the rest LGTM

README.md
wejdross: ok, I added this snippet

pkg/comp-functions/runtime/runtime.go
```diff
@@ -173,3 +174,16 @@ func parseCompFuncConfig(iof *Runtime) (*corev1.ConfigMap, error) {
 func (r *Runtime) GetRawFuncIO() *xfnv1alpha1.FunctionIO {
 	return &r.io
 }
+
+// GetBoolFromConfigMap is a small function to help us retrieve bool values from configMap
+func (r *Runtime) GetBoolFromConfigMap(key string) bool {
```
Kidswiss, line 179: Can we rename this function? `GetBoolFromConfigMap` sounds confusing as it doesn't indicate what config map is meant.

Maybe a name like `GetBootFromCompositionConfig` would be more helpful.

wejdross: applied it

PR #47, opened 2023-07-26
Fix issue that made redis release immutable

glrf:
We used the observed redis release to build the desired release. This completely dropped any changes from the composition. This PR fixes this by the desired released and only using the observed release to get the current version.

I also included a refactor as the initial bug likely got through review as currently the function is bit convoluted. 

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnredis/release.go
```diff
@@ -178,22 +194,18 @@ func getReleaseValues(r *v1beta1.Release) (map[string]interface{}, error) {
...
+func setReleaseVersion(comp *vshnv1.VSHNRedis, values map[string]interface{}, observed map[string]interface{}) error {
+	var err error
+	tag, _, err := unstructured.NestedString(observed, "image", "tag")
 	if err != nil {
 		return fmt.Errorf("cannot get image tag from values in release: %v", err)
 	}
-	// In case the tag is set then let the maintenance manage it
 	if tag != "" {
-		return nil
+		// In case the tag is set, keep the desired
```
Kidswiss, line 204: You mean keep the observed?

glrf: Yes. You're right I got confused updating these comments :sweat_smile:

Kidswiss (changes requested):
Just one confusing comment I found, the rest LGTM

pkg/comp-functions/functions/vshnredis/release.go
Kidswiss: and I got confused reading the comment :D

Kidswiss (approved):

PR #49, opened 2023-07-28
Allow upgrading redis to newer versions

glrf:
## Summary

Currently we do not allow any changes to the redis version after creation. With this PR we allow to change the version as long as it is a higher minor/major version.

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnredis/release.go
```diff
@@ -194,18 +195,32 @@ func getReleaseValues(r *v1beta1.Release) (map[string]interface{}, error) {
...
-		// In case the tag is set, keep the observed version
-		err = unstructured.SetNestedField(values, tag, "image", "tag")
-	} else {
-		// In case the tag is not set then this is a new Release therefore we need to set the version from the claim
-		err = unstructured.SetNestedField(values, comp.Spec.Parameters.Service.Version, "image", "tag")
+
+	desiredVersion, err := semver.ParseTolerant(comp.Spec.Parameters.Service.Version)
+	if err != nil {
+		l.Info("failed to parse desired redis version", "version", comp.Spec.Parameters.Service.Version)
+		// If the desired version is not parsable, just keep the observed one
```
TheBigLee, line 209: This scenario should probably be reported to the user somehow (eg with an event or something)

glrf: Well this should never happen because they can only choose between `6.2` and `7.0`, but you're right, it's probably better to just return an error and then they issue should bubble up to the user

TheBigLee: LGTM now

TheBigLee (approved):

Kidswiss (approved):

wejdross (approved):
looks good

PR #50, opened 2023-07-31
Fix display issue for Redis Backups

glrf:
There was a display issue caused by the classic Go iteration bug. We attached the same object to each row. This meant if kubectl tired to display additional columns (such as the namespace when listing all redis backups over all namespaces) it showed the same data of the last object for every row.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/apiserver/vshn/redis/table.go
+++ b/pkg/apiserver/vshn/redis/table.go
@@ -32,8 +32,8 @@ func (v *vshnRedisBackupStorage) ConvertToTable(_ context.Context, obj runtime.O
 		backups = append(backups, *backup)
 	}
 
-	for _, backup := range backups {
-		table.Rows = append(table.Rows, backupToTableRow(&backup))
+	for i := range backups {
+		table.Rows = append(table.Rows, backupToTableRow(&backups[i]))
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/apiserver/vshn/redis/table.go
```diff
@@ -32,8 +32,8 @@ func (v *vshnRedisBackupStorage) ConvertToTable(_ context.Context, obj runtime.O
 		backups = append(backups, *backup)
 	}
 
-	for _, backup := range backups {
-		table.Rows = append(table.Rows, backupToTableRow(&backup))
+	for i := range backups {
+		table.Rows = append(table.Rows, backupToTableRow(&backups[i]))
```
glrf, line 36: This is a classic Go bug.

The `backup` variable is reused for every loop iteration, by copying the item in the `backups` slice to the variable. This means that the pointer to `backup` is always the same and points to the `backup` variable and not to the item in the `backups` slice. This means all items the resulting `table.Rows` slice will contain the same pointer to the last object in `backups`.

Here is a discussion page that explains the issue well and also discusses how this could be improved in future Go versions without breaking backwards compatibility too much. https://github.com/golang/go/discussions/56010

PR #51, opened 2023-08-02
Support PVC resize for Redis

Kidswiss:
## Summary

Hanlde disk resizes for redis
    
This will add the ability to handle disk resizes for Redis.
    
As long as the default plan size is used, it will not do anything. But if `parameters.size.disk` is set it will trigger a job that will resize the sts properly and then trigger the helm deployment again.
    
The actual resize of the PVCs is handled by https://github.com/vshn/statefulset-resize-controller. And follows in another PR on that project.

## Checklist

- [ ] Update tests.


pkg/comp-functions/functions/common/rbac.go
```diff
@@ -0,0 +1,72 @@
...
+			Namespace: namespace,
+		},
+		Rules: policies,
+	}
+
+	saReference := xkubev1.Reference{
+		DependsOn: &xkubev1.DependsOn{
+			Name: compName + "-" + suffix + "-serviceaccount",
+		},
+	}
```
wejdross, line 39: I can see potential here for exporting this string to single variable

wejdross: that one: `compName+"-"+suffix+"-role"`

pkg/comp-functions/functions/common/rbac.go
```diff
@@ -0,0 +1,72 @@
...
+
+	roleReference := xkubev1.Reference{
+		DependsOn: &xkubev1.DependsOn{
+			Name: compName + "-" + suffix + "-role",
+		},
+	}
+
+	return iof.Desired.PutIntoObject(ctx, roleBinding, compName+"-"+suffix+"-rolobinding", roleReference, saReference)
```
wejdross, line 71: rolobonding? :P

pkg/comp-functions/functions/vshnredis/pvcresize.go
```diff
@@ -0,0 +1,276 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+)
+
+//go:embed script/recreate.sh
+var stsRecreateScript string
+
+// ResizePVCs will add a job to do the maintenance for the instance
```
wejdross, line 26: this comment looks wrong

pkg/comp-functions/functions/vshnredis/pvcresize.go
```diff
@@ -0,0 +1,276 @@
...
+		},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					ServiceAccountName: "sa-sts-deleter",
+					RestartPolicy:      corev1.RestartPolicyNever,
+					Containers: []corev1.Container{
+						{
+							Name:            "sts-deleter",
+							Image:           "bitnami/kubectl",
```
wejdross, line 182: is it always available image in cluster? I hit some issues in other project when image was not present and cluster could only use some trusted images

pkg/comp-functions/functions/vshnredis/pvcresize.go
```diff
@@ -0,0 +1,276 @@
...
+			Namespace: getInstanceNamespace(comp),
+		},
+	}
+
+	stsBytes, err := json.Marshal(statefulset)
+	if err != nil {
+		return err
+	}
+
+	observer := &xkubev1.Object{
```
wejdross, line 237: maybe it's easier to reuse wrapper func I did? I can do release quite fast and it'll be available on master to rebase

pkg/comp-functions/functions/vshnredis/script/recreate.sh
```diff
@@ -0,0 +1,29 @@
...
+  # Then we patch the right size to enforce an upgrade
+  # This is necessary as provider-helm doesn't actually retry failed helm deployments unless the values change.
+  kubectl patch release "$release" --type merge -p "{\"spec\":{\"forProvider\":{\"values\":{\"master\":{\"persistence\":{\"size\":\"foo\"}}}}}}"
+  kubectl patch release "$release" --type merge -p "{\"spec\":{\"forProvider\":{\"values\":{\"master\":{\"persistence\":{\"size\":\"$size\"}}}}}}"
+  while ! kubectl -n "$namespace" get sts "$name"; do
+    echo "waiting for sts to re-appear"
+    sleep 1
+  done
+  kubectl -n "$namespace" label sts "$name" --overwrite "sts-resize.vshn.net/resize-inplace=true"
+fi
```
wejdross, line 29: this script is kind of a hell, few remarks:
- can You add flag for debugging or just enable some `echo` statements to help engineers debug issues?
- can we add return code checks? currently this script is just yolo running some commands, `set -e` would just terminate, we rather prefer to know precisely where it failed, maybe add `x` flag?
- can we add logic to exit `while` loop instead of spamming our apis each second indefinitely? 

maybe rewrite it to other scripting language? python or something? why is it done in script? it looks pretty much like something that can be done in comp-function

wejdross (commented):
I have few findings, please address them before merging

pkg/comp-functions/functions/vshnredis/pvcresize.go
Kidswiss: This seems like a very specific case. Should I document in the readme that it depends on this image?

pkg/comp-functions/functions/vshnredis/pvcresize.go
Kidswiss: Yes I was hoping for that actually :)

pkg/comp-functions/functions/vshnredis/script/recreate.sh
Kidswiss: I tried to do it in a comp function and I ended up in race condition hell. So no this can't be done easily with comp functions. Also with comp functions this would span over multiple reconcile loops which would result in the whole process taking multiple minutes.

We're using shell scripts for such jobs in various places, for example redis restore and backup. If we insist on implementing this in a proper language then we should do it in go as well.

I'll implement your suggestions for the script.

pkg/comp-functions/functions/common/rbac.go
Kidswiss: I'm not sure I understand what you mean here?

pkg/comp-functions/functions/vshnredis/pvcresize.go
wejdross: yeah, should be fine

pkg/comp-functions/functions/common/rbac.go
wejdross: in line 37 and 22 You use the same multiple variables chains, I'd just replace it with serviceAccountName := fmt.Sprintf("%s-%s-%s-serviceaccount")

Kidswiss: Ah I see, yes I can optimize this. Also if other chains repeat as well.

wejdross (approved):
lgtm

PR #52, opened 2023-08-09
Fix an issue with statefulsets stuck in deletion

Kidswiss:
## Summary
This will add a timeout to the deletion of the sts and if it exceeds that time it will try to patch the finalizers.

There seems to be some weird issue on APPUiO that the first time an sts is deleted with `--cascade=orphan` it will get stuck in the deleting state with a finalizer called `orphan`.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM

PR #54, opened 2023-08-09
Adjusting code to cooperate with Cilium

wejdross:
## Cilium can't properly handle v1.Service without selector set, so I have to update 

* Removing code that updated underlying Stackgres Operator
* Adding v1.Service to allow:
** proper ExternalIP assignment 
** proper port-forward when network set to default or ClusterIP

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
+++ b/pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
@@ -2,18 +2,16 @@ package vshnpostgres
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/crossplane/crossplane/apis/apiextensions/fn/io/v1alpha1"
-	stackgresv1 "github.com/vshn/appcat/apis/stackgres/v1"
 	vshnv1 "github.com/vshn/appcat/apis/vshn/v1"
 	"github.com/vshn/appcat/pkg/comp-functions/runtime"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/pointer"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-var serviceObserverName = "loadbalancer-observer"
+var serviceName = "postgresql-master"
 
 func AddLoadBalancerIPToConnectionDetails(ctx context.Context, iof *runtime.Runtime) runtime.Result {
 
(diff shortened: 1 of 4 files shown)


pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -27,37 +26,55 @@ func AddLoadBalancerIPToConnectionDetails(ctx context.Context, iof *runtime.Runt
...
+	if comp.Spec.Parameters.Network.ServiceType == "LoadBalancer" {
+		k8sservice.Spec.Type = v1.ServiceTypeLoadBalancer
+	} else {
+		k8sservice.Spec.Type = v1.ServiceTypeClusterIP
 	}
 
-	serviceToApply, err := createServiceForLoadBalancer(comp)
-	if err != nil {
-		return runtime.NewFatalErr(ctx, "Cannot create load balancer service", err)
+	fmt.Println(comp.Spec.Parameters.Network.ServiceType)
```
Kidswiss, line 58: This looks like a debug statement, please remove or use the logger with `V(1)` to make it a proper debug log.

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -27,37 +26,55 @@ func AddLoadBalancerIPToConnectionDetails(ctx context.Context, iof *runtime.Runt
 		return runtime.NewFatalErr(ctx, "Cannot get composite from function io", err)
 	}
 
-	if comp.Spec.Parameters.Network.ServiceType != "LoadBalancer" {
-		return runtime.NewNormal()
+	k8sservice := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", comp.GetName(), serviceName),
```
Kidswiss, line 31: I would leave out the `comp.GetName()` here.

If we also want to use this service for port-forwarding it's much simpler for a user, if the service has the same name across the instances. It's also easier for us to document how they can port forward to that service.

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -5,15 +5,14 @@ import (
...
 	vshnv1 "github.com/vshn/appcat/apis/vshn/v1"
 	"github.com/vshn/appcat/pkg/comp-functions/runtime"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/pointer"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-var serviceObserverName = "loadbalancer-observer"
+var serviceName = "k8sservice"
```
Kidswiss, line 15: Maybe call it `postgresql-master`?

Kidswiss (changes requested):
A few small things.

I would have a fixed name for the service like `postgresql-master` or something.

Then we can tell users to use `kubectl -n $instanceNamespace port-forward postgresql-master 5432:5432` if they want to port forward their instance.

If we prepend the service with the comp name, they would firs have to figure that as well, but as they can't actually see the composite it's probably not that easy to get that value.

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
wejdross: whoopsi, leftover from tests, sry

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
wejdross: done

wejdross:
> 

`k -n $instanceNamespace port-forward services/postgresql-master 5432:5432` - possible

Kidswiss (approved):

PR #56, opened 2023-08-10
fix: Ensure APIs can be used by other Go modules

mhutter:
## Summary

The go module versioning rules states that packages with a major version greater than v1 must have a major version suffix
(https://go.dev/ref/mod#module-path)

Since the latest versions of this project are v4, the module name must end in `/v4`. Otherwise,

- regular 'go get' will fetch v0.2.3 of this repo
- 'go get ...@v4.18.1' will complain that the module name is invalid.

https://github.com/vshn/appcat-cli/pull/16

## Checklist

- [x] Update tests.


mhutter:
DRAFT because it needs a `go generate`

Kidswiss (approved):

PR #57, opened 2023-08-11
fixing url.go function

wejdross:
## Summary

* Change NewFatal to NewWarning in Url.go to fix composite function on very slow clusters

## Checklist


Kidswiss:
@wejdross looks like the go.sum change breaks all the checks, so the issue is probably on your machine...

wejdross:
@Kidswiss re-requesting for a review :) 

I don't know what was the issue, I spent more time to find out, but solution was to completely prune Go, install it from official website and reinstall vscode extensions.

PR #60, opened 2023-08-14
renaming object, so it has more meaningful name and it's easier to debug 

wejdross:
## Summary

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
+++ b/pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
@@ -12,7 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-var serviceName = "postgresql-master"
+var serviceName = "primary-service"
 
 func AddLoadBalancerIPToConnectionDetails(ctx context.Context, iof *runtime.Runtime) runtime.Result {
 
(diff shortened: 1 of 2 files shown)


pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
```diff
@@ -55,15 +55,15 @@ func AddLoadBalancerIPToConnectionDetails(ctx context.Context, iof *runtime.Runt
 		k8sservice.Spec.Type = v1.ServiceTypeClusterIP
 	}
 
-	if err := iof.Desired.PutIntoObject(ctx, k8sservice, fmt.Sprintf("%s-%s", comp.GetName(), serviceName)); err != nil {
+	if err := iof.Desired.PutIntoObject(ctx, k8sservice, fmt.Sprintf("%s-%s", comp.GetName(), "service")); err != nil {
```
Kidswiss, line 58: Maybe call it `masterservice` or `mainservice`?

If customers request it, we might also expose the slaves/replicas/secondaries in the future as well.

Kidswiss (changes requested):
Just one nitpick.

pkg/comp-functions/functions/vshn-postgres-func/loadBalancer.go
wejdross: I just renamed it to `primary-service` to comply with Your idea and @TheBigLee suggestions

Kidswiss (approved):

wejdross:
@Kidswiss @TheBigLee I forgot to update this `postgresql-master`, I just pushed commit that fixes that, so both object and underlying service name reuse the same name `primary-service`

This is how it looks like now:
```
Service:
primary-service                  LoadBalancer   10.96.192.252   172.18.0.200 5432:30160/TCP      44s
Object:
postgresql-prod-ww4mc-primary-service              Service             kubernetes       True     True    3m1s
```
wdyt?

Kidswiss (approved):
LGTM

Please adjust any already written docs with the new service name though.

PR #62, opened 2023-08-16
Adjust maintenance procedure for stackgres upgrades

TheBigLee:
## Summary

When upgrading stackgres, the mini controller running as a side-car for every postgresql cluster needs to be upgraded first, before we can do any minor upgrades

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/maintenance/postgresql.go
+++ b/pkg/maintenance/postgresql.go
@@ -6,10 +6,11 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
-	"k8s.io/apimachinery/pkg/watch"
 	"net/http"
 	"sort"
 	"time"
+
+	"k8s.io/apimachinery/pkg/watch"
 
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-version"
@@ -83,11 +84,27 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
 		p.log.Info("No sgcluster found in namespace, skipping maintenance")
 	}
 
-	clusterName := sgclusters.Items[0].GetName()
+	sgCluster := sgclusters.Items[0]
 	currentVersion := sgclusters.Items[0].Spec.Postgres.Version
 
+	p.log.Info("Checking for pending upgrades...")
+	upgradeRequired := p.checkRequiredUpgrade(sgCluster)
+
+	if upgradeRequired {
+		p.log.Info("Doing a security maintenance")
+		p.createSecurityUpgrade(sgCluster.GetName())
+		p.log.Info("Waiting for security maintenance to finish before checking for minor upgrades")
+		err = p.waitForUpgrade(ctx, su)
+		if apierrors.IsTimeout(err) {
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("cannot watch for maintenance sgdbops resources: %v", err)
+		}
+	}
+
 	p.log.Info("Checking for upgrades...")
-	op, err := p.upgradeVersion(currentVersion, clusterName)
+	op, err := p.upgradeVersion(currentVersion, sgCluster.GetName())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/maintenance/postgresql.go
```diff
@@ -102,15 +120,29 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
...
-func (p *PostgreSQL) upgradeVersion(currentVersion string, clusterName string) (OpName, error) {
+func (p *PostgreSQL) checkRestart(sgCluster stackgresv1.SGCluster) (bool, error) {
+	if sgCluster.Status.Conditions == nil {
+		return false, nil
+	}
+	sgClusterConditions := *sgCluster.Status.Conditions
+	for _, condition := range sgClusterConditions {
+		if *condition.Reason == "ClusterRequiresUpgrade" && *condition.Status == "True" {
+			p.log.Info("Doing a security maintenance")
+			return true, p.createSecurityUpgrade(sgCluster.GetName())
```
zugao, line 139: I would get this function out of `checkRestart()` as the name of the method is miss leading since you are doing a security upgrade in the happy path.

pkg/maintenance/postgresql.go
```diff
@@ -102,15 +120,29 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
...
+	err = p.createRepack(sgCluster.GetName())
 	if err != nil {
 		return fmt.Errorf("cannot create repack: %v", err)
 	}
 
 	return nil
 }
 
-func (p *PostgreSQL) upgradeVersion(currentVersion string, clusterName string) (OpName, error) {
+func (p *PostgreSQL) checkRestart(sgCluster stackgresv1.SGCluster) (bool, error) {
```
zugao, line 131: I don't understand why this function is call `checkRestart()`

zugao (commented):
Some things to clarify

pkg/maintenance/postgresql.go
TheBigLee: Taken it out of the function

pkg/maintenance/postgresql.go
TheBigLee: No it should make sense (see above thread)

pkg/maintenance/postgresql.go
```diff
@@ -102,15 +119,29 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
...
+	err = p.createRepack(sgCluster.GetName())
 	if err != nil {
 		return fmt.Errorf("cannot create repack: %v", err)
 	}
 
 	return nil
 }
 
-func (p *PostgreSQL) upgradeVersion(currentVersion string, clusterName string) (OpName, error) {
+func (p *PostgreSQL) checkRestart(sgCluster stackgresv1.SGCluster) bool {
```
Kidswiss, line 130: `checkRestart` seems like a misnomer. We don't check for the restart condition (which is actually a seperate condition that exists), but we check the `ClusterRequiresUpgrade` condition.

So maybe `checkRequiredUpgrade`?

pkg/maintenance/postgresql.go
```diff
@@ -83,11 +84,27 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
...
+	sgCluster := sgclusters.Items[0]
 	currentVersion := sgclusters.Items[0].Spec.Postgres.Version
 
+	p.log.Info("Checking for pending restarts...")
+	restartRequired := p.checkRestart(sgCluster)
+
+	if restartRequired {
+		p.log.Info("Doing a security maintenance")
+		p.createSecurityUpgrade(sgCluster.GetName())
+		p.log.Info("Waiting for security maintenance to finish before checking for upgrades")
```
Kidswiss, line 96:
```suggestion
		p.log.Info("Waiting for security maintenance to finish before checking for minor upgrades")
```

Kidswiss (changes requested):
Just some minor naming issues, the rest LGTM

zugao (approved):
LGTM

pkg/maintenance/postgresql.go
TheBigLee: Good point. Renamed the function and the related Log output

Kidswiss (approved):

PR #66, opened 2023-08-28
Fix docs workflow

zugao:
## Summary
 
The docs would not publish due to an error. This PR should resolve the issue.

## Checklist


Kidswiss (approved):
No idea what this is supposed to do but LGTM

PR #70, opened 2023-09-07
Add Miniobucket composition function

Kidswiss:
## Summary

This composition function will create:
    
* a bucket
* a policy
* a user
    
The policy is then assigned to the user and only allows that user to access exactly that bucket.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM bu I haven't tested it

Kidswiss:
> LGTM bu I haven't tested it

Yeah I just found out that the maintenance tests broke after all those dependency updates. I'm debugging that now. Maybe it uncovered a new race condition in the test...

Kidswiss:
@zugao could you have a look at `maintenance/postgresql_test.go`? I had to kick it quite a bit for the tests to work again.

Also I refactored it a bit, it silently dropped errors before, which made debugging this a lot harder.

Kidswiss:
> Any reason why you are paniking and not returning errors? Is it due to the goroutine?

I found two reasons:

1. yes because of the go routine, the previous version just returned the errors, but they were actually never handled.
2. if anything goes wrong in that function without panicking it will just be stuck and run into the timeout.

PR #74, opened 2023-09-14
Avoid dropping the Organization label

Kidswiss:
## Summary

This commit makes the adding of the quota annotations a bit more robust in a way that we don't drop it by mistake.

If it encounters a namespace without the organization label it will add it. As it's required for the quotas to work anyways.

## Checklist

- [x] Update tests.


wejdross (approved):
LGTM

PR #78, opened 2023-09-19
Fix resource calculation by taking sidecars into account

TheBigLee:
## Summary

* The resources of the sidecars are not taken into account when calculating the quota for the instance namespace. This results in too low quotas for the instance namespace when deploying multiple instances of a PostgreSQL instance, as the namespace doesn't have high enough quota defined.
* This PR addresses this issue, by taking the resources of the sidecars into account when calculating the required quotas for the instance namespace.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/common/quotas/quotas.go
+++ b/pkg/common/quotas/quotas.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	"github.com/vshn/appcat/v4/pkg/common/utils"
+	iofruntime "github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
@@ -25,12 +26,13 @@ type QuotaChecker struct {
 	gr                  schema.GroupResource
 	gk                  schema.GroupKind
 	checkNamespaceQuota bool
+	instances           int64
 }
 
 // NewQuotaChecker creates a new quota checker.
 // checkNamespaceQuota specifies whether or not the checker should also take the amount of existing
 // namespaces into account, or not.
-func NewQuotaChecker(c client.Client, claimName, claimNamespace, instanceNamespace string, r utils.Resources, gr schema.GroupResource, gk schema.GroupKind, checkNamespaceQuota bool) QuotaChecker {
+func NewQuotaChecker(c client.Client, claimName, claimNamespace, instanceNamespace string, r utils.Resources, gr schema.GroupResource, gk schema.GroupKind, checkNamespaceQuota bool, instances int64) QuotaChecker {
 	return QuotaChecker{
 		requestedResources:  r,
 		instanceNamespace:   instanceNamespace,
@@ -40,6 +42,7 @@ func NewQuotaChecker(c client.Client, claimName, claimNamespace, instanceNamespa
 		gr:                  gr,
 		gk:                  gk,
 		checkNamespaceQuota: checkNamespaceQuota,
+		instances:           instances,
 	}
 }
 
@@ -60,7 +63,7 @@ func (q *QuotaChecker) CheckQuotas(ctx context.Context) *apierrors.StatusError {
 		return nsErr
 	}
 
-	return q.requestedResources.CheckResourcesAgainstQuotas(ctx, q.client, q.claimName, q.instanceNamespace, q.gk)
+	return q.requestedResources.CheckResourcesAgainstQuotas(ctx, q.client, q.claimName, q.instanceNamespace, q.gk, q.instances)
 }
 
 // areNamespacesWithinQuota checks for the given organization if there are still enough namespaces available.
@@ -148,7 +151,7 @@ func (q *QuotaChecker) getNamespaceOverrides(ctx context.Context, c client.Clien
 
 // AddInitalNamespaceQuotas will add the default quotas to the namespace annotations.
 // It will only add them, if there are currently no such annotations in place.
-func AddInitalNamespaceQuotas(ns *corev1.Namespace) bool {
+func AddInitalNamespaceQuotas(ctx context.Context, iof *iofruntime.Runtime, ns *corev1.Namespace, s *utils.Sidecars, kind string) bool {
 	annotations := ns.GetAnnotations()
 	if annotations == nil {
 		annotations = map[string]string{}
@@ -156,28 +159,30 @@ func AddInitalNamespaceQuotas(ns *corev1.Namespace) bool {
 
 	added := false
 
+	r := utils.GetDefaultResources(kind, s)
+
 	if _, ok := annotations[utils.DiskAnnotation]; !ok {
 		annotations[utils.DiskAnnotation] = utils.DefaultDiskRequests.String()
 		added = true
 	}
 
 	if _, ok := annotations[utils.CpuRequestAnnotation]; !ok {
-		annotations[utils.CpuRequestAnnotation] = utils.DefaultCPURequests.String()
+		annotations[utils.CpuRequestAnnotation] = r.CPURequests.String()
 		added = true
 	}
 
 	if _, ok := annotations[utils.CpuLimitAnnotation]; !ok {
-		annotations[utils.CpuLimitAnnotation] = utils.DefaultCPULimit.String()
+		annotations[utils.CpuLimitAnnotation] = r.CPULimits.String()
 		added = true
 	}
 
 	if _, ok := annotations[utils.MemoryRequestAnnotation]; !ok {
-		annotations[utils.MemoryRequestAnnotation] = utils.DefaultMemoryRequests.String()
+		annotations[utils.MemoryRequestAnnotation] = r.MemoryRequests.String()
 		added = true
 	}
 
 	if _, ok := annotations[utils.MemoryLimitAnnotation]; !ok {
-		annotations[utils.MemoryLimitAnnotation] = utils.DefaultMemoryLimits.String()
+		annotations[utils.MemoryLimitAnnotation] = r.MemoryLimits.String()
 		added = true
 	}
 
--- a/pkg/common/utils/resources.go
+++ b/pkg/common/utils/resources.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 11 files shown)


pkg/common/utils/resources.go
```diff
@@ -73,14 +73,14 @@ var (
...
 
-	// defaultCPURequests 2* standard-8 will request 4 CPUs. This default has 500m as spare for jobs
-	DefaultCPURequests = resource.NewMilliQuantity(4500, resource.DecimalSI)
-	// defaultCPULimit by default same as DefaultCPURequests
-	DefaultCPULimit = DefaultCPURequests
-	// defaultMemoryRequests 2* standard-8 will request 16Gb. This default has 500mb as spare for jobs
-	DefaultMemoryRequests = resource.NewQuantity(17301504000, resource.BinarySI)
-	// defaultMemoryLimits same as DefaultMemoryRequests
-	DefaultMemoryLimits = DefaultMemoryRequests
+	// defaultCPURequests 2* standard-8 will request 4 CPUs + the sidecars will each request 850m. This default has 500m as spare for jobs
```
Kidswiss, line 76: I would not hardcode the sidecar limits, simply because not all of our services have sidecars.

For example this would give redis much more limits than it needs because it doesn't have sidecars.

I probably would have added it to the comp-function `AddInitialNamespaceQuotas` and inject the overhead depending on the service.

pkg/common/utils/resources.go
```diff
@@ -248,3 +248,11 @@ func (r *Resources) MultiplyBy(i int64) {
 	r.MemoryRequests.Set(r.MemoryRequests.Value() * i)
 	r.Disk.Set(r.Disk.Value() * i)
 }
+
+func (r *Resources) AddByResource(resource Resources) {
```
Kidswiss, line 252: This function name is confusing :)

`AddByResource` seems to imply that you do some addition depending on the type of the resource.

I would probably call it `AddResources` or something, so it's clear that `resources` is added to `r` when you do `r.AddResources(resources)`

pkg/common/utils/sidecars.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return Resources{}, err
+	}
+
+	r, err := convertSidecarToResource(sidecar, *s)
+
+	return r, err
+}
+
+func convertSidecarToResource(sidecar string, s Sidecars) (Resources, error) {
```
Kidswiss, line 79: `Sidecars` is already a custom type sou you could make this a method of this type:

`func (s *Sidecars) ConvertSidecarToResource(sidecar string) (Resources, error)`

Kidswiss (changes requested):
Looks good so far, just a few things to improve

pkg/common/utils/resources.go
TheBigLee: Fixed

pkg/common/utils/sidecars.go
TheBigLee: Done

pkg/common/utils/resources.go
TheBigLee: Did a lot of refactoring and changes here. Please check again

pkg/common/quotas/quotas.go
```diff
@@ -148,36 +153,50 @@ func (q *QuotaChecker) getNamespaceOverrides(ctx context.Context, c client.Clien
...
+	r := &utils.Resources{
+		CPURequests:    *utils.DefaultCPURequests,
+		CPULimits:      *utils.DefaultCPULimits,
+		MemoryRequests: *utils.DefaultMemoryRequests,
+		MemoryLimits:   *utils.DefaultMemoryLimits,
+	}
+
+	if objMeta.Kind == "XVSHNPostgreSQL" {
+		comp := vshnv1.VSHNPostgreSQL{}
+		iof.Desired.GetComposite(ctx, &comp)
```
Kidswiss, line 173: You're dropping an error here...

pkg/common/quotas/quotas.go
```diff
@@ -148,36 +153,50 @@ func (q *QuotaChecker) getNamespaceOverrides(ctx context.Context, c client.Clien
...
 
+	r := &utils.Resources{
+		CPURequests:    *utils.DefaultCPURequests,
+		CPULimits:      *utils.DefaultCPULimits,
+		MemoryRequests: *utils.DefaultMemoryRequests,
+		MemoryLimits:   *utils.DefaultMemoryLimits,
+	}
+
+	if objMeta.Kind == "XVSHNPostgreSQL" {
+		comp := vshnv1.VSHNPostgreSQL{}
```
Kidswiss, line 172: You could put this logic into its own function, then `AddInitalNamespaceQuotas` will be a bit shorter

pkg/common/utils/resources.go
```diff
@@ -181,24 +181,36 @@ func (r *Resources) getNamespaceQuotas(ctx context.Context, c client.Client, ins
 
 // checkAgainstResources compares this resources against the given resources.
 // Any non-populated fields are checked against their defaults.
-func (r *Resources) checkAgainstResources(quotaResources Resources) field.ErrorList {
+func (r *Resources) checkAgainstResources(quotaResources Resources, gk schema.GroupKind, instances int64) field.ErrorList {
 	foundErrs := field.ErrorList{}
 
 	errCPURequests := field.Forbidden(r.CPURequestsPath, "Max allowed CPU requests: "+DefaultCPURequests.String()+". Configured requests: "+r.CPURequests.String()+". "+contactSupportMessage)
+
+	rDef := &Resources{
```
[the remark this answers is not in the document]
TheBigLee: Good point. Created new function for that

pkg/common/quotas/quotas.go
TheBigLee: Extracted that logic into a separte function

pkg/common/quotas/quotas.go
```diff
@@ -148,39 +153,58 @@ func (q *QuotaChecker) getNamespaceOverrides(ctx context.Context, c client.Clien
...
 }
+
+// addSidecarResources will add the sidecar resources to the given resources.
+func addSidecarResources(ctx context.Context, objMeta *metaonly.MetadataOnlyObject, iof *iofruntime.Runtime, r *utils.Resources) {
+	if objMeta.Kind == "XVSHNPostgreSQL" {
+		comp := vshnv1.VSHNPostgreSQL{}
+		err := iof.Desired.GetComposite(ctx, &comp)
+		var instances int
+		if err != nil {
+			instances = 1
```
Kidswiss, line 204: Maybe log that there was an error here, or else $someone will go crazy figuring out why the quota calculation assumes 1 instance.

Kidswiss (changes requested):
One one more thing then I think it looks good for now.

pkg/common/quotas/quotas.go
TheBigLee: Added a log message

TheBigLee:
Deployed on Lab and ready for final review and testing

pkg/common/utils/resources.go
```diff
@@ -181,24 +181,31 @@ func (r *Resources) getNamespaceQuotas(ctx context.Context, c client.Client, ins
...
 // Any non-populated fields are checked against their defaults.
-func (r *Resources) checkAgainstResources(quotaResources Resources) field.ErrorList {
+func (r *Resources) checkAgainstResources(quotaResources Resources, gk schema.GroupKind, instances int64) field.ErrorList {
 	foundErrs := field.ErrorList{}
 
 	errCPURequests := field.Forbidden(r.CPURequestsPath, "Max allowed CPU requests: "+DefaultCPURequests.String()+". Configured requests: "+r.CPURequests.String()+". "+contactSupportMessage)
+
+	rDef := GetDefaultResources()
+
+	if gk.Kind == "VSHNPostgreSQL" {
```
Kidswiss, line 191: I'm still a bit triggered by this, but I think for now it's okay.

We can have another looks at this if/when we have more services with sidecars. (probably when we add the exporter to redis)

Kidswiss (approved):

pkg/common/utils/resources.go
```diff
@@ -248,3 +255,32 @@ func (r *Resources) MultiplyBy(i int64) {
...
+	r.MemoryLimits.Add(resource.MemoryLimits)
+	r.MemoryRequests.Add(resource.MemoryRequests)
+	r.Disk.Add(resource.Disk)
+}
+
+// AddPsqlSidecarResources adds the resource overhead for the PostgreSQL sidecar to the given resource.
+func (r *Resources) AddPsqlSidecarResources(instances int64) {
+	// Sidecar requests: 850m CPU, 1088Mi Memory
+	r.CPURequests.Add(*resource.NewMilliQuantity(850*instances, resource.DecimalSI))
+	r.MemoryRequests.Add(*resource.NewQuantity(1088*1024*1024*instances, resource.BinarySI))
```
Kidswiss, line 271: As discussed, please make these dynamically according to the definitions in the configMap.

pkg/common/quotas/quotas.go
```diff
@@ -148,39 +153,60 @@ func (q *QuotaChecker) getNamespaceOverrides(ctx context.Context, c client.Clien
...
+
+// addSidecarResources will add the sidecar resources to the given resources.
+func addSidecarResources(ctx context.Context, objMeta *metaonly.MetadataOnlyObject, iof *iofruntime.Runtime, r *utils.Resources) {
+	l := ctrl.LoggerFrom(ctx)
+	if objMeta.Kind == "XVSHNPostgreSQL" {
+		comp := vshnv1.VSHNPostgreSQL{}
+		err := iof.Desired.GetComposite(ctx, &comp)
+		var instances int
+		if err != nil {
+			l.Info("Could not get composite. Assuming 1 instance.")
```
wejdross, line 205: why do we assume? what are the situations when we can't access Composite?

pkg/common/utils/resources.go
wejdross: not perfect, but easy to append, I'd just leave it as it is, if works correctly

wejdross (commented):
overall very good, small things already catched by Simon

pkg/common/quotas/quotas.go
```diff
@@ -148,39 +153,60 @@ func (q *QuotaChecker) getNamespaceOverrides(ctx context.Context, c client.Clien
...
 	}
 
 	ns.SetAnnotations(annotations)
 	return added
 }
+
+// addSidecarResources will add the sidecar resources to the given resources.
+func addSidecarResources(ctx context.Context, objMeta *metaonly.MetadataOnlyObject, iof *iofruntime.Runtime, r *utils.Resources) {
+	l := ctrl.LoggerFrom(ctx)
+	if objMeta.Kind == "XVSHNPostgreSQL" {
```
zugao, line 200: This `objMeta` comes from  `iof.Desired.GetComposite(ctx, <some-object>)`. You expect to have an XRD out of it from the if clause then you do run the same method `iof.Desired.GetComposite(ctx, <some-object>)` and you expect a claim. I don't understand what is happening here.

pkg/common/quotas/quotas.go
TheBigLee: This has been removed in some refactorings

pkg/common/quotas/quotas.go
TheBigLee: This has been refactored completely.

pkg/common/utils/resources.go
TheBigLee: Refactored the code to get the values from the config

Kidswiss (approved):
After fixing the defaulting logic yesterday in a 1 on 1 and testing it on the lab it should now be okay to merge and release.

TheBigLee:
> After fixing the defaulting logic yesterday in a 1 on 1 and testing it on the lab it should now be okay to merge and release.
> 
> EDIT: maybe amend `docs/pages/explanations/webhooks.adoc` with the new logic.

Addes some documentation for the webhooks

Kidswiss (approved):

PR #79, opened 2023-09-20
Switching Reconciliation in SLI Prober from claim to composite resource

wejdross:
## Summary

Switching from composition to composite in Prober Reconciliation to avoid issues with monitoring and aleerting when instance is recreated with the same name in the same namespace.

Solution respects our decision to use the same credentials v1/Secret which user uses, this is why I'm fetching composition and take different Secret. 

Tested using:
- test files
- local kind cluster
- lab cluster


Kidswiss:
I guess the title should be "Switching Reconciliation in SLI Prober from claim to composite resource"?

pkg/sliexporter/vshnpostgresql_controller/vshnpostgresql_controller.go
```diff
@@ -137,7 +152,7 @@ func (r VSHNPostgreSQLReconciler) fetchProberFor(ctx context.Context, inst *vshn
 		sla = vshnv1.BestEffort
 	}
 
-	probe, err := r.PostgreDialer(vshnpostgresqlsServiceKey, inst.Name, inst.Namespace,
+	probe, err := r.PostgreDialer(vshnpostgresqlsServiceKey, inst.Name, inst.ObjectMeta.Labels[claimNamespaceLabel],
```
Kidswiss, line 155: You're using this quite a few times here, maybe put it in a var?

pkg/sliexporter/vshnredis_controller/vshnredis_controller.go
```diff
@@ -97,26 +105,37 @@ func (r *VSHNRedisReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	}
+
+	credentials := v1.Secret{}
+
 	err = r.Get(ctx, types.NamespacedName{
-		Name:      inst.Spec.WriteConnectionSecretToRef.Name,
-		Namespace: inst.Namespace,
-	}, &credSecret)
+		Name:      instance.Spec.WriteConnectionSecretToRef.Name,
+		Namespace: inst.ObjectMeta.Labels[claimNamespaceLabel],
```
Kidswiss, line 123: Here as well.

Kidswiss (approved):
LGTM

Just one nitpick

PR #81, opened 2023-09-21
Add instances label to prober metrics

zugao:
## Summary

This PR adds instances label to the prober metrics. This will allow to distinguish between HA and non HA instances.

## Checklist

- [x] Update tests.


go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.3
 	github.com/hashicorp/go-version v1.6.0
+	github.com/huang-wei/kubernetes v1.14.9
```
Kidswiss, line 20: Why did remove `k8s.io/kubernetes v1.26.3` and add this one?

zugao: Had some issues with my IDE, as discussed in the chat I will port in our code the necessary code.

zugao: Ported everything in `apis/custom`

Kidswiss (approved):

apis/custom/custom_types.go
```diff
@@ -0,0 +1,29 @@
+package custom
```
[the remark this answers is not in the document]
zugao: update the name, thanks.

Kidswiss:
I just had a though, while looking at https://github.com/vshn/component-appcat/pull/231

Why not just expose a label `HA` which would be a boolean? Because we have multiple values that could indicate HA setups. For PostgreSQL that would be 2 and 3 instances. And that could get very painful to relfect in the prom queries.

zugao:
You made a good point and I decided to change how we export the HA value.

Kidswiss (approved):
LGTM

With this we've already future proofed this logic a bit.

PR #82, opened 2023-09-22
Add Minio Maintenance

Kidswiss:
## Summary

This will add Minio's maintenance to the AppCat. Minio is a bit special with their releases, as it's a rolling release.
    
In order to support generic helm maintenances in the future a new module was created. This module can handle any helm release and update its tag according to a given compare strategy.

## Checklist

- [x] Update tests.


pkg/maintenance/helm/helmpatcher.go
```diff
@@ -0,0 +1,252 @@
...
+	// Name is the tag
+	Name string `json:"name,omitempty"`
+	// TagStatus indicates if the tag is active
+	TagStatus string `json:"tag_status,omitempty"`
+	// ContentType indicates whether this Result is an image
+	ContentType string `json:"content_type,omitempty"`
+}
+
+// VersionComparisonStrategy is a function that actually determines if a given release needs to be updated
+// The returned string should contain the latest version, the boolean determines if the version is actually newer or not
```
zugao, line 87: I'm confused here, which boolean you are referring to?

pkg/maintenance/helm/helmpatcher.go
```diff
@@ -0,0 +1,252 @@
...
+
+	h.log.Info("Getting current image tag, from path", "fieldPath", strings.Join(tagPath.AsSlice(), ","))
+
+	currentTag, err := h.getCurrentTagFromRelease(release, tagPath.AsSlice())
+	if err != nil {
+		return fmt.Errorf("cannot get current tag from release: %w", err)
+	}
+
+	h.log.Info("Getting latest image version")
+	version, err := versionComp(results, currentTag) //h.getNewSemanticVersion(release, results, currentTag)
```
zugao, line 127: This comment is confusing.

pkg/maintenance/minio.go
```diff
@@ -0,0 +1,154 @@
...
+		return err
+	}
+
+	return patcher.DoMaintenance(ctx, minioURL, valuesPath, compareMinioVersions)
+}
+
+// compareMinioVersions specifically checks for new Minio versions
+// Minio has a pretty unique rolling release cycle which is based on the date.
+// Each Minio image tag contains the date when it was created.
+// The form of the tags is like this: `RELEASE.2023-09-16T01-01-47Z`
```
zugao, line 59: Jesus Christ.

pkg/maintenance/minio.go
```diff
@@ -0,0 +1,154 @@
...
+func compareMinioVersions(results []helm.Result, currentTag string) (string, error) {
+
+	if currentTag == "" {
+		currentTag = minioTagPrefix + earliestTag
```
zugao, line 63: Any reason you have to deduce that? I mean you should be able to get the version of the current release no?

pkg/maintenance/minio.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	currentDate, err := parseMinioDate(currentTag)
```
zugao, line 66: If this is empty string wouldn't it return an error? if not you could return an error an the above if clause can be removed no?

pkg/maintenance/minio.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	newDate := currentDate
+
+	for _, res := range results {
+		if res.TagStatus == helm.ActiveTagStatus && res.ContentType == helm.ImageContentType {
+			v, err := parseMinioDate(res.Name)
+			if err != nil {
+				continue
```
zugao, line 80: you might want to log here at least in debug mode that a version could not be parsed.

pkg/maintenance/minio.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	_, found, err := unstructured.NestedString(values, valuesPath.AsSlice()...)
+	if err != nil {
+		return err
+	}
+
+	if !found {
+
+		m.log.Info("No tag found, adding empty tag")
```
zugao, line 133: Why would you expect here an empty field tag?

pkg/maintenance/helm/helmpatcher.go
Kidswiss: the one I removed after writing the comment and realizing it doesn't provide any benefits 🤪

pkg/maintenance/minio.go
Kidswiss: I know right? It's not even an RFC format...

pkg/maintenance/minio.go
Kidswiss: We don't set it initially, so the release takes the defaults, but that is not actually populated in the helm values. This logic is to catch this case, which is the very first maintenance after provisioning.

pkg/maintenance/minio.go
Kidswiss: Se explanation above, it's always empty string on the first maintenance.

zugao (approved):
LGTM

PR #95, opened 2023-10-31
Add Sli prober for VSHNMinio

wejdross:
## Summary

* Adding XObjectBucket to minio_deploy.go, so we have test bucket for each Minio server
* Adding Minio prober and Controller accordingly
* Adding tests

## Checklist

- [x] Update tests.


cmd/sliexporter.go
```diff
@@ -94,6 +97,20 @@ func (s *sliProber) executeSLIProber(cmd *cobra.Command, _ []string) error {
 			return err
 		}
 	}
+
+	if s.enableVSHNMinio || true {
```
Kidswiss, line 101: Is the `|| true` some debugging leftover? With this we can never disable the minio prober

pkg/comp-functions/functions/vshnminio/minio_deploy.go
```diff
@@ -295,3 +307,27 @@ func createServiceMonitor(ctx context.Context, comp *vshnv1.VSHNMinio, iof *runt
...
+				Name:      SLIBucketName,
+				Namespace: comp.GetInstanceNamespace(),
+			},
+			CompositionReference: crossplane.CompositionReference{
+				Name: fmt.Sprintf("%s.objectbuckets.appcat.vshn.io", xminioName),
+			},
+		},
+	}
+	return iof.Desired.PutWithResourceName(ctx, obj, comp.Name+"-vshn-test-bucket-for-sli")
+	//return iof.Desired.PutIntoObject(ctx, obj, comp.Name+"-vshn-test-bucket-for-sli")
```
Kidswiss, line 332: Also a leftover I guess?

pkg/comp-functions/functions/vshnminio/minio_deploy.go
```diff
@@ -295,3 +307,27 @@ func createServiceMonitor(ctx context.Context, comp *vshnv1.VSHNMinio, iof *runt
...
+func createSliBucket(ctx context.Context, comp *vshnv1.VSHNMinio, xminioName string, iof *runtime.Runtime) error {
+	obj := &v1.ObjectBucket{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      SLIBucketName,
+			Namespace: comp.GetInstanceNamespace(),
+		},
+		Spec: v1.ObjectBucketSpec{
+			Parameters: v1.ObjectBucketParameters{
+				BucketName: SLIBucketName,
+				Region:     "rma",
```
Kidswiss, line 320: I wouldn't use a cloudscale.ch region here. Please use Minio's default region `us-east-1`.

Currently it doesn't matter, as we don't enable region support in Minio, but if we ever would, it probably breaks with this region.

pkg/sliexporter/probes/minio.go
```diff
@@ -0,0 +1,63 @@
...
+		Namespace:     minio.Namespace,
+		HighAvailable: minio.HighAvailable,
+		Organization:  minio.Organization,
+		ServiceLevel:  minio.ServiceLevel,
+	}
+}
+
+func (minio VSHNMinio) Probe(ctx context.Context) error {
+
+	x := bytes.NewBufferString("This file is auto-generated, do not edit, VSHN internal purposes")
```
Kidswiss, line 37: You could even be more specific and write that it's for the SLI proper.

pkg/sliexporter/vshnminio_controller/vshnminio_controller.go
```diff
@@ -0,0 +1,158 @@
...
+		Name:      SLIBucketName,
+		Namespace: "vshn-minio-" + inst.Name,
+	}, &creds)
+
+	if err != nil {
+		return nil, err
+	}
+
+	org := inst.GetLabels()[utils.OrgLabelName]
+	if org == "" {
```
Kidswiss, line 120: This will always result in `vshn` org on any APPUiO Manage cluster, as they don't have this label.

If the label is empty:
Either we need some way to determine what organization any given APPUiO Managed cluster belongs to, or we make it configurable via some flag or env var. Then we can just set it in SYN during deployment.

Probably would make sense to have that logic for the other probers as well.

pkg/sliexporter/vshnminio_controller/vshnminio_controller.go
```diff
@@ -0,0 +1,158 @@
...
+		// for development
+		org = "vshn"
+	}
+
+	ha := false
+	sla := "besteffort"
+	if inst.Spec.Parameters.Instances >= 4 {
+		sla = "guaranteed"
```
Kidswiss, line 128: More than one instance doesn't automatically mean it's guaranteed. It's a separate flag. You can check PostgreSQL how it's handled there.

Kidswiss (changes requested):
Found a few things that don't quite match up.

pkg/sliexporter/vshnminio_controller/vshnminio_controller.go
wejdross: I talked wtih Sebastian about that and we have 2 additional labels coming from each managed Openshift Cluster with syn-tenant-id and cluster name. Isn't that sufficient?

pkg/sliexporter/vshnminio_controller/vshnminio_controller.go
wejdross: but we don't have logic for Service Level, I didn't add that field intentionally to not confuse customers if they `kubectl exaplain` claim and spot that config option, I'll default to "besteffort" for now, wdyt?

Kidswiss: Okay fine by me.

pkg/sliexporter/vshnminio_controller/vshnminio_controller.go
Kidswiss: Yeah that would work, but in that case we need:
* not setting the org to vshn if it's empty
* teach the sla reporter to look at `syn-tenant-id` if the organization label is not set

We should open a followup for that.

wejdross: https://github.com/vshn/appcat/pull/95/files#diff-3243738ee01fda0e17f64917b3b41de603fa1685b108a6fc6d89e052b1470ec8R129

removed defaulting logic

Kidswiss (approved):

PR #97, opened 2023-11-03
Fix prober when namespace claim is deleted

zugao:
## Summary

The SLI prober fails to stop the prober for an instance in a namespace that was deleted. This leads to a non ending request errors from the prober due to the fact that the namespace is not retrieved anymore from the request itself once it's marked as deleted. The solution is to get the namespace from the XRD labels in case the request is missing it.

## Checklist

### DIFF
--- a/pkg/sliexporter/probes/manager.go
+++ b/pkg/sliexporter/probes/manager.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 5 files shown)


Kidswiss (approved):
So that means the namespace is still relevant, even though we use the compites and not the claims?

zugao:
> So that means the namespace is still relevant, even though we use the compites and not the claims?

yes the namespace is still relevant, for instance in the metrics we return, we want to expose the namespace. In any case we can leave the namespace in place we just need to make sure we can stop the prober properly.

PR #99, opened 2023-11-07
Forward events on appcat objects to claim specified in annotation

TheBigLee:
## Summary

This PR will allow us to forward events from AppCat objects to the claim specified in the annotation `vshn.net/forward-events-to`

## Checklist

- [ ] Update tests.


pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,93 @@
...
+		return ctrl.Result{}, err
+	}
+	ef := obj.GetAnnotations()["vshn.net/forward-events-to"]
+	if ef != "" {
+		log.Info("Forwarding event", "event", event.Name, "to", ef)
+		knc := strings.Split(ef, "/")
+		a := knc[0] + "/" + knc[1]
+		k := knc[2]
+		ns := knc[3]
+		c := knc[4]
```
wejdross, line 43: please refactor it and use meaningful variable names

pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := e.Get(ctx, req.NamespacedName, event); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	obj := &unstructured.Unstructured{}
+	obj.SetGroupVersionKind(schema.FromAPIVersionAndKind(event.InvolvedObject.APIVersion, event.InvolvedObject.Kind))
+	if err := e.Get(ctx, types.NamespacedName{Name: event.InvolvedObject.Name, Namespace: event.InvolvedObject.Namespace}, obj); err != nil {
+		return ctrl.Result{}, err
+	}
+	ef := obj.GetAnnotations()["vshn.net/forward-events-to"]
```
wejdross, line 36: in which place do we set this annotation? or is it meant to be set by user?

wejdross (commented):
How about performance, does reconciliation over all possible events in cluster hits performance? Have You tested it on LAB for example?

wejdross:
also: this is still in Draft - intentionally?

pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,93 @@
...
+			InvolvedObject: corev1.ObjectReference{
+				Kind:            k,
+				Namespace:       ns,
+				Name:            c,
+				APIVersion:      a,
+				ResourceVersion: claim.GetResourceVersion(),
+				UID:             claim.GetUID(),
+			},
+		}
+		_, err := controllerutil.CreateOrPatch(ctx, e.Client, nev, func() error {
```
Kidswiss, line 64: I'm not sure if patching existing events makes sense here. Won't old events pup-up at the top again in such a case?

pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,93 @@
...
+		k := knc[2]
+		ns := knc[3]
+		c := knc[4]
+
+		claim := &unstructured.Unstructured{}
+		claim.SetGroupVersionKind(schema.FromAPIVersionAndKind(a, k))
+		if err := e.Get(ctx, types.NamespacedName{Name: c, Namespace: ns}, claim); err != nil {
+			return ctrl.Result{}, err
+		}
+		nev := &corev1.Event{
```
Kidswiss, line 50: We probably want at least `message`, `type` and `reason` as well here? Or otherwise the event is just empty? You can just copy it from the source event.

Kidswiss (changes requested):
Some things I found while looking at this.

Please don't be afraid to use descriptive variable names. The IDE has autocompletion, so typing less is not an argument ;)

Kidswiss:
> How about performance, does reconciliation over all possible events in cluster hits performance? Have You tested it on LAB for example?

The controller k8s client uses a cache and watchers for event based triggering. It *SHOULD* be able to handle the events as most API calls will hit the cache. We might want to consider some optimizations either way.

But testing the performance on the lab is a good idea.

pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (e *EventHandler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	event := &corev1.Event{}
+	if err := e.Get(ctx, req.NamespacedName, event); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	obj := &unstructured.Unstructured{}
```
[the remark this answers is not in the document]
TheBigLee: I can't use that one, because it is missing the `kind`, `apiVersion` and `group` fields.

cmd/controller.go
```diff
@@ -72,6 +73,18 @@ func (c *controller) executeController(cmd *cobra.Command, _ []string) error {
 	}
 
 	err = xpg.SetupWithManager(mgr)
+
```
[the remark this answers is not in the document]
TheBigLee: fixed

pkg/controller/events/event_forwarder.go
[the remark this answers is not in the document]
TheBigLee: Added basic documentation and some comments in the code.

pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,93 @@
...
+	obj.SetGroupVersionKind(schema.FromAPIVersionAndKind(event.InvolvedObject.APIVersion, event.InvolvedObject.Kind))
+	if err := e.Get(ctx, types.NamespacedName{Name: event.InvolvedObject.Name, Namespace: event.InvolvedObject.Namespace}, obj); err != nil {
+		return ctrl.Result{}, err
+	}
+	ef := obj.GetAnnotations()["vshn.net/forward-events-to"]
+	if ef != "" {
+		log.Info("Forwarding event", "event", event.Name, "to", ef)
+		knc := strings.Split(ef, "/")
+		a := knc[0] + "/" + knc[1]
```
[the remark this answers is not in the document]
TheBigLee: Fixed

pkg/controller/events/event_forwarder.go
TheBigLee: The idea with patching is, that we can leverage the two event properties: `firstTimestamp` and  `lastTimestamp`. With those two fields it's clear when it started and when it happened the last time.
This is only possible when patching the existing event. Otherwise, the user won't know if the event is actually still triggering or not.

pkg/controller/events/event_forwarder.go
TheBigLee: done

pkg/controller/events/event_forwarder.go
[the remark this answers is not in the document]
Kidswiss: It embeds the `metav1.TypeMeta` object which contains those fields and also implements `SetGroupVersionKind()`.

But as `unstructured.Unstructured` also exposes those it doesn't make that much of a difference.

pkg/controller/events/event_forwarder.go
TheBigLee: It depends. For helm based setups, this should be set in the composition by setting the annotations of the objects in the `helmValues`. (for Redis eg: https://github.com/vshn/component-appcat/pull/258
For provider specific setups (eg. PostgreSQL) this should most likely be done in the provider.

TheBigLee:
> also: this is still in Draft - intentionally?

No this was a mistake

pkg/controller/events/event_forwarder.go
Kidswiss: I was under the impression that each event is a distinct object and those fields are computed. But I'm not sure if that's really the case.

Does the update of the timestamp and counts work as expected with this solution? If yes it's fine, but if something is not working as intended, then we probably should use the `EventRecorder`, which will handle those details for us.

Kidswiss (changes requested):

pkg/controller/events/event_forwarder.go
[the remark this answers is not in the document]
TheBigLee: The `metadata.MetadataOnlyObject` doesn't have the `kind` registered in the scheme for the k8s client.

pkg/controller/events/event_forwarder.go
Kidswiss: As discussed, we'll use this solution for now, if we find any issues, we'll change to the recorder.

Kidswiss (approved):

docs/modules/ROOT/pages/explanations/eventforwarding.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+We use a controller to catch all events on a cluster.
+We check the owner object of every event for a specific annotation.
+
+The annotation contains the kind, namespace and name of a claim that should receive the event.
+
+
+== Annotation
+
+The annotation is called `vshn.net/forward-events-to` and has the following format: `<claim_kind>/<claim_namespace>/<claim_name>`.
```
zugao, line 13: I would add `appcat` word in here. We need to make sure this works only with appcat.

pkg/controller/events/event_forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (e *EventHandler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Event{}).
```
zugao, line 121: There may be many many events to be reconciled in a cluster. We might one to evaluate if we can reduce the amount of events being processed. Check method WithEventFilter(p predicate.Predicate) *Builder if we can do that.

zugao (approved):
Some things to check. Rest LGTM

zugao (changes requested):
Some things to look at

docs/modules/ROOT/pages/explanations/eventforwarding.adoc
TheBigLee: changed the annotation to `appcat.vshn.io/forward-events-to`

pkg/controller/events/event_forwarder.go
TheBigLee: Refactored the code to use the `WithEventFilter` to filter out a bunch of events from the get-go.

zugao (approved):
LGTM

Kidswiss (approved):

PR #103, opened 2023-11-20
Migrating to Composition Functions beta

Kidswiss:
## Summary
The beta had quite a lot of changes. For one the GRPC communication is now the way Crossplane communicates with functions.

Crossplane also provides an SDK which does a lot of heavy lifting right out of the box. However due to the way we handle things the official SDK is not sufficient. It has thus been wrapped in a `ServiceRuntime`.

Functions now have to be packaged in order to work. Some initial make targets have been added with this change. They can probably be improved over the next few weeks/months.

Additional changes:
* Naming and wording has been aligned to match Crossplane's upstream wording.
* The runtime is no longer divided between `desired` and `observed` sub- objects. (up to debate)
* The runtime mostly just tracks the desired changes and does marshalling of the objects. Everything else is off-loaded to the official go-sdk.
* Only metadata and status of the composite can be changed! No other changes are allowed anymore.

## Checklist

- [x] Update tests.


Kidswiss:
I will do docs in another PR.

apis/vshn/v1/dbaas_vshn_redis.go
```diff
@@ -151,3 +174,50 @@ type XVSHNRedisList struct {
 
 	Items []XVSHNRedis `json:"items"`
 }
+
+// GetMaintenanceDayOfWeek returns the currently set day of week
+func (n *VSHNRedis) GetMaintenanceDayOfWeek() string {
```
zugao, line 179: Can you abstract maintenance methods? Since the definition is the same across services.

pkg/comp-functions/functions/miniobucket/miniobucket.go
```diff
@@ -21,54 +20,47 @@ const (
...
+	err := svc.GetObservedComposite(bucket)
 	if err != nil {
-		return runtime.NewFatalErr(ctx, "Could not get ObjectBucket claim", err)
+		return runtime.NewFatalResult(err)
 	}
 
-	config, ok := iof.Config.Data["providerConfig"]
+	config, ok := svc.Config.Data["providerConfig"]
 	if !ok {
-		return runtime.NewFatalErr(ctx, "no providerConfig specified in composition config", fmt.Errorf("no providerConfig specified"))
```
zugao, line 35: Do I understand correctly, the error description is not relevant anymore?

zugao (approved):
This is huge, I haven't tested it but overall looks good. I have some comments/questions.

apis/vshn/v1/dbaas_vshn_redis.go
TheBigLee: Just had the same thought, when going through the code.

pkg/comp-functions/functions/miniobucket/miniobucket.go
Kidswiss: Yes, the new API just takes an error and no description.

If you want to add additional information to the error you can do it with standard golang error nesting.

Also, if you just want a default result you can just return nil, it handles a lot closer how vanilla error handling works.

apis/vshn/v1/dbaas_vshn_redis.go
Kidswiss: That was not my priority, let's leave that for a future improvement, this PR is already huge enough.

pkg/comp-functions/functions/vshnminio/maintenance.go
```diff
@@ -12,25 +14,27 @@ import (
...
-	err := iof.Observed.GetComposite(ctx, comp)
+	err := svc.GetObservedComposite(comp)
 	if err != nil {
-		return runtime.NewFatalErr(ctx, "can't get composite", err)
+		err = fmt.Errorf("cannot get observed composite: %w", err)
+		return runtime.NewFatalResult(err)
 	}
 
-	common.SetRandomSchedules(&comp.Spec.Parameters.Backup, &comp.Spec.Parameters.Maintenance)
+	common.SetRandomSchedules(comp, comp)
```
TheBigLee, line 26: I don't understand this change. The `SetRandomSchedules` function itself didn't change (or has not been committed).
The types of the function argument have changed here, so how can this still work?

pkg/comp-functions/functions/vshnminio/minio_deploy.go
```diff
@@ -236,32 +238,29 @@ func createServiceObserver(ctx context.Context, comp *vshnv1.VSHNMinio, iof *run
...
+		}
+		err = fmt.Errorf("cannot get observed connectiondetails: %w", err)
 		return err
 	}
 
-	iof.Desired.PutCompositeConnectionDetail(ctx, v1alpha1.ExplicitConnectionDetail{
-		Name:  "MINIO_URL",
-		Value: "http://" + service.Spec.ClusterIP + ":" + strconv.Itoa(int(service.Spec.Ports[0].Port)),
-	})
+	svc.SetConnectionDetail("MINIO_URL", []byte("http://"+service.Spec.ClusterIP+":"+strconv.Itoa(int(service.Spec.Ports[0].Port))))
```
TheBigLee, line 257:
```suggestion
	svc.SetConnectionDetail("MINIO_URL", []byte("http://" + service.Spec.ClusterIP + ":" + strconv.Itoa(int(service.Spec.Ports[0].Port))))
```
For readablity

pkg/comp-functions/functions/vshnminio/maintenance.go
Kidswiss: I've moved the interfaces from the `Parameters.Maintenance` and `Parameters.Backup` to the composition itself.

See: https://github.com/vshn/appcat/pull/103/files/eaf2230e7034f956b0e60a132447088c1662944f#diff-5fb6d956babb9d91192150922edcb8854865f4f84b2b338f744f0ee63483d53cR179

I could however wrap everything in yet another interface to get rid of passing the same object twice.

pkg/comp-functions/functions/vshnminio/minio_deploy.go
Kidswiss: `go fmt` says nope :)

PR #111, opened 2023-12-04
Remove any excess finalizers

Kidswiss:
## Summary

The controller now checks if there are more than 1 appcat finalizers on the composite and remove all except one if it's the case.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/controller/postgres/deletion_protection.go
+++ b/pkg/controller/postgres/deletion_protection.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/postgres/deletion_protection.go
```diff
@@ -83,32 +84,55 @@ func getRequeueTime(ctx context.Context, inst client.Object, deletionTime *metav
...
+	dupes := []int{}
 	for i, finalizer := range inst.GetFinalizers() {
 		if finalizer == finalizerName {
 			index = i
+			dupes = append(dupes, i)
 		}
 	}
 
+	// if this is a noop and we have exactly one or no dupe, then we're done here
+	if op == jsonpatch.JSONopNone && (len(dupes) == 1 || len(dupes) == 0) {
```
zugao, line 99: WHy do you expect to have exactly 2 finalizers before proceeding?

zugao: I thought you already had such check before but written differently, let's hope this works.

zugao (approved):

pkg/controller/postgres/deletion_protection.go
Kidswiss: The previous logic didn't really act on the amount of finalizers found apart from knowing which one to remove.

This here is basically the main check to see if we have multiple finalizers or not, if we do, then the rest of this function will determine that and clean them up.

So far all of the unit tests are succeeding and it worked on the lab.

PR #120, opened 2024-01-09
Improve maintenance detection

Kidswiss:
## Summary

* The SLI exporter now sets the maintenance label to true if a maintenance is occurring on the current cluster
* Maintenance status is directly gathered from the `UpgradeJob` CRD on OpenShift
  * There's a controller that watches for any changes in `UpgradeJob` CRs and tracks if the maintenance is running or not
  * The tracked data is provided to the probemanager which injects the new label accordingly.

These exceptions don't apply to HA instances. Those will still alert if the uptime metric is firing during maintenance, as that should not happen.

## Checklist

- [x] Update tests.


pkg/sliexporter/maintenance_controller/controller.go
```diff
@@ -0,0 +1,97 @@
...
+	maintenance atomic.Bool
+	client.Client
+}
+
+// MaintenanceStatus returns the current state of the maintenance.
+type MaintenanceStatus interface {
+	IsMaintenanceRunning() bool
+}
+
+// Reconcile reconciles on OpenShift upgradeJob objects and tracks if the it's currently running or not.
```
wejdross, line 30: that doesn't sound correct ;)

pkg/sliexporter/probes/manager.go
```diff
@@ -140,19 +146,21 @@ func (m Manager) runProbe(ctx context.Context, p Prober) {
...
 		"namespace":    pi.Namespace,
 		"name":         pi.Name,
 		"organization": pi.Organization,
 		"ha":           strconv.FormatBool(pi.HighAvailable),
 		"sla":          pi.ServiceLevel,
+		"maintenance":  strconv.FormatBool(m.maintenanceStatus.IsMaintenanceRunning()),
 	})
 	if err != nil {
-
+		l.Error(err, "failed to instanciate prmetheus histogram")
```
wejdross, line 161: another typo

wejdross (approved):
pretty nice PR, I love the simplicity, please fix typos, code-wise i'm happy

PR #122, opened 2024-01-11
Add MariaDB Backups

Kidswiss:
## Summary

* The backup is done via K8up and MariaBackup. The backup command is added via annotation to the pods. MariaBackup streams the backup to stdout, where K8up picks it up.
* Generalized a lot of logic which injects the annotations and scripts into any given helm values.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/backup/backup.go
+++ b/pkg/comp-functions/functions/common/backup/backup.go
@@ -1,0 +1,269 @@
+package backup
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	k8upv1 "github.com/k8up-io/k8up/v2/api/v1"
+	"github.com/sethvargo/go-password/password"
+	appcatv1 "github.com/vshn/appcat/v4/apis/v1"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/functions/common"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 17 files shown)


Kidswiss:
Docs for manual restores are located here: https://github.com/vshn/component-appcat/pull/301

pkg/comp-functions/functions/common/backup/backup.go
```diff
@@ -0,0 +1,267 @@
...
+	l := controllerruntime.LoggerFrom(ctx)
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      k8upRepoSecretName,
+			Namespace: comp.GetInstanceNamespace(),
+		},
+	}
+
+	err := svc.GetObservedKubeObject(secret, comp.GetName()+"-k8up-repo-pw")
```
TheBigLee, line 88: Wouldn't it make sense to put the suffix in a variable or const as well to not duplicate it three times?
Or even the whole name of the object.

pkg/comp-functions/functions/common/backup/backup.go
```diff
@@ -0,0 +1,267 @@
...
+	pw, err := password.Generate(64, 5, 5, false, true)
+	if err != nil {
+		return err
+	}
+
+	secret.Data = map[string][]byte{
+		k8upRepoSecretKey: []byte(pw),
+	}
+
+	return svc.SetDesiredKubeObject(secret, comp.GetName()+"-k8up-repo-pw")
```
TheBigLee, line 107: Why are you not using the `AddCredentialsSecret` function you created for this purpose?

Kidswiss: I just refactored the code without changing too much of the logic.

I like the idea, this way we can save one more object.

TheBigLee (approved):

PR #127, opened 2024-01-24
Don't handle the organization label in quotas

Kidswiss:
## Summary
The quota handling still assumes that the org label is patched onto the composite, but that isn't true anymore. Also the organization creation is handled in the `BootstrapInstanceNs()` function. This generates an racecondition between those two functions.

Additionally:
* Don't return fatal in the quota function, this can lead to deadlocks
* Cleaned up some dropped errors

## Checklist

- [x] Update tests.


TheBigLee (approved):

wejdross (approved):

Kidswiss:
Tested on the lab. Works so far, the org label is set and the quotas as well.

PR #129, opened 2024-02-02
Add automatic maintenance to Keycloak

Kidswiss:
## Summary

This commit adds automatic maintenance for Keycloak.
    
The automatic maintenance now supports private registries. It also now     supports any other registry apart from docker hub.
    
We've been using the docker hub's rich API until now. But that is exclusive to docker hub. Normal v2 registries have another API. The maintenance now supports that API as well and will automatically fall back to it, if the URL isn't from `https://hub.docker.com`.

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_keycloak.go
+++ b/apis/vshn/v1/dbaas_vshn_keycloak.go
@@ -84,11 +84,11 @@ type VSHNKeycloakServiceSpec struct {
 	// +kubebuilder:default="/"
 	RelativePath string `json:"relativePath,omitempty"`
 
-	// +kubebuilder:validation:Enum="23"
-	// +kubebuilder:default="23"
+	// +kubebuilder:validation:Enum="23.0.5-202402021353-44-af6cea11"
+	// +kubebuilder:default="23.0.5-202402021353-44-af6cea11"
 
 	// Version contains supported version of keycloak.
-	// Multiple versions are supported. The latest version 22 is the default version.
+	// Multiple versions are supported. The latest version 23 is the default version.
 	Version string `json:"version,omitempty"`
 
 	// +kubebuilder:validation:Enum="besteffort";"guaranteed"
--- a/cmd/maintenance.go
+++ b/cmd/maintenance.go
@@ -10,6 +10,7 @@ import (
 	"github.com/thediveo/enumflag/v2"
 	"github.com/vshn/appcat/v4/pkg"
 	"github.com/vshn/appcat/v4/pkg/maintenance"
+	"github.com/vshn/appcat/v4/pkg/maintenance/auth"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -27,6 +28,7 @@ const (
 	redis
 	minio
 	mariadb
+	keycloak
 )
 
 var maintenanceServices = map[service][]string{
@@ -34,6 +36,7 @@ var maintenanceServices = map[service][]string{
 	redis:      {"redis"},
 	minio:      {"minio"},
 	mariadb:    {"mariadb"},
+	keycloak:   {"keycloak"},
 }
 
 var serviceName service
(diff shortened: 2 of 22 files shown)


apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -84,11 +84,11 @@ type VSHNKeycloakServiceSpec struct {
 	// +kubebuilder:default="/"
 	RelativePath string `json:"relativePath,omitempty"`
 
-	// +kubebuilder:validation:Enum="23"
-	// +kubebuilder:default="23"
+	// +kubebuilder:validation:Enum="23.0.5-202402021353-44-af6cea11"
```
Kidswiss, line 87: I've asked them, if they can multitag the images, this looks like a great source for user errors.

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -216,13 +230,115 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
...
+			"start",
+			"--http-enabled=true",
+			"--http-port=8080",
+			"--hostname-strict=false",
+			"--hostname-strict-https=false",
+			"--spi-events-listener-jboss-logging-success-level=info",
+			"--spi-events-listener-jboss-logging-error-level=warn",
+		},
+		"image": map[string]any{
+			"repository": "docker-registry.inventage.com:10121/keycloak-competence-center/keycloak-managed",
```
TheBigLee, line 326: Would it make sense, to make this configurable and not hardcoded?

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -235,43 +351,21 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
...
-			},
-		},
-		"extraVolumes": []map[string]any{
-			{
-				"name": "postgresql-certs",
-				"secret": map[string]any{
-					"secretName":  pgSecretName,
-					"defaultMode": 420,
+		// See https://github.com/keycloak/keycloak/issues/11286
+		// readOnlyRootFilesystem: true
```
TheBigLee, line 358: Why is this field even there and commented out?

pkg/maintenance/keycloak.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"net/http"
+
+	"github.com/go-logr/logr"
+	"github.com/vshn/appcat/v4/pkg/maintenance/helm"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	keycloakURL = "https://docker-registry.inventage.com:10121/v2/keycloak-competence-center/keycloak-managed/tags/list"
```
TheBigLee, line 13: same here. Should this really be hardcoded?

pkg/comp-functions/functions/vshnkeycloak/deploy.go
Kidswiss: I copy pasted it over from Inventages values.

Might be something to change once the bug is resolved.

pkg/comp-functions/functions/vshnkeycloak/deploy.go
Kidswiss: Did Inventage say anything about changing the repository?

We also have hardcoded charts/images for other services though.

TheBigLee: No they didn't. I just thought it might make it easier to change it if we have to. But fine by me to leave it as is.

pkg/comp-functions/functions/vshnkeycloak/deploy.go
TheBigLee: Ok. Fine by me

TheBigLee:
Don't forget to update the CRDs as well

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -235,43 +351,21 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
...
 		},
 		"nodeSelector": nodeSelector,
-		"metrics": map[string]any{
+		"dbchecker": map[string]any{
 			"enabled": true,
-			"serviceMonitor": map[string]any{
-				"enabled": true,
-			},
-		},
-		"httpRelativePath": comp.Spec.Parameters.Service.RelativePath,
```
TheBigLee, line 244: This should not be removed, as the upstream Relative path is set by default to `/auth`

Kidswiss: It's set here: https://github.com/vshn/appcat/pull/129/files#diff-9d57e64a014aff26a646229e449affc4afe8e96fc5238920a77758f105cddc14R368

cmd/maintenance.go
```diff
@@ -83,17 +86,30 @@ func (c *controller) runMaintenance(cmd *cobra.Command, _ []string) error {
...
 }
+
+func getHTTPClient() *http.Client {
+	if viper.GetString("REGISTRY_USERNAME") == "" || viper.GetString("REGISTRY_PASSWORD") == "" {
+		return &http.Client{
+			Timeout: 30 * time.Second,
+		}
+	}
+	return auth.GetAuthHTTPClient(viper.GetString("REGISTRY_USERNAME"), viper.GetString("REGISTRY_PASSWORD"))
+}
```
wejdross, line 115: shouldn't that be rather a secret?

Kidswiss: What do you mean by secret? The credentials are passed to the maintenance pod via env vars, we're reading them here again.

> so we allow just one image version? that won't work - we have request for different minor images right now, we can expect them later, maybe leave default and delete Enum?

Inventage currently only provides this image to us. So yes we only have this one at the moment. It can be extended in the future.

apis/vshn/v1/dbaas_vshn_keycloak.go
Kidswiss: Inventage currently only provides this image to us. So yes we only have this one at the moment. It can be extended in the future. We're in contact with Inventage.

This is currently still very early stage and many changes will come soon.

wejdross: ok, fine for me

wejdross (approved):

PR #132, opened 2024-02-15
block too long db names early

wejdross:
## Summary

* Adding logic to catch too long Redis and PostgreSQL names early
* We add various postfixes to resource name and our limitation is https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names for cronjob resources 
* if we allow resource creation then VSHNRedis is created but underlying objects not, therefore customer would wait indefinitely for database to come up

## Checklist

- [ ] Update tests.

### DIFF
--- a/crds/stackgres.io_sgpoolingconfigs.yaml
+++ b/crds/stackgres.io_sgpoolingconfigs.yaml
@@ -32,7 +32,7 @@ spec:
           metadata:
             type: object
           spec:
-            description: Spec defines the desired state of a VSHNPostgreSQL.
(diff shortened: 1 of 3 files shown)


crds/stackgres.io_sgpoolingconfigs.yaml
```diff
@@ -32,7 +32,7 @@ spec:
           metadata:
             type: object
           spec:
```
wejdross, line 34: those lines were added by `generate` target, unrelated to changes in this PR, but they're so small I decided to leave them

DebakelOrakel (approved):

PR #135, opened 2024-03-04
Add external admin account

Kidswiss:
## Summary

This commit adds an admin account which is intended to be exposed to the user. This enables the user to change the password of the admin user without breaking the internal scripts of the image.

However, there are no safeguards available to prevent the user from changing the password of the internal admin as well.

## Checklist

- [x] Update tests.


Kidswiss:
There are also two commits that change the image tags, as Inventage now provides shorter tags as well.

zugao (approved):
LGTM

PR #136, opened 2024-03-05
Ability to disable billing for selected services

Kidswiss:
## Summary

With this commit it's possible to disable the billing for selected services.
    
It works by deploying a separate config map containing this information in the `appcat-control` namespace. This config map is then read by the service composition function and removes the billing accordingly.
    
For more detailed information please have a look at the docs in the commit.

## Checklist

- [x] Update tests.

### DIFF
--- a/docs/modules/ROOT/pages/explanations/disable-billing.adoc
+++ b/docs/modules/ROOT/pages/explanations/disable-billing.adoc
@@ -1,0 +1,23 @@
+= How Disabling Billing Works
+
+For services that depend on other services, like Keycloak -> PostgreSQL, we don't want to bill the customer with both services.
+
+In order to achieve that, we need a mechanism with which we can disable the billing for any given service, but in such a way that it's not possible for the user to manipulate it.
+
+As all settings between the claim and the composite are synced, we can't expose any configuration regarding this in the claim. Otherwise, the user would be able to simply disable the billing for any service. In order to prevent that, the configuration to disable billing should be passed via means that are inaccessible for the end-user but accessible to the composition functions.
+
+There's already a namespace for such things: `appcat-control`. That's a namespace specifically used for things that the end-user should not be able to see or manipulate.
+
+== Flow
+
+When a dependency service is deployed, which should not have any billing, then the service depending on it should deploy a specific config map at the same time. This config map should have the same name as the instance namespace of the dependency service. It should also be deployed in the `appcat-control` namespace. This config map should contain a key called billingDisabled. If that key is not set, the billing is enabled as usual.
+
+*Note:* due to some limitations of Crossplane, each service will always create an empty such config map. Because it's not possible to determine if an object exists or not via provider-kubernetes.
+
+image::disable-billing-flow.drawio.svg[]
+
+== Code
+
+There's a function `DisableBilling(namespace string, svc *runtime.ServiceRuntime)` which will create the config map for the given instance namespace. This needs to be called by the parent composition function, in the example above Keycloak, at the same time the dependency (PostgreSQL) composite is applied.
+
+On the dependency side, in our example PostgreSQL, nothing special needs to be done. The `BootstrapInstanceNs()` automatically determines if the config map exists and if the key to disable the billing is set. If the config map doesn't exist it will be automatically created. This ensures that provider-kubernetes can actually resolve the observed object.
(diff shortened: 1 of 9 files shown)


docs/modules/ROOT/pages/explanations/disable-billing.adoc
```diff
@@ -0,0 +1,23 @@
+= How Disabling Billing Works
+
+For services that depend on other services, like Keycloak -> PostgreSQL, we don't want to bill the customer with both services.
+
+In order to achieve that, we need a mechanism with which we can disable the billing for any given service, but in such a way that it's not possible for the user to manipulate it.
+
+As all settings between the claim and the composite are synced, we can't expose any configuration regarding this in the claim. Otherwise, the user would be able to simply disable the billing for any service. In order to prevent that, the configuration to disable billing should be passed via means that are inaccessible for the end-user but accessible to the composition functions.
+
+There's already a namespace for such things: `appcat-control`. That's a namespace specifically used for things that the end-user should not be able to see or manipulate.
```
zugao, line 9: Don't we already have `syn-appcat `namespace? Is it the same namespace?

pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -93,6 +94,19 @@ func createInstanceNamespace(ctx context.Context, serviceName, compName, claimNa
...
+	if !ok {
+		return fmt.Errorf("controlNamespace not specified")
+	}
+
+	disabled, err := isBillingDisabled(controlNS, instanceNamespace, compName, svc)
+	if err != nil {
+		return err
+	}
+	if disabled {
+		ns.ObjectMeta.Labels["appuio.io/billing-name"] = ""
```
zugao, line 107: Are we sure the label will not be changed in the future? It would be a good idea to dynamically get these labels if possible instead of hard coding them here.

pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -139,3 +153,70 @@ func createNamespacePermissions(ctx context.Context, instance string, instanceNs
 	}
 	return svc.SetDesiredKubeObjectWithName(roleBinding, instance+"-service-rolebinding", "namespace-permissions")
 }
+
+// DisableBilling deploys a special config map to the appcat control namespace.
```
zugao, line 157: Wouldn't it be easier to understand if we would have only one CM with a list of disabled namespaces? I feel like having a CM per each disabled instance is a bit of an overkill

pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -139,3 +153,70 @@ func createNamespacePermissions(ctx context.Context, instance string, instanceNs
...
+		return err
+	}
+
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      instanceNamespace,
+			Namespace: controlNS,
+		},
+		Data: map[string]string{
+			billingCMKey: "true",
```
zugao, line 180: This sets as disabled? If yes name of the key is confusing.

zugao (commented):
There are a couple of things to check

docs/modules/ROOT/pages/explanations/disable-billing.adoc
Kidswiss: `syn-appcat` is where appcat itself is deployed. `appcat-control` is where some of the jobs run with clusterrolebindings, that a customer should not be able to access at all. Mostly restore and maintenance jobs.

pkg/comp-functions/functions/common/instance_namespace.go
Kidswiss: We fully control the queries for the billing ourselves, see here: https://github.com/vshn/component-appcat/tree/master/component/component/promql

Chaning these will need a lot of consideration, planing and testing, so I don't think dynamically passing them dynamically would really give a big benefit.

pkg/comp-functions/functions/common/instance_namespace.go
Kidswiss: There are various things that would make it a lot harder. It would be somewhat like your URO issue. The main problem would also be to clean up the CM again, if an instance gets deleted. As soon as a claim/composite has a `deletionTimestamp` Crossplane will not trigger the compositions anymore. So we would need some external controller again to actually make sure that the CM is properly synced up with the state of the all the instances.

Also, we could re-use the CM in the future to support other user hidden settings, if they come up.

pkg/comp-functions/functions/common/instance_namespace.go
Kidswiss: Do you mean the name of the constant `billingCMKey`? Or the value of the constant `billingDisabled `?

zugao: The constant, you might consider changing the name.

zugao (approved):
LGTM

PR #140, opened 2024-03-08
Mitigate namespace deletion race-condition

Kidswiss:
## Summary

* There was a race condition which could re-create the instance namespace during the upgrade of the appcat function
* Also fixed the tests that broke during a hotfix deployment yesterday

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/instance_namespace.go
+++ b/pkg/comp-functions/functions/common/instance_namespace.go
@@ -101,7 +101,9 @@ func createInstanceNamespace(ctx context.Context, serviceName, compName, claimNa
 
 	disabled, err := isBillingDisabled(controlNS, instanceNamespace, compName, svc)
 	if err != nil {
-		return err
+		// we don't return here, otherwise we risk the namespace getting deleted
+		svc.Log.Error(err, "cannot determine billing status of the service")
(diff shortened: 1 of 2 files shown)


pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -101,7 +101,9 @@ func createInstanceNamespace(ctx context.Context, serviceName, compName, claimNa
 
 	disabled, err := isBillingDisabled(controlNS, instanceNamespace, compName, svc)
 	if err != nil {
-		return err
+		// we don't return here, otherwise we risk the namespace getting deleted
```
wejdross, line 104: can You further explain? so I understand whole picture in detail

zugao (approved):
LGTM please deploy as soon as possible

pkg/comp-functions/functions/common/instance_namespace.go
Kidswiss: Was the explanation during the daily on Friday enough?

wejdross: yes of course

wejdross (approved):

PR #142, opened 2024-03-12
Fix Keycloak securityContext for OpenShift

Kidswiss:
## Summary

* The helm chart sets some securityContext by default, this PR removes that

## Checklist

- [x] Update tests.


zugao (commented):
Will this work in both k8s and OpenShift?

Kidswiss:
> Will this work in both k8s and OpenShift?

Yes, it's tested on kind and currently also rolled out ot the lab.

zugao (approved):
LGMT

PR #144, opened 2024-03-21
Add ingress and internal TLS to Keycloak

Kidswiss:
This commit enables TLS on the https port in Keycloak's service. So the service can be used via https internally.

Additionally, if the FQDN is set in the claim, it will also enable ingress via the https service.

## Summary

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnkeycloak/ingress.go
```diff
@@ -0,0 +1,94 @@
...
+		return runtime.NewWarningResult(fmt.Sprintf("cannot set desired release: %s", err))
+	}
+
+	return nil
+}
+
+func enableIngresValues(comp *vshnv1.VSHNKeycloak, values map[string]any) {
+	fqdn := comp.Spec.Parameters.Service.FQDN
+
+	if fqdn != "" {
```
zugao, line 61: Didn't you already tested for empty value in line 25 ?

Kidswiss: hmmm... yeah I guess it's redudant here. I'll remove it.

zugao (approved):
LGTM

PR #145, opened 2024-03-25
Implement annotation handling for the ingress

Kidswiss:
## Summary

* Some settings for the ingress are managed via annotations
  * For example what certificates to use or some OpenShift specifics
* We can inject the annotations via the inputs
* Fixed a bug with the merging of the backup settings
* Some more OpenShift specifics so that the connection between ingress and Keycloak is also using HTTPS

## Checklist

- [x] Update tests.


wejdross (approved):
Please test it well on LAB, code-wise is OK for me. LGTM

PR #146, opened 2024-04-05
injecting minimalistic postgresql-exporter configuration file

wejdross:
Stackgres uses very outdated postgresql-exporter which has by default enabled tons of telemetry scraping. Such scrapes tends to OOM for bigger databases, with this PR I was able to limit memory usage for database (100 tables each having 100 columns and each column having 1000 rows) to around ~11-12 MB of `container_memory_usage_bytes{}` and ~10-11 MB of rss. It SHOULD be backward compatible with our Grafana dashboards. 

Fun fact:
queries were reduced by ~800 lines :zany_face: 

## Checklist

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/pgqexporter_workaround.go
+++ b/pkg/comp-functions/functions/vshnpostgres/pgqexporter_workaround.go
@@ -2,21 +2,26 @@ package vshnpostgres
 
 import (
 	"context"
+	_ "embed"
 	"fmt"
 
-	xkubev1 "github.com/crossplane-contrib/provider-kubernetes/apis/object/v1alpha1"
+	xkube "github.com/crossplane-contrib/provider-kubernetes/apis/object/v1alpha1"
 	"github.com/crossplane/function-sdk-go/proto/v1beta1"
 	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	controllerruntime "sigs.k8s.io/controller-runtime"
 )
 
+/*
+	This code ensures we have minimalistic postgresql-exporter configuration
+	It prevents high memory usage for big databases
+*/
+
+//go:embed files/queries.yml
+var queries string
+
 func PgExporterWorkaround(ctx context.Context, svc *runtime.ServiceRuntime) *v1beta1.Result {
-	log := controllerruntime.LoggerFrom(ctx)
 
 	comp := &vshnv1.VSHNPostgreSQL{}
 	err := svc.GetObservedComposite(comp)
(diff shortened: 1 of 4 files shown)


pkg/comp-functions/functions/vshnpostgres/pgqexporter_workaround.go
```diff
@@ -0,0 +1,566 @@
...
+	configMap.Data["queries.yaml"] = queries
+	_, err = clientset.CoreV1().ConfigMaps(comp.GetNamespace()).Update(context.Background(), configMap, metav1.UpdateOptions{})
+	if err != nil {
+		return runtime.NewWarningResult(fmt.Sprintf("cannot update configmap: %s", err))
+	}
+
+	return nil
+}
+
+const queries = `
```
Kidswiss, line 69: Can you please put this in a file and use go embed? It will make this a million times more manageable.

See an example here: https://github.com/vshn/appcat/blob/master/pkg/comp-functions/functions/vshnmariadb/backup.go#L20

pkg/comp-functions/functions/vshnpostgres/pgqexporter_workaround.go
```diff
@@ -0,0 +1,566 @@
...
+	obj := &xkubev1.Object{}
+	err = svc.GetObservedComposedResource(obj, "cluster")
+	if err != nil {
+		return runtime.NewWarningResult(fmt.Sprintf("cannot get the sgcluster object: %s", err))
+	}
+
+	// get configmap
+	configMap := &corev1.ConfigMap{}
+	// create k8s client inside cluster
+	config, err := rest.InClusterConfig()
```
Kidswiss, line 37: Please use the kube objects here. This adhoc k8s client will bite us in the ass if we ever want to do something on remote clusters.

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnpostgres/pgqexporter_workaround.go
```diff
@@ -0,0 +1,61 @@
...
+
+/*
+	This code ensures we have minimalistic postgresql-exporter configuration
+	It prevents high memory usage for big databases
+*/
+
+//go:embed files/queries.yml
+var queries string
+
+func PgExporterWorkaround(ctx context.Context, svc *runtime.ServiceRuntime) *v1beta1.Result {
```
Kidswiss, line 24: Maybe call it `PgExporterConfig`?

pkg/comp-functions/functions/vshnpostgres/register.go
```diff
@@ -69,6 +69,10 @@ func init() {
 				Name:    "pgbouncer-settings",
 				Execute: addPGBouncerSettings,
 			},
+			{
+				Name:    "pg-exporter-Workaround",
```
Kidswiss, line 73: Same here.

wejdross: done

wejdross:
> One more thing: did you verify that your test database actually breaks the exporter with the default config?

nope, we have numerous examples of that in our clusters

Kidswiss:
> > One more thing: did you verify that your test database actually breaks the exporter with the default config?
> 
> nope, we have numerous examples of that in our clusters

Sure but those are production databases on other clusters. They have probably more load and connections than your test database, so it's not quite the same situation.

If we don't know that your test database OOMs with the default exporter settings, how can we know that the fix is actually working?

Kidswiss (approved):
As you can't reproduce the OOM in a test env, let's roll it out and see.

If it still doesn't work, we'll do another round of optimizations.

PR #147, opened 2024-04-08
Add Keycloak SLI exporter

Kidswiss:
## Summary

* Adds Keycloak SLI exporter via https
* Introduce a generic http prober
* Refactor reconciler logic, now all services use the same reconciler
  * Fixed some bugs discovered while refactoring 

## Checklist

- [x] Update tests.


wejdross (approved):
Reconcile() abstraction was really necessary for this codebase.
LGTMG

pkg/sliexporter/probes/http.go
```diff
@@ -0,0 +1,72 @@
...
+
+	l := log.FromContext(ctx).WithValues("http_prober", h.url)
+
+	l.V(1).Info("Starting get request")
+	resp, err := h.client.Get(h.url)
+	if err != nil {
+		return err
+	}
+
+	l.V(1).Info("Checking response")
```
wejdross, line 66: Maybe log level could be higher here?

Kidswiss: No, I'd like to keep the clean logging in the SLI export we currently have.

It will only log a few specific things:
* when a probe starts
* when a probe stops
* when a probe fails

If we have more than a trivial amount of Keycloaks the amount of logs it generates would not be helpful anymore.

PR #151, opened 2024-04-12
Webhook based deletion protection for backend objects

Kidswiss:
## Summary

Certain objects are now protected by a webhook, as long as the parent composite still exists and has the protection enabled. The protected objects are:
    
* Instance Namespaces
* PVCs of an instance
* XObjectBucket of an instance
    
This commit does not implement protection on the claim/composite itself. That's for a future implementation.

## Checklist

- [x] Update tests.


Kidswiss:
Please read the docs, before diving into the code, it should help with understanding what it's doing.

pkg/comp-functions/functions/vshnpostgres/objectbucket.go
```diff
@@ -0,0 +1,36 @@
...
+
+	comp := &vshnv1.VSHNPostgreSQL{}
+	err := svc.GetObservedComposite(comp)
+	if err != nil {
+		return runtime.NewFatalResult(fmt.Errorf("Cannot get composite from function io: %w", err))
+	}
+
+	bucket := &appcatv1.XObjectBucket{}
+
+	err = svc.GetDesiredComposedResourceByName(bucket, "pg-bucket")
```
zugao, line 25: We could add a wrapper function here, it's confusing to read this code, you get and the set the same object :)

pkg/controller/webhooks/deletionprotection.go
```diff
@@ -0,0 +1,123 @@
...
+	if !ok {
+		return compositeInfo{Exists: false, Name: ownerName}, fmt.Errorf("cannot determine deletion protection state of: %s", ownerName)
+	}
+
+	return compositeInfo{Exists: dpi.IsDeletionProtected(), Name: ownerName}, nil
+}
+
+// checkUnmanagedObject tries to get the composite information about objects that are not directly managed by Crossplane.
+// It checks if the namespace it belongs to is managed by a composite, if that's the case it uses the same logic to
+// determine the state of the deletionprotection.
```
zugao, line 108:
```suggestion
// determine the state of the deletion protection.
```

pkg/controller/webhooks/deletionprotection.go
```diff
@@ -0,0 +1,123 @@
...
+// DeletionProtectionInfo provides information about the composite's deletion
+// protection state.
+type DeletionProtectionInfo interface {
+	IsDeletionProtected() bool
+}
+
+// checkManagedObject will find the highest composite for any object that is deployed via Crossplane or provider-kubernetes.
+// For example: XObjectBucket, Release, Namespace, etc.
+// Anything that either contains an owner reference to a composite or an owner annotation.
+func checkManagedObject(ctx context.Context, obj client.Object, c client.Client, l logr.Logger) (compositeInfo, error) {
```
zugao, line 34: This function can halt deletion of namespaces and PVCs in the whole cluster if I understood it right, please test it throughly.

zugao (approved):
LGTM, there are some small things. Please test the critical function throughly

apis/vshn/v1/common_types.go
```diff
@@ -9,6 +9,11 @@ type K8upBackupSpec struct {
 	Schedule string `json:"schedule,omitempty"`
 
 	Retention K8upRetentionPolicy `json:"retention,omitempty"`
+
+	// DeletionProtection will protect the instance's data from being deleted while the composite or claim still exists.
+	// This is enabled by default.
+	// +kubebuilder:default=true
+	DeletionProtection *bool `json:"deletionProtection,omitempty"`
```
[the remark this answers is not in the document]
Kidswiss: Yes. Having multiple deletion protection flags will only be confusing for the customer.

Although the only service that currently has that is PostgreSQL and that implementation is not very good. I'd like to have the webhook based deletion protection for the claims rather sooner than later. I don't want to wait with that until we have an incident...

pkg/comp-functions/functions/vshnpostgres/objectbucket.go
Kidswiss: I don't quite follow. How would a wrapper make this more readable?

This function is only necessary as long as we still have the PostgreSQL PnT composition, afterward it will be obsolete.

pkg/controller/webhooks/deletionprotection.go
Kidswiss: Ongoing ;)

apis/vshn/v1/common_types.go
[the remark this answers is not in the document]
zugao: Why do we have to expose this Deletion Protection to the customer, this level of protection with webhooks should be abstracted away from customer all together, don't you thing so?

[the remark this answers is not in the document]
Kidswiss: An override to temporarily disable the backend deletion protection will be useful. For example if we need to reinitialize Galera Clusters.

Having an override on the claim inevitably exposes it to the customer. So my though was, that we put it under the same key that the customer can use to control the frontend deletion protection in the future.

This makes it easier to understand and leaves less room for miss configuration. The protection is either on or off.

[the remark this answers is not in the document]
zugao: This approach means that we do not have our base data loss protection when the customer disables deletion protection. This is wrong in my opinion, the customer would not want in any case that his data is less protected even when he disables front end deletion protection. I would not expose the backend deletion protection to the customer and would enable it by default all the time.

pkg/comp-functions/functions/vshnpostgres/objectbucket.go
zugao: Never mind, it's ok like that. The function description is enough for now.

zugao (commented):
We need to discuss further the exposure of the deletion protection before merging.

apis/vshn/v1/common_types.go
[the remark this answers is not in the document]
Kidswiss: > I would not expose the backend deletion protection to the customer and would enable it by default all the time.

The problem here is: it could get pretty messy to have an override for the deletion protection that is not exposed to the customer as well. See the billing override. I would rather not have that much complexity for that, as it could increase the chances of failure.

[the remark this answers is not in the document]
zugao: Let's discuss that during the daily. I feel I am missing something here.

Kidswiss:
> We need to discuss further the exposure of the deletion protection before merging.

As discussed, I've changed the approach for the override, please have another look.

docs/modules/ROOT/pages/explanations/webhook-protection.adoc
```diff
@@ -0,0 +1,56 @@
...
+* For nested services, like Keycloak, the ownerReferences only point to the next composite in the chain. If for any reason one of the parent composites is missing, we can't get to the parent anymore.
+
+=== Override deletion protection
+Sometimes it's necessary to delete some PVC even if the instance still exists.
+For example if one node of a Galera cluster gets corrupted and needs to be re-initialized.
+
+To achieve this, following label has to be set on that object.
+That label will override the protection for exactly that object.
+
+`appcat.vshn.io/webhook-allowdeletion: "true"`
```
zugao, line 54: This applies to all the protected objects am I right? PVC Namespace and object bucket.

Kidswiss: Exactly.

PR #152, opened 2024-04-24
Fix loadbalancer issues

Kidswiss:
## Summary

* The loadbalancer logic still relied on the old connection detail secret, that has been removed 
* This prevented the service to ever get synced/ready, as it was erroring while looking for the secret

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnminio/securitycontext.go
```diff
@@ -0,0 +1,34 @@
+package vshnminio
```
Kidswiss, line 1: WTF, this is not supposed to be here :D

PR #155, opened 2024-04-25
Fix typo preventing the correct api being called

Kidswiss:
## Summary

There was a typo in the api detection which lead to chosing the wrong API, thus breaking the maintenance.
## Checklist

- [x] Update tests.


wejdross (approved):
OMG XD
LGTMT

PR #156, opened 2024-04-26
Fix custom mariadb settings

zugao:
## Summary

* This PR will make sure that custom user MariaDB settings are being applied along with the default settings for Galera.

## Checklist

- [x] Update tests.


Kidswiss (approved):
LGTM

Did you test that we can actually set custom settings now?

zugao:
> Did you test that we can actually set custom settings now?

Of course, with and without :)

PR #158, opened 2024-04-30
Support for both Vacuum and Repack

wejdross:
## Summary

* This PR gives possibility to set vacuum and/or repack to be run during maintenance

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_postgresql.go
+++ b/apis/vshn/v1/dbaas_vshn_postgresql.go
@@ -148,10 +148,9 @@ type VSHNPostgreSQLServiceSpec struct {
 	// PgBouncerSettings passes additional configuration to the pgBouncer instance.
 	PgBouncerSettings *sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni `json:"pgBouncerSettings,omitempty"`
 
-	// +kubebuilder:validation:Enum=true;false
 	// +kubebuilder:default=true
+	// This is default option if neither repack or vacuum are selected
 	RepackEnabled bool `json:"repackEnabled,omitempty"`
-	// +kubebuilder:validation:Enum=true;false
 	// +kubebuilder:default=false
 	VacuumEnabled bool `json:"vacuumEnabled,omitempty"`
 }
--- a/apis/vshn/v1/dbaas_vshn_postgresql.go
+++ b/apis/vshn/v1/dbaas_vshn_postgresql.go
@@ -147,6 +147,13 @@ type VSHNPostgreSQLServiceSpec struct {
 
 	// PgBouncerSettings passes additional configuration to the pgBouncer instance.
 	PgBouncerSettings *sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni `json:"pgBouncerSettings,omitempty"`
+
+	// +kubebuilder:validation:Enum=true;false
+	// +kubebuilder:default=true
+	RepackEnabled bool `json:"repackEnabled,omitempty"`
+	// +kubebuilder:validation:Enum=true;false
+	// +kubebuilder:default=false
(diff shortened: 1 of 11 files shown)


Kidswiss (approved):

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -144,6 +144,13 @@ type VSHNPostgreSQLServiceSpec struct {
 
 	// PgBouncerSettings passes additional configuration to the pgBouncer instance.
 	PgBouncerSettings *sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni `json:"pgBouncerSettings,omitempty"`
+
+	// +kubebuilder:validation:Enum=true;false
+	// +kubebuilder:default=true
+	RepackEnabled bool `json:"repackEnabled,omitempty"`
+	// +kubebuilder:validation:Enum=true;false
```
zugao, line 151: Why do you need a validation enum for a boolean value?

pkg/maintenance/postgresql.go
```diff
@@ -121,10 +122,33 @@ func (p *PostgreSQL) DoMaintenance(ctx context.Context) error {
...
+	}
+
+	if p.Repack {
+		p.log.Info("Repacking databases...")
+		err = p.createRepack(sgCluster.GetName())
+		if err != nil {
+			return fmt.Errorf("cannot create repack: %v", err)
+		}
+	}
+	// default to repack if for some reason it was possible to disable both vacuum and repack
```
zugao, line 146: We have to add this important information into the description of our repack field in the CRD.

zugao (changes requested):
Some things to take a look at.

apis/vshn/v1/dbaas_vshn_postgresql.go
wejdross: fixed

pkg/maintenance/postgresql.go
wejdross: fixed

wejdross:
I'd like to ask You for re-review as while implementing e2e tests I realized there was (pretty dumb) bug&race condition which is now fixed

Kidswiss (approved):

wejdross:
go tests - passed,
e2e in kind cluster - passed,
lab revealed other issues :face_with_head_bandage:  and I think that if we want to stay sane during operations we need to implement that as we do our helm based maintenance, I will come back to You later with ready change

wejdross:
yet another approach, this solves all of the issues, manual tests in kindev work like a charm, currently fixing e2e in component-appcat....

pkg/comp-functions/functions/vshnpostgres/maintenance.go
```diff
@@ -116,6 +127,9 @@ func addSchedules(ctx context.Context, svc *runtime.ServiceRuntime) *xfnproto.Re
 		return runtime.NewFatalResult(fmt.Errorf("cannot get cluster object: %w", err))
 	}
 
+	extraEnvVars[3].Value = strconv.FormatBool(comp.Spec.Parameters.Service.RepackEnabled)
```
Kidswiss, line 130: You could use `append()` and assign the array to a new variable to add new envVars instead of manipulating some hard coded array indices.

```suggestion
	additionalVars := append(extraEnvVars, corev1.EnvVar(..))

  // line 145
  WithExtraEnvs(additionalVars...)
```

That will make it more readable and easiert to maintain.

Kidswiss (changes requested):
Just one thing, rest LGTM

pkg/comp-functions/functions/vshnpostgres/maintenance.go
wejdross: implemented and tested it, looks fine

Kidswiss (approved):

PR #159, opened 2024-05-01
Add allow namespaces to all services

zugao:
## Summary

* This PR adds custom NS access from claim to all our services.
* This PR adds a small refactoring and moves the netpol creation to comp-func inside the boostrap function.

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -48,6 +48,17 @@ func BootstrapInstanceNs(ctx context.Context, comp Composite, serviceName, names
 		return fmt.Errorf("cannot create rbac rules for %s instance: %w", serviceName, err)
 	}
 
+	l.Info("Creating namespace policy to allow access to " + serviceName + " instance")
+	sourceNS := append(comp.GetAllowedNamespaces(), comp.GetClaimNamespace())
+	// Was only available for MariaDB though if it's not enabled, no new namespaces will be added
+	if svc.GetBoolFromCompositionConfig("slosEnabled") {
```
zugao, line 54: Not sure if this flag is necessary, I haven't seen it anywhere in the config. Meanwhile inside `CreateNetworkPolicy() `function there is the `sliNS` being checked, maybe it's the same flag repurposed?

pkg/comp-functions/functions/common/netpol.go
```diff
@@ -8,10 +8,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TODO: it would probably make sense to call this from the `BootstrapInstanceNs()` function. For future reference...
-
+// CreateNetworkPolicy creates network policy in the instance namespace to allow other namespaces access to the service
 func CreateNetworkPolicy(sourceNs []string, instanceNs string, instance string, svc *runtime.ServiceRuntime) error {
```
zugao, line 12: Line 25 bellow we might want to check if sliNamespace is set? what if a service does not have slos enabled

pkg/comp-functions/functions/common/instance_namespace.go
zugao: After another round of debugging it seems we do not need `slosEnabled `. It's not even used in mariadb, it's always false

Kidswiss: Grepping over the component, it doesn't look like it's used anywhere. Maybe a leftover?

I'd suggest to get rid of this if.

pkg/comp-functions/functions/common/netpol.go
Kidswiss: Yeah makes sense, otherwise we add a `""` allow rule, not sure what happens then.

Kidswiss (changes requested):
I've answered your qestions.

Otherwise, LGTM.

Kidswiss (approved):

PR #161, opened 2024-05-03
Inject loadbalancer specific annotations

Kidswiss:
## Summary

* Some APPUiO managed instances need specific labels on the `service` objects to work properly
* This change allows passing these annotations via the composition inputs
* Also fix an issue where the `primary-service` is not deployed if loadbalancer is disabled. It's still necessary for port-forwarding

## Checklist

- [x] Update tests.


zugao (approved):
LGTM, please test it in lab before deploying to prod

PR #163, opened 2024-05-08
Add allowAllNamespaces field

zugao:
## Summary

* This PR exposes another field `allowAllNamespaces` which overrides the values in `allowNamespaces`.
* Tested in lab

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/common_types.go
+++ b/apis/vshn/v1/common_types.go
@@ -178,6 +178,9 @@ type VSHNScheduleStatus struct {
 
 // Security defines the security of a service
 type Security struct {
+	// AllowAllNamespaces allows the service to be accessible from all namespaces, this supersedes the AllowedNamespaces field
+	// +kubebuilder:default=false
+	AllowAllNamespaces bool `json:"allowAllNamespaces,omitempty"`
 	// AllowedNamespaces defines a list of namespaces from where the service can be reached in the claim namespace
 	AllowedNamespaces []string `json:"allowedNamespaces,omitempty"`
 }
--- a/apis/vshn/v1/dbaas_vshn_keycloak.go
+++ b/apis/vshn/v1/dbaas_vshn_keycloak.go
@@ -278,7 +278,15 @@ func (v *VSHNKeycloak) GetFullMaintenanceSchedule() VSHNDBaaSMaintenanceSchedule
 	return schedule
 }
 
-// GetAllowedNamespaces returns allowed namespaces to access this service
+// GetAllowAllNamespaces returns the AllowAllNamespaces field of this service
+func (v *VSHNKeycloak) GetAllowAllNamespaces() bool {
+	return v.Spec.Parameters.Security.AllowAllNamespaces
+}
+
+// GetAllowedNamespaces returns the AllowedNamespaces array of this service
 func (v *VSHNKeycloak) GetAllowedNamespaces() []string {
-	return v.Spec.Parameters.Security.AllowedNamespaces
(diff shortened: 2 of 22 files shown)


apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -278,7 +278,7 @@ func (v *VSHNKeycloak) GetFullMaintenanceSchedule() VSHNDBaaSMaintenanceSchedule
 	return schedule
 }
 
-// GetAllowedNamespaces returns allowed namespaces to access this service
-func (v *VSHNKeycloak) GetAllowedNamespaces() []string {
-	return v.Spec.Parameters.Security.AllowedNamespaces
+// GetSecurity returns the security object this service
```
Kidswiss, line 281: Please add this to the API template as well. The next guy who creates a new service will thank you. :D

It's under `hack/bootstrap/template`

apis/vshn/v1/common_types.go
```diff
@@ -178,6 +178,8 @@ type VSHNScheduleStatus struct {
 
 // Security defines the security of a service
 type Security struct {
+	// AllowAll allows the service to be accessible from all namespaces, this supersedes the AllowedNamespaces field
```
Kidswiss: Should we maybe set the default to `false` explicitly via a controller-gen comment?

pkg/comp-functions/functions/common/netpol.go
```diff
@@ -9,29 +9,31 @@ import (
 )
 
 // CreateNetworkPolicy creates network policy in the instance namespace to allow other namespaces access to the service
-func CreateNetworkPolicy(sourceNs []string, instanceNs string, instance string, svc *runtime.ServiceRuntime) error {
+func CreateNetworkPolicy(sourceNs []string, instanceNs string, instance string, allowAll bool, svc *runtime.ServiceRuntime) error {
```
Kidswiss, line 12: We could change parameters so it takes a `Composite` interface type, then we can reduce the amount of params and make it easier to handle.

You could even specify a `SecurityGetter` interface which only contains the `GetSecurity()` function, to make it even more specific. Then you can add the `SecurityGetter` back to the `Composite` as an anonymous field, so every `Composite` is automaticalyl also a `SecurityGetter`.

Something like this:

```
type Composite interface {
	InfoGetter
	client.Object
        SecurityGetter
	SetInstanceNamespaceStatus()
	GetAllowedNamespaces() []string
}

type SecurityGetter interface {
       GetSecurity() vshnv1.Security
}

```

```suggestion
func CreateNetworkPolicy(source SecurityGetter, instanceNs string, instance string, svc *runtime.ServiceRuntime) error {
```

Kidswiss (changes requested):
Looks good, some ideas for optimizations though.

apis/vshn/v1/common_types.go
zugao: no problem from my side

pkg/comp-functions/functions/common/netpol.go
zugao: The only reason I didn't use the interface it's because on some cases such as keycloak pg I pass the value as false

zugao: We came to an agreement after PP.

apis/vshn/v1/dbaas_vshn_keycloak.go
zugao: I added it, now sure if it's right, please check

PR #169, opened 2024-05-21
Add Keycloak&Mariadb alerting, non-sla rules

wejdross:
## Summary

* Adding non-sla alerts for Keycloak and Mariadb as well as user alerting

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/common/mailgun_alerting.go
```diff
@@ -35,6 +35,7 @@ func MailgunAlerting(obj client.Object) func(ctx context.Context, svc *runtime.S
 		instanceNamespace := alertConfig.GetInstanceNamespace()
 		name := obj.GetName()
 
+		log.Info("here i am once again")
```
zugao, line 38: I suppose this should be removed

pkg/comp-functions/functions/common/non_sla_prom_rules.go
```diff
@@ -135,6 +102,44 @@ func generatePromeRules(elem InfoGetter, svc *runtime.ServiceRuntime) error {
 		},
 	}
 
+	if serviceName != "keycloak" {
```
zugao, line 105: Only MemoryCritical is active for keycloak? Also I don't like this `if` in common. It's not suppose to work like that. Either find a way to pass this from `keycloak` package or don't use _common_ package.

zugao (changes requested):

pkg/comp-functions/functions/common/mailgun_alerting.go
wejdross: fixed

pkg/comp-functions/functions/common/non_sla_prom_rules.go
wejdross: The problem is, Keycloak is really the only service we have that stores completely nothing, it just connects to DB. Alternative solution would be to copy-paste file for vshnkeycloak and adjust is for just keycloak - which I find even worse than `if` statement. This way it's consistent, file in common solves issue for all of the services, if we introduce some day another service that doesn't use any storage, we then append `if` statement

zugao: {quote}
Alternative solution would be to copy-paste file for vshnkeycloak and adjust is for just keycloak
{quote}
Why would you copy paste files? Just change the code in such a way to not copy paste and not use if clause. Change or add functions definition that would achieve that.

zugao: I would go with this solution `Another one could be to define a map of service - list of rules in a separate file.` You create a new struct like `ServicePromRule` which has a map inside. In that map you add the rules per each service. Since most services use the same set of rules you don't copy paste anything. Then you can pass this object `ServicePromRule` to  `GenerateNonSLAPromRules()` or figure another way how to do it.

Kidswiss: I like the map idea.

zugao: `
type ServicePromRule struct {
	a map[string][]promV1.Rule
}
`

Then you initialise and populate it somewhere in a new file named `service-prom-rules.go`. That way we have all the rules per each service clearly displayed. 

OR

`
type ServicePromRule struct {
	service string
	r []promV1.Rule
}
`
And you add it to the definition of the `GenerateNonSLAPromRules()`

zugao: I was just thinking about this for a bit longer. This kind of refactoring will help us get out from the business logic only the code changes based on the service we deploy. At some point we will have enough "generic" code that we might be able to create a framework that in the end will ask in input the set of prometheus rules to deploy for the service.

pkg/comp-functions/functions/common/non_sla_prom_rules.go
```diff
@@ -22,14 +22,29 @@ var (
...
+	Memory        bool
+}
+
+type RuleStruct struct {
+	name, serviceName, instanceNamespace string
+	instanceNamespaceRegex               string
+	ruleList                             []promV1.Rule
+}
+
+func GenerateNonSLAPromRules(obj client.Object, alerts AlertsEnabled) func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
```
zugao, line 43: The initial idea was to use some sort of `RuleStruct` in here and not `AlertsEnabled`. From my point of view there is no need for 2 structs. We just need one and to expose it. My previous proposal was something around the definition of `RuleStruct` but we can also just use a set of booleans like you defined in `AlertsEnabled`. That's actually not bad. So I would remove RuleStruct all together and use here `AlertsEnabled`. Also I would remove the `All` `bool` from the struct and create a constructor called AlertsEnabledAll() that would set all the alert types to true. It's more clean that way. @Kidswiss WDYT?

Kidswiss: I'd still go with your first idea. Because it would generalize service specific metrics as well. With the current approach we only generalize the metrics provided by `kube-state-metrics`. However, it would be nice if this could be extended to deploy any other service specific alerts as well. For example max connections for PostgreSQL (which we dropped from the PnT composition but haven't moved to the comp functions, as I just noticed... ).

wejdross:
@zugao I just added tests and fixed one small bug with PostgreSQL as it has different container name and namespace. Please take a look

Kidswiss (approved):

PR #170, opened 2024-05-27
Add simple user management for PostgreSQL

Kidswiss:
## Summary

This adds a simple user management for PostgreSQL. It supports creating users, databases and assigning privileges.

## Checklist

- [x] Update tests.


Kidswiss:
I had to download all the CRDs for the `provider-sql` and run the Crossplane codegen against it. As they haven't been updated yet to the latest `managed.Resoucre` interface.

zugao (commented):
I noticed you don't manage users that have been deleted, am I right? also there are two folders _mssql_ and _mysql_. Is that how it is intended?

Kidswiss:
> I noticed you don't manage users that have been deleted, am I right? also there are two folders _mssql_ and _mysql_. Is that how it is intended?

If the user gets removed from the array, then it will effectively be deleted from the instance.

Yes those two folders are directly imported from `provider-sql`, as it has different API packages for different.

zugao:
> I had to download all the CRDs for the `provider-sql` and run the Crossplane codegen against it. As they haven't been updated yet to the latest `managed.Resoucre` interface.

Can we separate those CRDs from ours?

zugao:
One thing that may bother me is the fact that unintentionally removing a database will delete data. Also a bug in the code can delete whole databases from the array. Does this PR addresses that? If not can we discuss this as I think it's important.

Kidswiss:
> > I had to download all the CRDs for the `provider-sql` and run the Crossplane codegen against it. As they haven't been updated yet to the latest `managed.Resoucre` interface.
> 
> Can we separate those CRDs from ours?

How do you mean separate them? They are in their own packages for now.

zugao:
> > > I had to download all the CRDs for the `provider-sql` and run the Crossplane codegen against it. As they haven't been updated yet to the latest `managed.Resoucre` interface.
> > 
> > 
> > Can we separate those CRDs from ours?
> 
> How do you mean separate them? They are in their own packages for now.

I meant the generated crds from  `crds` directory.

Kidswiss:
> One thing that bothers me is the fact that unintentionally removing a database will delete data. Also a bug in the code can delete whole databases from the array. Does this PR addresses that? If not can we discuss this as I think it's important.

The code is currently written in a way that it never returns on error, so it will always process the whole array, this should avoid the kind of bug we had with the namespaces, where we returned before processing the namespace. Alternatively I could try to introduce fatal errors, which would completely stop the execution of the function. Though this tends to create deadlocks during initial provisioning.

The array itself cannot be changed from composition functions, as only the status is writable, so if a database gets removed in the claim it's always an outside influence, either intentional or not. If you have an idea to make it harder for a user to unintentionally remove a database, I'm all ears.

Kidswiss:
> > > > I had to download all the CRDs for the `provider-sql` and run the Crossplane codegen against it. As they haven't been updated yet to the latest `managed.Resoucre` interface.
> > > 
> > > 
> > > Can we separate those CRDs from ours?
> > 
> > 
> > How do you mean separate them? They are in their own packages for now.
> 
> I meant the generated crds from `crds` directory.

Ah I see, let me look into that.

zugao:
One last thing, why don't we use the existing apis from https://github.com/crossplane-contrib/provider-sql/tree/master/apis? Why do we have to generate them here and clutter our repo? What am I missing?

Kidswiss:
> One last thing, why don't we use the existing apis from https://github.com/crossplane-contrib/provider-sql/tree/master/apis? Why do we have to generate them here and clutter our repo? What am I missing?

This:
https://github.com/vshn/appcat/pull/170#issuecomment-2133026623

The `provider-sql` uses an older crossplane code gen, which doesn't include the latest interface changes. Which makes it incompatible with the crossplane libraries we're using. So we have to download it into our repo and run the current code gen against it. Until they update it on their end.

zugao:
> > One last thing, why don't we use the existing apis from https://github.com/crossplane-contrib/provider-sql/tree/master/apis? Why do we have to generate them here and clutter our repo? What am I missing?
> 
> This: [#170 (comment)](https://github.com/vshn/appcat/pull/170#issuecomment-2133026623)
> 
> The `provider-sql` uses an older crossplane code gen, which doesn't include the latest interface changes. Which makes it incompatible with the crossplane libraries we're using. So we have to download it into our repo and run the current code gen against it. Until they update it on their end.

ah sry, I read at the beginning and then forgot about it. Good.

zugao (approved):
LGTM

apis/vshn/v1/common_types.go
```diff
@@ -184,3 +187,24 @@ type Security struct {
...
+	Database *string `json:"database,omitempty"`
+
+	// Privileges specifies the privileges to grant the user. Please check
+	// the database's docs for available privileges.
+	Privileges []string `json:"privileges,omitempty"`
+
+	// WriteConnectionSecretToReference specifies the namespace and name of a
+	// Secret to which any connection details for this user should
+	// be written.
+	// If not specified, a secret with the name $claimname-$username will be
```
zugao, line 207: Can we do that for the main secret please? :D

zugao (approved):
LGTM but if you could separate our vshn generated CRDs from thouse of provider-sql that would be great.

wejdross (approved):

apis/vshn/v1/common_types.go
Kidswiss: Not sure, as the main `writeConnectionSecretToReference` is controlled by Crossplane.

I found out that we can't actually access the `writeConnectionSecretToReference` of the claim. We can only access the field as set in the composite. Which is distinct from the one in the claim and points to the designated `writeConnectionSecretsToNamespace` as specified in the composition. We have 0 control over that from within the comp functions.

That's why it defaults to `$claimname-$username` and not to `$mainsecretname-$username`, because there's no easy way to get that value...

PR #172, opened 2024-05-29
updating dependencies

wejdross:
## Summary

* update providers and dependencies
* update `crank` command in package's Makefile
## Checklist

- [x] Update tests.


Kidswiss:
Is the title of the PR maybe wrong? You only updated the crossplane library here. There are no changes specific to the providers.

package/package.mk
```diff
@@ -2,7 +2,7 @@
 package-function: docker-build
 	yq e '.spec.image="${GHCR_IMG}"' package/crossplane.yaml.template > package/crossplane.yaml
 	rm -f package/*.xpkg
-	go run github.com/crossplane/crossplane/cmd/crank xpkg build -f package --verbose --embed-runtime-image=${GHCR_IMG} -o package/package-function-appcat.xpkg
+	go run github.com/crossplane/crossplane/cmd/crank@v1.16.0 xpkg build -f package --verbose --embed-runtime-image=${GHCR_IMG} -o package/package-function-appcat.xpkg
```
zugao, line 5: If version is not specified wouldn't it get the latest version?

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -235,6 +235,9 @@ type VSHNPostgreSQLStatus struct {
 	NetworkPolicyConditions      []v1.Condition `json:"networkPolicyConditions,omitempty"`
 	LocalCAConditions            []v1.Condition `json:"localCAConditions,omitempty"`
 	CertificateConditions        []v1.Condition `json:"certificateConditions,omitempty"`
+	// ObservedGeneration represents the .metadata.generation that the condition was set based upon.
+	// For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date with respect to the current state of the instance.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
zugao, line 240: Is this automatically managed by K8s?

package/package.mk
Kidswiss: It's a long story, but no. It would get the version defined in the modules `go.mod`. Usually this is fine, but we can't currently use this version of crank via `go.mod` because it has some dependency that conflicts with another dependency in `functions-go-sdk`.

apis/vshn/v1/dbaas_vshn_postgresql.go
Kidswiss: It doesn't look like we need to add this manually. The Crossplane code generator added it in the CRD yamls.

zugao:
Please test this important update in lab while having each service instance already present in the cluster!!!

wejdross:
> What happened with all the MinIO APIs? We did they all get deleted?

https://github.com/vshn/appcat/pull/172/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R39

wejdross:
> Please test this important update in lab while having each service instance already present in the cluster!!!

it was tested in local kindev as well as in LAB, downgrading provider-kubernetes is really painful, but if You want to test it manually, it's all deployted into LAB cluster right now

apis/vshn/v1/dbaas_vshn_postgresql.go
wejdross: fixed it

Kidswiss (approved):

PR #173, opened 2024-05-29
Migrate PostgreSQL from P&T to comp-functions

TheBigLee:
## Summary

* This PR migrates all P&T of VshnPostgreSQL to composition functions. This gives us better flexibility and one single point of thruth.

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
```diff
@@ -2,31 +2,604 @@ package vshnpostgres
...
+	return nil
+}
+
+func createSgInstanceProfile(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error {
+	plan := comp.Spec.Parameters.Size.GetPlan(svc.Config.Data["defaultPlan"])
+
+	resources, err := utils.FetchPlansFromConfig(ctx, svc, plan)
+	if err != nil {
+		err = fmt.Errorf("cannot fetch plans from the composition config, maybe they are not set: %w", err)
+		return err
```
Kidswiss, line 190: You have to be careful with these when reading from the config. Especially if the input is new.

There's a chance the new comp function runs before the composition is updated. That's because there's no ordering in our argocd deployment. That could lead to the missing config here, and then it would delete the a whole lot of stuff, because it doesn't get added to the desired state. That's basically how we deleted all the namespaces.

It would probably make sense to add a fatal result right here to the desired state via `svc.AddResult()`. As long as there's at least one fatal result, Crossplane will not do any changes at all.

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
```diff
@@ -2,31 +2,604 @@ package vshnpostgres
...
+
+	return nil
+}
+
+func createSgInstanceProfile(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error {
+	plan := comp.Spec.Parameters.Size.GetPlan(svc.Config.Data["defaultPlan"])
+
+	resources, err := utils.FetchPlansFromConfig(ctx, svc, plan)
+	if err != nil {
+		err = fmt.Errorf("cannot fetch plans from the composition config, maybe they are not set: %w", err)
```
Kidswiss, line 189: Don't assign `err` just to return it. Return the `fmt.Errorf()` directly.

Kidswiss (changes requested):
You currently have the exact same race condition in there that wiped out almost all of the postgresql instances.

This race condition here will have the same effect.

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
```diff
@@ -2,31 +2,604 @@ package vshnpostgres
...
+		return runtime.NewWarningResult(fmt.Errorf("cannot create tls certificate: %w", err).Error())
+	}
+
+	l.Info("Create Stackgres objects")
+	err = createStackgresObjects(ctx, comp, svc)
+	if err != nil {
+		return runtime.NewWarningResult(fmt.Errorf("cannot create stackgres objects: %w", err).Error())
+	}
+
+	l.Info("Create ObjectBucket")
```
Kidswiss, line 66: Thinking about this a bit more, it probably makes sense to create the bucket a lot earlier in the logic, as it stands now, the backups would also be destroyed.

Kidswiss:
This PR shows, that we need to drastically rethink how we should handle errors. Golang's "return early" paradigm is unfortunately very unhelpful here.

We should never return early during a comp function, the chance to skip something important is too high. We should come up with a set of rules, make use of the fatal result and bake that logic into the runtime somehow. What do you think @zugao ?

zugao:
> You currently have the exact same race condition in there that wiped out almost all of the postgresql instances.
> 
> This race condition here will have the same effect.

Maybe we could add some guidelines to README in regards to that? When to throw fatal errors and when simply return an error?

zugao:
> We should never return early during a comp function, the chance to skip something important is too high. We should come up with a set of rules, make use of the fatal result and bake that logic into the runtime somehow. What do you think @zugao ?

Yes, we definitely need to discuss this further and come up with a solid foundation in the runtime package.

Kidswiss:
> > You currently have the exact same race condition in there that wiped out almost all of the postgresql instances.
> > This race condition here will have the same effect.
> 
> Maybe we could add some guidelines to README in regards to that? When to throw fatal errors and when simply return an error?

I feel like that alone is not enough for something that essential, see my other comment for baking this into the runtime.

zugao:
I already have some ideas how we could do it.

Makefile
```diff
@@ -93,14 +93,27 @@ generate-stackgres-crds:
...
+
 	# The generator for the pool config CRD unfortunately produces a broken result. However if we ever need to regenerate it in the future, please uncomment this.
 	# curl ${STACKGRES_CRD_URL}/SGPoolingConfig.yaml?inline=false -o apis/stackgres/v1/sgpoolconfigs_crd.yaml
 	# yq -i e apis/stackgres/v1/sgpoolconfigs.yaml --expression ".components.schemas.SGPoolingConfigSpec=load(\"apis/stackgres/v1/sgpoolconfigs_crd.yaml\").spec.versions[0].schema.openAPIV3Schema.properties.spec"
 	# yq -i e apis/stackgres/v1/sgpoolconfigs.yaml --expression ".components.schemas.SGPoolingConfigStatus=load(\"apis/stackgres/v1/sgpoolconfigs_crd.yaml\").spec.versions[0].schema.openAPIV3Schema.properties.status"
 	# go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen --package=v1 -generate=types -o apis/stackgres/v1/sgpoolconfigs.gen.go apis/stackgres/v1/sgpoolconfigs.yaml
 	# perl -i -0pe 's/\*struct\s\{\n\s\sAdditionalProperties\smap\[string\]string\s`json:"-"`\n\s}/map\[string\]string/gms' apis/stackgres/v1/sgpoolconfigs.gen.go
 
+	curl ${STACKGRES_CRD_URL}/SGObjectStorage.yaml?inline=false -o apis/stackgres/v1beta1/sgobjectstorage_crd.yaml
```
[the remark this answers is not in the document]
TheBigLee: I've commented those out

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
```diff
@@ -2,31 +2,604 @@ package vshnpostgres
...
+		l.Info("Create copy job")
+		err = createCopyJob(comp, svc)
+		if err != nil {
+			return runtime.NewWarningResult(fmt.Errorf("cannot create copyJob object: %w", err).Error())
+		}
+	}
+	return nil
+}
+
+func createCerts(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error {
```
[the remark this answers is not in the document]
TheBigLee: Moved to common package

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
TheBigLee: Good point. I moved it up before the instance namespace creation

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
TheBigLee: As I'm also using `svc.AddResult()` now, I still need the `err` variable.

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
TheBigLee: Added `svc.AddResult()` where necessary

zugao (approved):
It seems fine for me though the change is massive. Please test it throughly in lab cluster with existing PostgreSQL(you can check on central DB) and a new one before deploying into prod.

Kidswiss (approved):
You need to rebase, the minio types can now be completely removed.

I also agree with Gabriel to test it thoroughly on the lab.

wejdross:
From my side: please add some tests, PostgreSQL is our most advanced feature and I believe it would be cool to have tests :) there are e2e as well in component-appcat, please make sure, they're all passing on lab

TheBigLee:
I've added some tests.This should now be ready for rollout

zugao (approved):
LGTM

Kidswiss (approved):

PR #174, opened 2024-06-04
Event forward in comp function services

zugao:
## Summary

* Adds event forwarding for all services under appcat. This applies to services that have their CRs managed by comp functions

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -1081,7 +1081,7 @@ func (s *ServiceRuntime) ForwardEvents() error {
 	for _, res := range s.desiredResources {
 		r := res.Resource
 
-		// For kube objects set the annotation 'EventForwardAnnotation' in managed resource and not in kube object itself
+		// For kube objects set the annotation 'EventForwardAnnotation' for themselves and for managed resource
 		if isKubeObject(r) {
 			p := "spec.forProvider.manifest.metadata.annotations"
 			v, _ := r.GetValue(p)
@@ -1094,17 +1094,13 @@ func (s *ServiceRuntime) ForwardEvents() error {
 			if err != nil {
 				return fmt.Errorf("cannot set event forward annotations for managed object %s: %w", r.GetName(), err)
 			}
-
-			// For non kube objects set directly into the managed resource
-		} else {
-			annotations := r.GetAnnotations()
-			if annotations == nil {
-				annotations = map[string]string{}
-			}
-			annotations[EventForwardAnnotation] = eventForwardValue
-			r.SetAnnotations(annotations)
 		}
-
+		annotations := r.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		annotations[EventForwardAnnotation] = eventForwardValue
+		r.SetAnnotations(annotations)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


wejdross (approved):

Kidswiss (changes requested):
Please make the PR name more descriptive, like `Event forwarding for external object storage`

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1064,3 +1069,45 @@ func (s *ServiceRuntime) addUsages() error {
 	}
 	return nil
 }
+
+func (s *ServiceRuntime) ForwardEvents() error {
+	claimRef := s.observedComposite.GetClaimReference()
```
Kidswiss, line 1074: We probably drop some events before the composite reference is available though.

zugao: No they are not, and I don't even care. I had the same problem with desired, some times I don't have anything in there. Do you still want to go with desired?

zugao:
> Please make the PR name more descriptive, like `Event forwarding for external object storage`

for all services :)

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: I guess it's fine for now.

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1064,3 +1069,45 @@ func (s *ServiceRuntime) addUsages() error {
...
+				mrAnnotations = v.(map[string]any)
+			}
+			mrAnnotations[EventForwardAnnotation] = eventForwardValue
+			err := r.SetValue(p, mrAnnotations)
+			if err != nil {
+				return fmt.Errorf("cannot set event forward annotations for managed object %s: %w", r.GetName(), err)
+			}
+
+			// For non kube objects set directly into the managed resource
+		} else {
```
Kidswiss, line 1098: Shouldn't we also add the annotation to the outer kubeobject as well? Sometimes they might also fail to actually apply an object due to webhooks or other issues.

wejdross: wouldn't that cause doubled entries in 99% of cases? outer object contains inner object's events and that would make readibility much worse

Kidswiss:
> > Please make the PR name more descriptive, like `Event forwarding for external object storage`
> 
> for all services :) I updated the title

See, making it more descriptive helps :D

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: Not necessarily, the outer object is more or less only concerned with applying the inner object to the cluster. For example if there's a webhook that blocks `.spec.important = false`, or if you try to apply a namespaced object without a namespace. That would only show up in the outer object. In such a case the object would not even reach the kubernetes api, and we'd have to describe the kubeobject to figure out why.

The events of the inner object are created by the controller that handles said object, only after its creation.

zugao: I will add them to the kube Objects as well

TheBigLee (approved):

Kidswiss (approved):

PR #176, opened 2024-06-05
Fix controller not having enough RBAC

Kidswiss:
Some combinations of OpenShift and client library can cause this issue. Either OpenShift requires more stricter RBAC or the client library do changes they didn't before. Either way: https://access.redhat.com/solutions/5372461

## Summary

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/minio/v1/bucket_types.go
+++ b/apis/minio/v1/bucket_types.go
@@ -48,6 +48,7 @@ type Bucket struct {
 
 type BucketSpec struct {
 	xpv1.ResourceSpec `json:",inline"`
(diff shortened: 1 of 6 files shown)


apis/minio/v1/bucket_types.go
```diff
@@ -48,6 +48,7 @@ type Bucket struct {
 
 type BucketSpec struct {
 	xpv1.ResourceSpec `json:",inline"`
+	ProviderReference *xpv1.Reference  `json:"providerReference,omitempty"`
```
Kidswiss, line 51: This came in via `make generate` as provider-minio was updated recently.

zugao (approved):

PR #179, opened 2024-06-12
Escape Object names before applying

Kidswiss:
## Summary

All managed resources and kube objects will now be automatically escaped. K8s naming allows different naming schemes for different objects, the current logic should cover most of them. Unfortunately there's no official docs about which objects support which naming.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/runtime/naming.go
+++ b/pkg/comp-functions/runtime/naming.go
@@ -1,0 +1,93 @@
+package runtime
+
+import (
+	"crypto/sha1"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/crossplane/function-sdk-go/resource/composed"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// escapeK8sNames will figure out what naming scheme applies to the given object
+// and tries to find the correct naming scheme for it.
+// Then it will escape the name accordingly.
+func escapeK8sNames(obj client.Object) {
+	kind, _, err := composed.Scheme.ObjectKinds(obj)
+	if err != nil {
+		obj.SetName(escapeDNS1123(obj.GetName(), false))
+	}
+
+	switch kind[0].Kind {
+	case "Pod":
+		obj.SetName(escapeDNS1123Label(obj.GetName()))
+	case "Namespace":
+		obj.SetName(escapeDNS1123Label(obj.GetName()))
+	case "Service":
+		obj.SetName(escapeDNS1123Label(obj.GetName()))
+	case "Role":
+		obj.SetName(escapeDNS1123(obj.GetName(), true))
+	case "ClusterRole":
+		obj.SetName(escapeDNS1123(obj.GetName(), true))
+	case "RoleBinding":
+		obj.SetName(escapeDNS1123(obj.GetName(), true))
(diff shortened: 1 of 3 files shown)


pkg/comp-functions/runtime/naming.go
```diff
@@ -0,0 +1,76 @@
...
+
+// escapeDNS1123 will always return a string that conforms to K8s' DNS subdomain naming scheme
+// contain no more than 253 characters
+// contain only lowercase alphanumeric characters, '-' or '.'
+// start with an alphanumeric character
+// end with an alphanumeric character
+// We also remove any '.' here, so that it can be used as a base function for escaping
+// label names as well.
+func escapeDNS1123(name string, allowColons bool) string {
+	escapeNameStart := regexp.MustCompile("^[^a-zA-Z0-9]+")
```
Kidswiss, line 51: My regex foo isn't the best, maybe there are better ways to do this.

pkg/comp-functions/runtime/naming.go
```diff
@@ -0,0 +1,82 @@
...
+	default:
+		obj.SetName(escapeDNS1123(obj.GetName(), false))
+	}
+}
+
+// escapeDNS1123Label does the same as escapeDNS1123 but also limit to 63 chars
+func escapeDNS1123Label(name string) string {
+	name = escapeDNS1123(name, false)
+	if len(name) > 63 {
+		name = name[:63]
```
wejdross, line 44: This is somehow Ok imho, there is a danger of overriding values if too long labels are truncated instead of returning an error:
```
app.kubernetes.io/this-is-a-very-long-label-key-to-reach-the-limit-user1111: "user1111"
app.kubernetes.io/this-is-a-very-long-label-key-to-reach-the-limit-user2222: "user2222"
```

if we truncate, we'll end up with only one label of value:
```
app.kubernetes.io/this-is-a-very-long-label-key-to-reach-the-l: user2222
```
competely losing logic

pkg/comp-functions/runtime/naming.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	escapedName := strings.ToLower(name)
+
+	escapedName = escapeExpression.ReplaceAllString(escapedName, "-")
+	escapedName = escapeNameStart.ReplaceAllString(escapedName, "")
+	escapedName = escapeNameEnd.ReplaceAllString(escapedName, "")
+
+	if len(escapedName) > 253 {
+		escapedName = escapedName[:253]
```
wejdross, line 78: same objection as above

pkg/comp-functions/runtime/naming.go
Kidswiss: I though about the same, I could introduce some randomness at the. Like what a deployment does.

BTW, you're confusing what label means in this context. Here label refers to a DNS label, which is RFC1123 and specifies what's allowed in a kubernetes object name, see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names. We're escaping `metadata.name` here, not object labels, although labels probably also have some constraints to them I guess.

wejdross (approved):

zugao (approved):
LGTM, finally we are managing this

PR #182, opened 2024-06-18
Escape the names of generated secrets

Kidswiss:
## Summary

I think I dropped this escape call at one point when doing a rebase...

## Checklist

- [ ] Update tests.


zugao (approved):
LGTM

PR #183, opened 2024-06-18
Ensure all services deploy quotas

Kidswiss:
## Summary

Previously the quota annotations had to be added as a separate step. It's now part of the namespace provisioning, so every current and future service will get it automatically.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM

PR #184, opened 2024-06-25
Refactor and fix Resource calculations

TheBigLee:
## Summary

The resource calculation between the used plan and the overridden values in the claim for the respective resources contained wrong calculations.

We refactored the whole calculations sections to get correct calculations while keeping the code as readable as possible.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/resources.go
+++ b/pkg/comp-functions/functions/common/resources.go
@@ -17,47 +17,48 @@ type Resources struct {
 // GetResources will return a `Resources` object with the correctly calculated requests,
 // limits and disk space according to the definitions in the plan as well as the overrides
 // in the claim.
-func GetResources(size *vshnv1.VSHNSizeSpec, plan utils.Resources) (Resources, error) {
+func GetResources(size *vshnv1.VSHNSizeSpec, plan utils.Resources) (Resources, []error) {
 	reqMem := resource.Quantity{}
 	reqCPU := resource.Quantity{}
 	mem := resource.Quantity{}
 	cpu := resource.Quantity{}
 	disk := plan.Disk
 
+	var errors []error
 	var err error
 
 	if size.Requests.Memory != "" {
 		reqMem, err = resource.ParseQuantity(size.Requests.Memory)
 		if err != nil {
-			return Resources{}, err
+			errors = append(errors, err)
 		}
 	}
 
 	if size.Requests.CPU != "" {
 		reqCPU, err = resource.ParseQuantity(size.Requests.CPU)
 		if err != nil {
-			return Resources{}, err
+			errors = append(errors, err)
 		}
 	}
 
 	if size.Memory != "" {
 		mem, err = resource.ParseQuantity(size.Memory)
 		if err != nil {
-			return Resources{}, err
+			errors = append(errors, err)
 		}
 	}
 
 	if size.CPU != "" {
 		cpu, err = resource.ParseQuantity(size.CPU)
 		if err != nil {
-			return Resources{}, err
+			errors = append(errors, err)
 		}
 	}
 
 	if size.Disk != "" {
 		disk, err = resource.ParseQuantity(size.Disk)
 		if err != nil {
-			return Resources{}, err
+			errors = append(errors, err)
 		}
 	}
 
@@ -73,7 +74,7 @@ func GetResources(size *vshnv1.VSHNSizeSpec, plan utils.Resources) (Resources, e
 		Mem:    memLimit,
 		CPU:    cpuLimit,
 		Disk:   disk,
-	}, nil
+	}, errors
 }
 
 // getLimit will compare a given limit and request as well as the limit defined in the plan
--- a/pkg/comp-functions/functions/common/resources.go
+++ b/pkg/comp-functions/functions/common/resources.go
@@ -14,6 +14,9 @@ type Resources struct {
 	Disk   resource.Quantity
 }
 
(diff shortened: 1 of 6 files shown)


pkg/comp-functions/functions/common/resources.go
```diff
@@ -3,43 +3,111 @@ package common
...
-	reqCPU := size.Requests.CPU
-	mem := size.Memory
-	cpu := size.CPU
-	disk := size.Disk
+func GetResources(size *vshnv1.VSHNSizeSpec, plan utils.Resources) (Resources, error) {
+	reqMem := resource.Quantity{}
+	reqCPU := resource.Quantity{}
+	mem := resource.Quantity{}
+	cpu := resource.Quantity{}
+	disk := plan.Disk
```
zugao, line 22: Don't we have the same problem with the disk as well? For disk we should always get the highest value, we cannot go back on a lower value

pkg/comp-functions/functions/common/resources.go
```diff
@@ -3,43 +3,111 @@ package common
...
 
-	if reqMem == "" {
-		reqMem = r.MemoryRequests.String()
+	var err error
+
+	if size.Requests.Memory != "" {
+		reqMem, err = resource.ParseQuantity(size.Requests.Memory)
+		if err != nil {
+			return Resources{}, err
```
zugao, line 29: Consider not returning any error here but just continue, you would have an empty value which will be discarded later on in the logic. I fear we might create more problems by returning here an error. Applies to the rest of the variables as well.

pkg/comp-functions/functions/common/resources.go
```diff
@@ -3,43 +3,111 @@ package common
...
-		Mem:    mem,
-		CPU:    cpu,
+		ReqMem: memReq,
+		ReqCPU: cpuReq,
+		Mem:    memLimit,
+		CPU:    cpuLimit,
 		Disk:   disk,
+	}, nil
+}
+
```
zugao, line 75: please add a comment for the logic of this function, you can add it here or you could add in each `if else` clause.

pkg/comp-functions/functions/common/resources_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/vshn/appcat/v4/pkg/common/utils"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/functions/commontest"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestResources(t *testing.T) {
+
+	ctx := context.Background()
+	svc := commontest.LoadRuntimeFromFile(t, "plans.yaml")
+	resources, err := utils.FetchPlansFromConfig(ctx, svc, svc.Config.Data["defaultPlan"])
```
zugao, line 18: change the variable please, it was difficult to figure out it's the plan resource variable. Like `defaultPlan`. Also add a comment here or to each unit test what are the default plan's values, like memory, cpu. It will help to not check the file each time :)

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -233,7 +233,11 @@ func getResources(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1
 		return common.Resources{}, err
 	}
 
-	res := common.GetResources(&comp.Spec.Parameters.Size, resources)
+	res, err := common.GetResources(&comp.Spec.Parameters.Size, resources)
+	if err != nil {
+		err = fmt.Errorf("Cannot get Resources from plan and claim: %w", err)
+		return common.Resources{}, err
```
zugao, line 239: Again as said before, we might want to not return any error here...

zugao (changes requested):
Some things to take in consideration. The big problem I see is that we don't do any checks on the disk. We should always get the highest value for the disk as well.

pkg/comp-functions/functions/common/resources.go
TheBigLee: No. Because we need to be able to set a lower disk size than the plan provides on initial provisioning.
I would ignore the "lowering" disk space problem for now. It won't break anything. It will just not do anything.
Afaik documenting that lowering disk space is not possible should be enough.

pkg/comp-functions/functions/common/resources.go
TheBigLee: Might it make sense, to collect the errors and return the errors to the main routine and if there are errors, just print them out?

zugao: Yes that is a good compromise. We should not see any errors of this type here as I expect the validation to happen much earlier in the logic. So yeah let's gather all the errors and then return as a warning, might make sense that way.

pkg/comp-functions/functions/common/resources.go
zugao: I guess it's fine by me.

wejdross: can You cover that case in e2e? so we're sure that change doesn't affect end user with time (in case Crossplane introduce feature that will break things in that field)

pkg/comp-functions/functions/common/resources.go
TheBigLee: Done

pkg/comp-functions/functions/common/resources_test.go
TheBigLee: Done

zugao (approved):
LGTM

PR #185, opened 2024-06-26
Add support for nextcloud

TheBigLee:
## Summary

* This PR adds support for Nextcloud

## Checklist

- [x] Update tests.


apis/vshn/v1/vshn_nextcloud.go
```diff
@@ -0,0 +1,276 @@
...
+// XVSHNNextcloud represents the internal composite of this claim
+type XVSHNNextcloud struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   XVSHNNextcloudspec   `json:"spec"`
+	Status XVSHNNextcloudstatus `json:"status,omitempty"`
+}
+
+// XVSHNNextcloudspec defines the desired state of a VSHNNextcloud.
```
zugao, line 180:
```suggestion
// XVSHNNextcloudSpec defines the desired state of a VSHNNextcloud.
```

apis/vshn/v1/vshn_nextcloud.go
```diff
@@ -0,0 +1,276 @@
...
+type XVSHNNextcloudspec struct {
```
zugao, line 181:
```suggestion
type XVSHNNextcloudSpec struct {
```

pkg/comp-functions/functions/vshnnextcloud/deploy.go
```diff
@@ -0,0 +1,395 @@
...
+				"usernameKey": adminUserSecretField,
+				"passwordKey": adminPWSecretField,
+			},
+			"configs": map[string]string{
+				"vshn-nextcloud.config.php": nextcloudConfig,
+			},
+			"containerPort":   8080,
+			"securityContext": nil,
+			"podSecurityContext": map[string]any{
+				"runAsUser":                nil,
```
zugao, line 282: Will this work on vanilla k8s? Also it seems we are decreasing the security for k8s.

pkg/comp-functions/functions/vshnnextcloud/files/vshn-nextcloud.config.php
```diff
@@ -0,0 +1,4 @@
+<?php
+$CONFIG = array (
+  'check_data_directory_permissions' => false,
```
zugao, line 3: I guess this file will be merged with the default config.php right?

pkg/comp-functions/functions/vshnnextcloud/files/000-default.conf
```diff
@@ -0,0 +1,5 @@
+<VirtualHost *:8080>
```
zugao, line 1: Do we need this file? can't we use the default one from the image

zugao (changes requested):
Looks good, some small things

pkg/comp-functions/functions/common/password.go
```diff
@@ -74,3 +80,26 @@ func genPassword() (string, error) {
 
 	return gen.Generate(16, 4, 0, false, true)
 }
+
+func AddStaticFieldToSecret(staticFields map[string]string) CredentialSecretOption {
```
Kidswiss, line 84: Please add a comment what this function does. As it's part of the public API of the common package, it's important that it's documented.

pkg/comp-functions/functions/vshnnextcloud/deploy.go
```diff
@@ -0,0 +1,395 @@
...
+
+	err = addRelease(ctx, svc, comp, adminSecret)
+	if err != nil {
+		return runtime.NewWarningResult(fmt.Sprintf("cannot create release: %s", err))
+	}
+
+	return nil
+}
+
+func addPostgreSQL(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNNextcloud) error {
```
Kidswiss, line 120: This looks very much copy pasted from Keycloak. Maybe something to make generic?

pkg/comp-functions/functions/vshnnextcloud/deploy.go
```diff
@@ -0,0 +1,395 @@
...
+	// Unfortunately k8up and stackgres backups don't match up very well...
+	// if no daily backup is set we just do the default.
+	retention := 6
+	if comp.Spec.Parameters.Backup.Retention.KeepDaily != 0 {
+		retention = comp.Spec.Parameters.Backup.Retention.KeepDaily
+	}
+
+	configs := map[string]string{
```
Kidswiss, line 128: If I remember correctly, this is specific for Keycloak. We might not need this here.

pkg/comp-functions/functions/vshnnextcloud/deploy.go
Kidswiss: I agree. In `svc.Config` you can access the information if it's OpenShift or not.

Kidswiss (changes requested):
Some smaller things

pkg/comp-functions/functions/vshnnextcloud/files/vshn-nextcloud.config.php
TheBigLee: Yes

pkg/comp-functions/functions/vshnnextcloud/files/000-default.conf
TheBigLee: The default one binds to Port 80. That is the only reason why I have this file, as we can't bind to Port 80 on Openshift without additional SCCs.

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: Yes, this was copy&pasted by accident. I've removed it

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: I thought the same. But it's gonna make it rather complicated as I can't use our interface for the composition, since I need to access various fields in the composition. We should quickly discuss how to do that properly.

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: Refactored the code, to take this into consideration

pkg/comp-functions/functions/vshnnextcloud/deploy.go
Kidswiss: >  I can't use our interface for the composition

You can add functions to it as needed. Or maybe even specify a new interface specifically for composites with embedded PostgreSQLs.

pkg/comp-functions/functions/common/password.go
TheBigLee: Done

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: I've refactored the function and moved it to the `common` package

pkg/comp-functions/functions/common/resources.go
```diff
@@ -3,43 +3,133 @@ package common
...
-func GetResources(size *vshnv1.VSHNSizeSpec, r utils.Resources) Resources {
-	reqMem := size.Requests.Memory
-	reqCPU := size.Requests.CPU
-	mem := size.Memory
-	cpu := size.CPU
-	disk := size.Disk
+// GetResources will return a `Resources` object with the correctly calculated requests,
+// limits and disk space according to the definitions in the plan as well as the overrides
+// in the claim.
+func GetResources(size *vshnv1.VSHNSizeSpec, plan utils.Resources) (Resources, []error) {
```
zugao, line 20: Wasn't this function already merged into master?

apis/vshn/v1/vshn_nextcloud.go
```diff
@@ -0,0 +1,288 @@
...
+	Version string `json:"version,omitempty"`
+
+	// +kubebuilder:validation:Enum="besteffort";"guaranteed"
+	// +kubebuilder:default="besteffort"
+
+	// ServiceLevel defines the service level of this service. Either Best Effort or Guaranteed Availability is allowed.
+	ServiceLevel VSHNDBaaSServiceLevel `json:"serviceLevel,omitempty"`
+
+	// // +kubebuilder:default=true"
+	//
```
zugao, line 110: Please add a comment here that if this paramater is false, Nextcloud will use the default non managed sql lite database. Also add the fact that the PostgreSQL is fully managed.

TheBigLee: Added the comment

pkg/comp-functions/functions/common/resources.go
TheBigLee: Yes. I just rebased to the latest master. Diff is now gone

Kidswiss (approved):

PR #189, opened 2024-07-08
Fix wrong usage of `useExternalPostgreSQL` parameter and add better tests

TheBigLee:
## Summary

* This PR fixes a bug, where the `useExternalPostgreSQL` parameter operates exactly the wrong way. Instead of deploying a PostgreSQL instance when this is enabled, it will deploy one when this parameter is disabled.
* Added tests to catch this issue

## Checklist

- [x] Update tests.


Kidswiss (approved):

zugao (approved):
LGTM

PR #192, opened 2024-07-10
Add Nextcloud maintenance

zugao:
## Summary

* This PR will add automatic patch upgrades for enxtcloud.

## Checklist

- [x] Update tests.


cmd/maintenance.go
```diff
@@ -100,6 +102,10 @@ func (c *controller) runMaintenance(cmd *cobra.Command, _ []string) error {
 	case keycloak:
 		k := maintenance.NewKeycloak(kubeClient, getHTTPClient())
 		return k.DoMaintenance(cmd.Context())
+
+	case nextcloud:
```
Kidswiss, line 106: This starts to feel like something that can be optimized. Not sure for how many more services this switch case will scale :D

Kidswiss (approved):

cmd/maintenance.go
zugao: I added an interface here but i have no idea how to remove the switch and improve the code at the same time. Do you have any inputs?

Kidswiss: Maybe a similar pattern that we use for the functions?

Kidswiss (approved):

PR #193, opened 2024-07-10
Set the FQDN for nextcloud as a required field

TheBigLee:
## Summary

The FQDN needs to be always set, otherwise the lifeness and readiness probes will fail. Furthermore it doesn't make sense to run a nextcloud without a defined FQDN

## Checklist

- [] Update tests.


zugao (approved):
LGTM, just do a fast test in lab when you deploy

PR #194, opened 2024-07-12
Fix hardcoded servicename

TheBigLee:
## Summary

* The servicename is currently hardcorded in the composition function.
* This PR fixes this issue and sets the correct label for the service type

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/instance_namespace.go
+++ b/pkg/comp-functions/functions/common/instance_namespace.go
@@ -100,12 +100,17 @@ func createInstanceNamespace(serviceName, compName, claimNamespace, instanceName
 		svc.AddResult(runtime.NewWarningResult("cannot get claim namespace"))
 	}
 
+	mode := "standalone"
+	if _, exists := svc.Config.Data["mode"]; exists {
+		mode = svc.Config.Data["mode"]
+	}
+
 	ns := &corev1.Namespace{
 
 		ObjectMeta: metav1.ObjectMeta{
 			Name: instanceNamespace,
 			Labels: map[string]string{
-				"appcat.vshn.io/servicename":     serviceName + "-standalone",
+				"appcat.vshn.io/servicename":     serviceName + "-" + mode,
 				"appcat.vshn.io/claim-namespace": claimNamespace,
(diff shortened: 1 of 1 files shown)


pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -105,7 +105,7 @@ func createInstanceNamespace(serviceName, compName, claimNamespace, instanceName
 		ObjectMeta: metav1.ObjectMeta{
 			Name: instanceNamespace,
 			Labels: map[string]string{
-				"appcat.vshn.io/servicename":     serviceName + "-standalone",
+				"appcat.vshn.io/servicename":     serviceName + "-" + svc.Config.Data["mode"],
```
wejdross, line 108: Can You do some validation before? If You try to access key in map that doesn't exist, then it's going to be a nil pointer reference and ugly stack trace in logs at least, worst - restart of container

zugao: there will be no nil pointer but empty string. I checked a random composition and I didn't see "mode" value in config. It will end up as "<servicename>-". I would change to default if string is missing thus "-standalone"

TheBigLee: It's going to be set in any case in the component. Not sure if we should really check for that. And afaik it should not return a null pointer but an empty string as a result?

TheBigLee:
See https://github.com/vshn/component-appcat/pull/409 for reference

pkg/comp-functions/functions/common/instance_namespace.go
Kidswiss: I'd go with a default as well.

something like this:
```go
mode := "standalone"
if _, exists := svc.Config.Data["mode"]; exists {
  mode = svc.Config.Data["mode"]
}
```

TheBigLee: Thanks for the suggestions. I've implemented it accordingly

zugao (approved):
LGTM

Kidswiss (approved):

PR #195, opened 2024-07-15
Refactor time of day in maintenance

zugao:
## Summary

* This PR does refactoring for TimeOfDay field.
* This PR makes sure the database maintenance runs after main service maintenance.

## Checklist

- [ ] Update tests.

### DIFF
--- a/apis/vshn/v1/common_types.go
+++ b/apis/vshn/v1/common_types.go
@@ -3,7 +3,10 @@ package v1
 import (
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	alertmanagerv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	"time"
 )
+
+type TimeOfDay string
 
 // K8upBackupSpec specifies when a backup for redis should be triggered.
 // It also contains the retention policy for the backup.
@@ -67,10 +70,11 @@ type VSHNDBaaSMaintenanceScheduleSpec struct {
 	DayOfWeek string `json:"dayOfWeek,omitempty"`
 
 	// +kubebuilder:validation:Pattern="^([0-1]?[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])$"
+	// +kubebuilder:validation:Type=string
 
 	// TimeOfDay for installing updates in UTC.
 	// Format: "hh:mm:ss".
-	TimeOfDay string `json:"timeOfDay,omitempty"`
+	TimeOfDay TimeOfDay `json:"timeOfDay,omitempty"`
 }
 
 // GetMaintenanceDayOfWeek returns the currently set day of week
@@ -79,7 +83,7 @@ func (n *VSHNDBaaSMaintenanceScheduleSpec) GetMaintenanceDayOfWeek() string {
 }
 
 // GetMaintenanceTimeOfDay returns the currently set time of day
-func (n *VSHNDBaaSMaintenanceScheduleSpec) GetMaintenanceTimeOfDay() string {
+func (n *VSHNDBaaSMaintenanceScheduleSpec) GetMaintenanceTimeOfDay() TimeOfDay {
 	return n.TimeOfDay
 }
 
@@ -89,7 +93,7 @@ func (n *VSHNDBaaSMaintenanceScheduleSpec) SetMaintenanceDayOfWeek(dow string) {
 }
 
 // SetMaintenanceTimeOfDay sets the time of day to the given value
-func (n *VSHNDBaaSMaintenanceScheduleSpec) SetMaintenanceTimeOfDay(tod string) {
+func (n *VSHNDBaaSMaintenanceScheduleSpec) SetMaintenanceTimeOfDay(tod TimeOfDay) {
 	n.TimeOfDay = tod
 }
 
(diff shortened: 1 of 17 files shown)


apis/vshn/v1/common_types.go
```diff
@@ -208,3 +212,31 @@ type VSHNAccess struct {
 	// created in the namespace where the claim is located.
 	WriteConnectionSecretToReference *xpv1.SecretReference `json:"writeConnectionSecretToRef,omitempty"`
 }
+
+func (a *TimeOfDay) GetTime() time.Time {
+	v, err := time.Parse(time.TimeOnly, string(*a))
+	if err != nil {
```
zugao, line 218: Instead of getting an error I thought it's better to set the time to 00:00:00. I could also try to return a nil time? what do you think

wejdross: I'm OK with such approach, but can You please write tests for those fuctions? Working with time is always tricky and I would like to see some tests in here for future

bastjan: Either return an error and let the caller side decide what to do with it or at least document it. 

Also this will fail on `nil` TimeOfDay objects. Either fix the signature to `func (a TimeOfDay) GetTime()`, which is fine because it does not modify `a` or add a nil check.

zugao: > Either return an error and let the caller side decide what to do with it or document it. Returning zero time is ok (if documented) since the caller can use `time.Time.IsZero()`.

Thanks for input

> Also this will fail on `nil` TimeOfDay objects. Either fix the signature to `func (a TimeOfDay) GetTime()`, which is fine because it does not modify `a` or add a nil check.

Unfortunately I change the it [here](https://github.com/vshn/appcat/pull/195/files/0be649d04b7edefcedcc5d73870888cace0bf853#diff-e4106c7dc200d769b1c8a5341f5b5bb8d4cd00ed738e444a0d19ea04a6932923R235)

pkg/comp-functions/functions/common/postgresql.go
```diff
@@ -19,29 +17,61 @@ const (
 	PgSecretName         = "pg-creds"
 )
 
-func AddPostgreSQL(svc *runtime.ServiceRuntime, comp InfoGetter, psqlParams *vshnv1.VSHNPostgreSQLParameters, pgBouncerConfig map[string]string) error {
+type PostgreSQLDependencyBuilder struct {
```
TheBigLee, line 20: I'm having issues with the naming of this type and the function below. But I don't really have a better name in mind right now

apis/vshn/v1/common_types.go
zugao: > I'm OK with such approach, but can You please write tests for those fuctions? Working with time is always tricky and I would like to see some tests in here for future

Added unit tests and also checking for nil.

pkg/comp-functions/functions/common/postgresql.go
zugao: I also have no idea what would be the best name :)

apis/vshn/v1/common_types.go
zugao: > You can still change the signature of the GetTime function though

I didn't changed the signature because I am saving the state therefore I can only use a pointer. Nevertheless I am checking for nil pointers now and returning time zero.

bastjan: > I am saving the state therefore I can only use a pointer

You're not saving state in the getter. So you don't need a pointer receiver there. It's very much save to mix and match. In fact the standard library does that https://pkg.go.dev/time#Time

But plus one for `nil` check. This allow ppl to call GetTime() on a typed nil.

bastjan (dismissed):

apis/vshn/v1/common_types.go
```diff
@@ -208,3 +212,41 @@ type VSHNAccess struct {
...
+	return !a.IsSet()
+}
+
+// SetTime will set TimeOfDay from time.Time with time only format
+func (a *TimeOfDay) SetTime(t time.Time) {
+	*a = TimeOfDay(t.Format(time.TimeOnly))
+}
+
+// AddTime adds duration to current time
+func (a *TimeOfDay) AddTime(d time.Duration) TimeOfDay {
```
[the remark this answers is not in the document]
bastjan: (sorry for the approve. clicked the wrong button )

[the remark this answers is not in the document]
zugao: It's convenient for our use case. Will keep it that way for now.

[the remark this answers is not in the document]
bastjan: I tend to use the principle of least surprise when coding. but it's not my repo.

I personally would wrap `time.Time` like `metav1.Time` does.

pkg/comp-functions/functions/common/schedule.go
```diff
@@ -21,8 +22,8 @@ type BackupScheduler interface {
 type MaintenanceScheduler interface {
 	GetMaintenanceDayOfWeek() string
 	SetMaintenanceDayOfWeek(string)
-	GetMaintenanceTimeOfDay() string
-	SetMaintenanceTimeOfDay(string)
+	GetMaintenanceTimeOfDay() *v1.TimeOfDay
+	//SetMaintenanceTimeOfDay(v1.TimeOfDay)
```
wejdross, line 26: if it's not required anymore, then please delete that

wejdross (approved):
LGTM

PR #196, opened 2024-07-15
adding support for Keycloak 25, as our partners released an image

wejdross:
## Summary

## Checklist

- [ ] Update tests.


apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -99,7 +99,7 @@ type VSHNKeycloakServiceSpec struct {
 	// +kubebuilder:default="/"
 	RelativePath string `json:"relativePath,omitempty"`
 
-	// +kubebuilder:validation:Enum="23";"24"
+	// +kubebuilder:validation:Enum="23";"24";"25"
```
TheBigLee, line 102: I'm not sure if we should even do such a validation here, as we would have to adjust this field on every new release.
Maybe it would make sense to just specify a Minimum with `+kubebuilder:validation:Minimum`?

wejdross: but then we would have to update that minimum with almost every release, time saving is minimal here. Either we disable that or keep like it is imho

zugao: The best approach would be to do the validation into a webhook. In the webhook we could check direclty what versions are available. In that case no need to track this and update each time there is a new version

zugao (approved):
For now it's ok, but I would open a ticket in the backlog to add webhooks for all services that we do version validation

apis/vshn/v1/dbaas_vshn_keycloak.go
TheBigLee: As we see with this version, there can be potentially breaking changes that make never version non-working at all (eg. changed values in the helm chart).
Therefore we actually should keep this list hardcoded with all versions tested and verified by us (and Inventage)

Kidswiss: I agree, that was my concern with this from the beginning.

Kidswiss (approved):

PR #197, opened 2024-07-16
Add feature to allow setting a custom configuration for keycloak

TheBigLee:
## Summary

* This PR adds support for adding a custom Keycloak configuration

## Checklist

- [ ] Update tests.


pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -583,3 +614,40 @@ exit 0`,
 	}
 	return extraInitContainersMap, nil
 }
+
+func copyConfig(comp *vshnv1.VSHNKeycloak, svc *runtime.ServiceRuntime) error {
```
zugao, line 618: I don't understand what is happening in this go function, you are setting 2 objects in the desired function and then getting from observe, can you briefly explain why?

TheBigLee: I'm observing the configmap in the claim, to be able to copy the content of the configmap to a new configmap in the instance namespace

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -583,3 +614,40 @@ exit 0`,
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        *comp.Spec.Parameters.Service.Configuration,
+			Namespace:   comp.GetInstanceNamespace(),
+			Labels:      configMapClaim.Labels,
+			Annotations: configMapClaim.Annotations,
+		},
+		Data: configMapClaim.Data,
+	}
+
+	return svc.SetDesiredKubeObject(configMapInstance, comp.GetName()+"config-map")
```
wejdross, line 652: My only concern here is as discussed in a meeting:
I'd create a follow up to engineer restart mechanism if customer update the cofigmap. Otherwise changes would be applied on next maintenance, or after we do manual restart after ticket.

wejdross (approved):

apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -120,6 +120,11 @@ type VSHNKeycloakServiceSpec struct {
 	// The themes need to be be placed in the `/themes` directory of the custom image.
 	// the providers need to be placed in the `/providers` directory of the custom image.
 	CustomizationImage VSHNKeycloakCustomizationImage `json:"customizationImage,omitempty"`
+
+	// Configuration can be used to provide a configmap containing configurations for the
+	// keycloak instance. The config is a JSON file based on the keycloak export files.
+	// the referenced configmap, must have the configuration in a field called `keycloak-config.json`
+	Configuration *corev1.LocalObjectReference `json:"configuration,omitempty"`
```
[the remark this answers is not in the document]
TheBigLee: Done

apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -120,6 +120,11 @@ type VSHNKeycloakServiceSpec struct {
 	// The themes need to be be placed in the `/themes` directory of the custom image.
 	// the providers need to be placed in the `/providers` directory of the custom image.
 	CustomizationImage VSHNKeycloakCustomizationImage `json:"customizationImage,omitempty"`
+
+	// Configuration can be used to provide a configmap containing configurations for the
+	// keycloak instance. The config is a JSON file based on the keycloak export files.
+	// The referenced configmap, must have the configuration in a field called `keycloak-config.json`
+	Configuration *string `json:"configuration,omitempty"`
```
Kidswiss, line 127: I assume the config cannot contain any sensitive data?

If it can, then I would rather go with a secretRef here instead of config maps.

Kidswiss (changes requested):
Just one small thing

apis/vshn/v1/dbaas_vshn_keycloak.go
TheBigLee: I'm actually not 100% sure about that. I'll have to check this with Inventage.

zugao (approved):
LGTM, before merging figure out with Inventage to use a secret or cm.

apis/vshn/v1/dbaas_vshn_keycloak.go
TheBigLee: I've improved the code to allow for specifying env variables via a {{secret}} that can then be referenced in the configuration in the {{configMap}}

apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -120,6 +120,17 @@ type VSHNKeycloakServiceSpec struct {
...
+
+	// CustomEnvVariables can be used to provide custom environment variables from a
+	// provided secret for the keycloak instance. The environment variables provided
+	// can for example be used in the custom JSON configuration provided in the `Configuration`
+	// field with `$(env:<ENV_VAR_NAME>:-<some_default_value>)`
+	CustomEnvVariables *string `json:"customEnvVariables,omitempty"`
```
zugao, line 133: I would add `SecretRef` to this name.

zugao (approved):
Just one thing, rest looks good. Please test this in lab  carefully, for a new instance and for an existing instance by deleting or adding configuration configMap/secret.

apis/vshn/v1/dbaas_vshn_keycloak.go
TheBigLee: I thought about this as well, but then the behaviour would be inconsistent. I would either leave it as is, or change both to refs (SecretReference and LocalObjectReference respectively)

Kidswiss: I would actually change the naming of the variables, so it's consistent with what we already have for the alertmanager configs: https://github.com/vshn/appcat/blob/master/apis/vshn/v1/common_types.go#L159

So everything that's a reference should contain `ref`. I would even go so far as to call the other field `CustomConfigurationTemplateRef` because it's a template.

TheBigLee: I've renamed the fields to include the `ref` part.
However, I disagree with the `CustomConfigurationTemplateRef`. This is not really a template, but a configuration itself.

Kidswiss: I mean it kinda is, you can add variables that get replaced by actual values. Sounds very templaty to me :)

But I see that it might get too long then. So it's fine for me.

Kidswiss (approved):

PR #198, opened 2024-07-17
Don't return an error on deletion

Kidswiss:
## Summary

This blocks the deletion of namespaces.

## Checklist

- [ ] Update tests.


Kidswiss:
> There are more places where we through this error. Is this enough to resolve the problem?

Yes, the new API Server implementation defaults to this noop if the function is not implemented for any given virtual resource. And delete is currently not implemented in any of them.

zugao (approved):
LGTM

PR #199, opened 2024-07-18
Fix a bug, where you can't edit an existing VSHNPostgreSQL instance

TheBigLee:
## Summary

There is currently a bug in the AddConnectionDetails function that takes the SGcluster from the `observed` rather then the `desired` state. Since this function is run *after* the `DeployPostgreSQL` function and takes the cluster configuration from the `observed` state and at the end writes it to the `desired` state, it will effectively override every change done to the `SGCluster` object in the `DeployPostgreSQL` function.

The function should rather get the `SGCluster` object from the desired stage to avoid this issue.

## Checklist

- [x] Update tests.


Kidswiss (approved):

zugao (approved):
LGTM

PR #200, opened 2024-07-22
Fix connection detail name

Kidswiss:
## Summary
If the claim-name contains `nextcloud` it will not append `nextcloud` to the `fullName` variable in the helm chart.

However if the claim-name doesn't contain `nextcloud`, then the `fullName` will get a `-nextcloud` appendix, resulting in broken service names.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM.

We might have other services in the same situation

wejdross (approved):

Kidswiss:
> LGTM.
> 
> We might have other services in the same situation

For Minio this has already been mitigated, but maybe Redis is affected

PR #203, opened 2024-07-24
Implement deletionProtection for claims

TheBigLee:
## Summary

* This PR adds a new field `spec.parameters.security.deletionProtection` which is enabled by default, that will prevent the deletion of a claim by accident.
* The deletion protection is handled via webhooks and a user must disable the `deletionProtection` in the claim before being able to delete it.

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_keycloak.go
+++ b/apis/vshn/v1/dbaas_vshn_keycloak.go
@@ -17,6 +17,7 @@ import (
 //go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.service.properties.postgreSQLParameters.default={})"
 //go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.tls.default={})"
 //go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.backup.default={})"
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.security.default={})"
 
 // +kubebuilder:object:root=true
 
@@ -329,4 +330,8 @@ func (v *VSHNKeycloak) GetPDBLabels() map[string]string {
 	return map[string]string{
 		"app.kubernetes.io/name": "keycloakx",
 	}
+}
+
+func (v *VSHNKeycloak) GetSecurity() *Security {
+	return &v.Spec.Parameters.Security
 }
--- a/pkg/controller/postgres/deletion_protection.go
+++ b/pkg/controller/postgres/deletion_protection.go
@@ -3,19 +3,14 @@ package postgres
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"sort"
 	"strconv"
-	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	logging "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/go-logr/logr"
 	"github.com/vshn/appcat/v4/pkg/common/jsonpatch"
-	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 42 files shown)


apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -17,6 +17,7 @@ import (
 //go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.service.properties.postgreSQLParameters.default={})"
 //go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.tls.default={})"
 //go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.backup.default={})"
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_vshnkeycloaks.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.security.default={})"
```
Kidswiss, line 20: Probably add this to the API template as well...

pkg/controller/webhooks/common.go
```diff
@@ -0,0 +1,17 @@
+package webhooks
```
Kidswiss, line 1: Maybe call the file `claim_deletionprotection.go` or something. Common sounds like it's some sort of library for various common stuff.

pkg/controller/webhooks/keycloak.go
```diff
@@ -0,0 +1,78 @@
+package webhooks
```
Kidswiss, line 1: Maybe we could template this as well for new services? Like we do for the CRDs?

pkg/controller/webhooks/keycloak.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil, nil
+}
+
+// ValidateUpdate implements webhook.CustomValidator so a webhook will be registered for the type
+func (r *KeycloakWebhookHandler) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+	return nil, nil
+}
+
+// ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type
+func (r *KeycloakWebhookHandler) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
Kidswiss, line 60: I just realized that this function takes a `runtime.Object`. There's potential to generalize this away as well.

We could have a default webhook handler that implements this validation function. But instead of casting to concrete a concrete type, we create a new interface that returns the value of the boolean (`ProtectionChecker` or something), that we can use to cast to.

Then each of the concrete handlers we have can "inherit" this default handler and override the functions if necessary.  A similar approach has been used with the API Server: https://github.com/vshn/appcat/blob/master/pkg/apiserver/vshn/mariadb/backup.go#L25. Here you can see that the `noop.Noop` handler is an anonymous field of the `vshnMariaDBBackupStorage` handler. Which acts somewhat like inheritance in other languages.

There are various pros to this:
* Get rid of the same code over and over
* The webhook template for new services can be much smaller
* We have the code once and if it needs changes we need to change it once for all services
* Adding a new service to the webhooks is easy as "inheriting" the default handler already does 99% of the work for you

Kidswiss (changes requested):
Feel free to schedule a call to discuss the default handler idea in more detail.

apis/vshn/v1/dbaas_vshn_keycloak.go
TheBigLee: Done

pkg/controller/webhooks/keycloak.go
TheBigLee: I've refactored the code as discussed

Kidswiss (changes requested):

pkg/controller/webhooks/default_handler.go
```diff
@@ -0,0 +1,67 @@
...
+		client:    mgrClient,
+		log:       logger,
+		withQuota: withQuota,
+		obj:       obj,
+	}
+}
+
+// ValidateCreate implements webhook.CustomValidator so a webhook will be registered for the type
+func (r *DefaultWebhookHandler) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	return nil, nil
```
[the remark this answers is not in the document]
TheBigLee: Yes. Haven't come around to do it yet though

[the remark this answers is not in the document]
zugao: What these have to do with deletion protection? It's difficult to follow the logic when random logic is being added. We need to create separate PRs for that matter.

pkg/controller/webhooks/default_handler.go
```diff
@@ -0,0 +1,263 @@
...
+	comp, ok := obj.(common.Composite)
+	if !ok {
+		return nil, fmt.Errorf("provided manifest is not a valid " + r.gk.Kind + " object")
+	}
+
+	allErrs = GetClaimDeletionProtection(comp.GetSecurity(), allErrs)
+
+	if len(allErrs) != 0 {
+		return nil, apierrors.NewInvalid(
+			nextcloudGK,
```
zugao, line 157: I suppose this is wrong, I guess this code should be generic for any service.

pkg/controller/webhooks/keycloak_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/stretchr/testify/assert"
+	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	"github.com/vshn/appcat/v4/pkg"
+	"github.com/vshn/appcat/v4/pkg/common/utils"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestSetupKeycloakWebhookHandlerWithManager_ValidateDelete(t *testing.T) {
```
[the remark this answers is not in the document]
zugao: Second this

pkg/controller/webhooks/default_handler.go
```diff
@@ -0,0 +1,263 @@
...
+func New(mgrClient client.Client, logger logr.Logger, withQuota bool, obj runtime.Object, name string, gk schema.GroupKind, gr schema.GroupResource, length int) *DefaultWebhookHandler {
+
+	return &DefaultWebhookHandler{
+		client:    mgrClient,
+		log:       logger,
+		withQuota: withQuota,
+		obj:       obj,
+		gk:        gk,
+		gr:        gr,
+		length:    length,
```
zugao, line 45: Can we brainstorm this parameter? I understand that for each service the limit is different but we need to figure out a way that all services get the same limit.

pkg/controller/webhooks/default_handler.go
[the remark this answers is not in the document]
zugao: Absolutely. Seeing random length numbers for each service is not very encouraging. We cannot just solve this issue under the disguise of deletion protection.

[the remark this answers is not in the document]
Kidswiss: The initial idea was just to generalize the existing checks though.

pkg/controller/webhooks/default_handler.go
TheBigLee: Fixed

pkg/controller/webhooks/keycloak.go
TheBigLee: Done

pkg/controller/webhooks/common.go
TheBigLee: Done

pkg/controller/webhooks/keycloak_test.go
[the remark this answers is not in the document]
TheBigLee: Done

pkg/controller/webhooks/redis.go
```diff
@@ -147,7 +147,22 @@ func (r *RedisWebhookHandler) ValidateUpdate(ctx context.Context, oldObj, newObj
 
 // ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type
 func (r *RedisWebhookHandler) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
-	// NOOP for now
+	allErrs := field.ErrorList{}
```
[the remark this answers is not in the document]
TheBigLee: Done

pkg/controller/webhooks/default_handler.go
TheBigLee: Yes we should talk about this

pkg/controller/postgres/deletion_protection.go
```diff
@@ -40,15 +40,6 @@ func handle(ctx context.Context, inst client.Object, enabled bool, retention int
 		return getPatchObjectFinalizer(log, inst, op)
 	}
 
-	if !controllerutil.ContainsFinalizer(inst, finalizerName) && inst.GetDeletionTimestamp() == nil {
```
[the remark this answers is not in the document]
TheBigLee: Done

pkg/controller/postgres/reconciler.go
```diff
@@ -59,7 +60,7 @@ func (p *XPostgreSQLReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 func (p *XPostgreSQLReconciler) handleDeletionProtection(ctx context.Context, inst *vshnv1.XVSHNPostgreSQL) error {
 	log := logging.FromContext(ctx, "namespace", inst.GetNamespace(), "instance", inst.GetName())
 
-	protectionEnabled := inst.Spec.Parameters.Backup.DeletionProtection
```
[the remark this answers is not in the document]
TheBigLee: Done

pkg/controller/webhooks/postgresql.go
```diff
@@ -179,7 +179,23 @@ func (p *PostgreSQLWebhookHandler) ValidateUpdate(ctx context.Context, oldObj, n
 
 // ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type
 func (p *PostgreSQLWebhookHandler) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
-	// NOOP for now
+	allErrs := field.ErrorList{}
```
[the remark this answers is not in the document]
TheBigLee: Done

zugao (approved):
LGTM. Please resolve all Simon's concerns, also please make sure we don't block anything with the new length parameter by testing all service in lab cluster (especially create op).

zugao (approved):
LGTM

Kidswiss (approved):

PR #204, opened 2024-07-24
Add PDB 

wejdross:
## Summary

* to prevent downtime caused by software error after maintenance we need PDBs for our HA databases

## Checklist

- [x] Update tests.


keycload.yaml
```diff
@@ -0,0 +1,201 @@
+apiVersion: helm.crossplane.io/v1beta1
```
Kidswiss, line 1: Leftover?

pkg/comp-functions/functions/common/interfaces.go
```diff
@@ -39,3 +39,15 @@ type AllowedNamespaceGetter interface {
...
 }
+
+// Required to get info required for alerting.
+type Alerter interface {
+	GetVSHNMonitoring() vshnv1.VSHNMonitoring
+	GetInstanceNamespace() string
+}
+
+// minimalist PDB interface
+type PDB interface {
```
Kidswiss, line 50: Both of these functions are already part of the `InfoGetter` interface.

If you want to keep them separate, you could remove those two functions from the `InfoGetter` interface and then add the `PDB` interface to `InfoGetter`. This way we'll have the definition exactly once.

pkg/comp-functions/functions/common/pbp.go
```diff
@@ -0,0 +1,78 @@
...
+	ptr "k8s.io/utils/ptr"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	fnproto "github.com/crossplane/function-sdk-go/proto/v1beta1"
+	v1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+)
+
+func AddPDBSettings(obj client.Object) func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
```
Kidswiss, line 20:
```suggestion
func AddPDBSettings(comp client.Object) func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
```

Let's call it comp to be consistent with the rest of the functions.

pkg/comp-functions/functions/common/pbp.go
```diff
@@ -0,0 +1,78 @@
...
+
+		log := controllerruntime.LoggerFrom(ctx)
+
+		err := svc.GetObservedComposite(obj)
+		if err != nil {
+			return runtime.NewFatalResult(fmt.Errorf("can't get composite: %w", err))
+		}
+		pdb, ok := obj.(PDB)
+		if !ok {
+			return runtime.NewWarningResult(fmt.Sprintf("Type %s doesn't implement PDB interface", reflect.TypeOf(obj).String()))
```
Kidswiss, line 31: I'd also return a fatal here, as that's an error that could only happen if the XRD is wrongly implemented.

pkg/comp-functions/functions/common/pbp.go
```diff
@@ -0,0 +1,78 @@
...
+
+	fnproto "github.com/crossplane/function-sdk-go/proto/v1beta1"
+	v1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+)
+
+func AddPDBSettings(obj client.Object) func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
+	return func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
+
+		log := controllerruntime.LoggerFrom(ctx)
```
Kidswiss, line 23: You can use the logger from `svc`. It will have the right context about the current function.


```suggestion
		log := svc.Log
```

pkg/comp-functions/functions/common/pbp.go
```diff
@@ -0,0 +1,78 @@
...
+			min.IntVal = 1
+		} else {
+			// no matter if there are 3,5,10 instances, 2 should be the minimum to keep HA working
+			min.IntVal = 2
+		}
+
+		emptyMap := map[string]string{}
+
+		// to add new case, simply print labels, for example `kubectl -n vshn-mariadb-something-123sa get pod mariadb-something-123sa-0 -o json | jq .metadata.labels` and add the label to the map
+		switch obj.(type) {
```
Kidswiss, line 57: This won't scale, ideally the comp should know the necessary labels and return them as part of the `PDB` interface.

pkg/comp-functions/functions/common/pbp.go
```diff
@@ -0,0 +1,78 @@
...
+		case *v1.VSHNKeycloak:
+			emptyMap["app.kubernetes.io/name"] = obj.GetLabels()["crossplane.io/composite"]
+		case *v1.VSHNPostgreSQL:
+			emptyMap["stackgres.io/cluster-name"] = obj.GetLabels()["crossplane.io/composite"] + "-pg"
```
Kidswiss, line 61: For nested services using `obj.GetLabels()["crossplane.io/composite"]` will be very problematic as it will always point to the composite belonging to the claim, which might not have the same name. For example, the `-pg` appendix you have here only applies to PG instances that are nested with Nextcloud or Keycloak. A normal PostgreSQL instance doesn't have that appendix.

Just use `obj.GetName()` that will return the right value because `obj` is our composite anyway.

Kidswiss (changes requested):

pkg/comp-functions/functions/common/interfaces.go
wejdross: I prefer it that way as I added one new function that is usable only by PDB

pkg/comp-functions/functions/common/pbp.go
tobru, line 1: Is there a reason why this file was name `pbp.go`? I would expect that this is name `pdb.go`

wejdross: I would expect that as well, I have no idea why I misstyped that in the first place

pkg/comp-functions/functions/common/pdb.go
```diff
@@ -0,0 +1,64 @@
...
+	return func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
+
+		log := svc.Log
+
+		err := svc.GetObservedComposite(comp)
+		if err != nil {
+			return runtime.NewFatalResult(fmt.Errorf("can't get composite: %w", err))
+		}
+		pdb, ok := comp.(PDB)
+		if !ok {
```
zugao, line 28: It would be great to catch this problem at build time and not at runtime. I thought about a solution here but nothing comes into my mind. @Kidswiss maybe you have an idea here?

pkg/comp-functions/functions/common/interfaces.go
Kidswiss: With my suggestion it would still be the same, we'd just get rid of the duplication.

pkg/comp-functions/functions/common/pdb.go
Kidswiss: I tried to think of something during my first review, unfortunately I wasn't able to get a working idea.

I'll have another look now that all my other remarks have been implemented.

pkg/comp-functions/functions/common/interfaces.go
wejdross: not, because InfoGetters started to complain that they do not implement specific method:
```
cannot use comp (variable of type *"github.com/vshn/appcat/v4/apis/vshn/v1".VSHNNextcloud) as common.Composite value in argument to common.BootstrapInstanceNs: *"github.com/vshn/appcat/v4/apis/vshn/v1".VSHNNextcloud does not implement common.Composite (missing method GetPDBLabels)
```

pkg/comp-functions/functions/common/pdb.go
```diff
@@ -0,0 +1,64 @@
...
+				Namespace: pdb.GetInstanceNamespace(),
+			},
+		}
+		min := intstr.IntOrString{}
+		if pdb.GetInstances() == 2 {
+			// 1 working instance is still better than nothing
+			min.IntVal = 1
+		} else {
+			// no matter if there are 3,5,10 instances, 2 should be the minimum to keep HA working
+			min.IntVal = 2
```
zugao, line 49: I think it would be better to go with percentage as explained here. Check out [this](It would be great to catch this problem at build time and not at runtime. I thought about a solution here but nothing comes into my mind. @Kidswiss maybe you have an idea here?)

pkg/comp-functions/functions/common/interfaces.go
wejdross: with such separation it will be easier for us to catch issues imho as compiler tells You there's a problem

pkg/comp-functions/functions/common/pdb.go
Kidswiss: > I think it would be better to go with percentage as explained here. Check out [this](It would be great to catch this problem at build time and not at runtime. I thought about a solution here but nothing comes into my mind. @Kidswiss maybe you have an idea here?)

Did you want to paste a link in there? :D

wejdross:
@zugao @Kidswiss I just pushed new commit doing a small cleanup with interfaces and enabling it for Minio. By the way, compiler catches if type implement the interface, I could probably delete that check, I rather want to be double sure here

pkg/comp-functions/functions/common/pdb.go
Kidswiss: Huh... okay it's actually rather simple:

Make the `PDB` interface also implement `client.Object`:

```go
type PDB interface {
	GetInstanceNamespace() string
	GetInstances() int
	GetPDBLabels() map[string]string
	client.Object
}
```

Accept `PDB` in the `AddPDBSettings` function:

```go
func AddPDBSettings(comp PDB) func(ctx context.Context, svc *runtime.ServiceRuntime) *fnproto.Result {
```

Remove the type assert and rename all `pdb` variables to `comp`. No more type assertion needed and the compiler will scream at you if you forget to implement the interface.

apis/vshn/v1/dbaas_vshn_keycloak.go
```diff
@@ -324,3 +324,9 @@ func (v *VSHNKeycloak) GetMonitoring() VSHNMonitoring {
 func (v *VSHNKeycloak) GetInstances() int {
 	return v.Spec.Parameters.Instances
 }
+
+func (v *VSHNKeycloak) GetPDBLabels() map[string]string {
```
Kidswiss, line 328: Can you please add this to `hack/bootstrap/template/api.txt`?

Kidswiss:
> @zugao @Kidswiss I just pushed new commit doing a small cleanup with interfaces and enabling it for Minio. By the way, compiler catches if type implement the interface, I could probably delete that check, I rather want to be double sure here

I've added a comment how to make it work: https://github.com/vshn/appcat/pull/204#discussion_r1692659935

pkg/comp-functions/functions/common/pdb.go
zugao: hmm something is wrong with my comment. I updated it

pkg/comp-functions/functions/common/pdb.go
zugao: perfect. Awesome. Thank for this :)

pkg/comp-functions/functions/common/pdb.go
Kidswiss: Yeah, I guess we could just unconditionally set it to 50% for all our services, makes it easier and it will always ensure half are available.

zugao:
> By the way, compiler catches if type implement the interface

Since you get already the PDB object in the function there's no need to check for anything else at runtime. It should work and in worse case scenario (which should not happen) the application will panick

zugao (approved):
Mow looks great and clean

pkg/comp-functions/functions/common/pdb.go
wejdross: the whole point I wanted to keep it in different interface was that not all services are ready for HA from the beginning, not all services even have a field `instances` in their claim. Now it's using InfoGetter{} as mentioned in the beginning and I completely got rid of PDB{} interface

Kidswiss: The idea of my comment wasn't to get rid of the `PDB` interface completely, just reduce the amount of duplication.

> not all services even have a field instances in their claim. 

In the long run we want all our service to have HA as well.

pkg/comp-functions/functions/common/pdb.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		log.Info("Checking if PDB is needed", "service", comp.GetName(), "instances", comp.GetInstances())
+
+		if comp.GetInstances() < 2 {
+			return runtime.NewNormalResult("Not HA, no pdb needed")
+		}
+		log.Info("HA detected, adding pdb", "service", comp.GetName())
+
+		min := intstr.IntOrString{IntVal: int32(comp.GetInstances()) / 2}
```
zugao, line 32: 1. I'm not sure this will work for uneven numbers.
2. I get this warning in my IDE for variable min:

```Variable 'min' collides with the 'builtin' function 
 Inspection info: Reports declarations of variables, arguments or functions that overlap with the built-in or reserved keyword.
If you receive this error then your code might not be as explicit as possible and might confuse other users.
Example:
type byte struct{}
type string interface{}
Types byte and string collide with the built-in type aliases. Therefore, they will be highlighted. Consider renaming such declarations```

Kidswiss, line 32: If the string doesn't work, can we at least make sure that we always round up here? Then we'll have the same functionality as using `50%`

Kidswiss (changes requested):

zugao (changes requested):
Needs a bit more rework

pkg/comp-functions/functions/common/pdb.go
wejdross: ok, I found the issue with `50%`, Type: intval is default one, when setting a strVal: "50%", You must add explicitly Type: strVal

zugao (approved):
From my side now it's ok. Please resolve all Simon's concerns and make sure you don't go back to casting approach.

Kidswiss (approved):

PR #206, opened 2024-07-30
Fix bucket naming clashes for cloudscale and exoscale

Kidswiss:
## Summary

If a nested services needed more than one bucket, the provisioning would fail. That's because the bucket CRs for cloudscale and exoscale patched value of the `crossplane.io/composite` label to the `metadata.name` of the resulting objects. As this will always be the name of the top most composite, there were name clashes.

This change is implemented in a backward compatible fashion. If there are already existing CRs, it will simply take their name. If the CRs are freshly created, the new naming will be used.

Also the naming follows what has been implement for the minio buckets. It takes the value of the `bucketName` as the name of the CRs. As bucket names should be unique per provider anyway.

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/buckets/cloudscalebucket/cloudscalebucket.go
```diff
@@ -0,0 +1,141 @@
...
+
+	err = populateEndpointConnectionDetails(svc)
+	if err != nil {
+		return runtime.NewFatalResult(err)
+	}
+
+	return nil
+}
+
+func addBucket(svc *runtime.ServiceRuntime, bucket *appcatv1.ObjectBucket, config string) error {
```
TheBigLee, line 60: This function is identical to the `addBucket` function in the `exoscalebucket` package. It would probably make sense to generalize this function.

pkg/comp-functions/functions/buckets/cloudscalebucket/cloudscalebucket.go
```diff
@@ -0,0 +1,141 @@
...
+
+	objName := util.GetBucketObjectName(svc, bucket, bucketResName, mb.DeepCopy())
+
+	mb.ObjectMeta.Name = objName
+	mb.Spec.ForProvider.CredentialsSecretRef.Name = objName
+
+	return svc.SetDesiredComposedResourceWithName(mb, bucketResName)
+}
+
+func addUser(svc *runtime.ServiceRuntime, bucket *appcatv1.ObjectBucket, config string) error {
```
TheBigLee, line 89: This function also has a huge overlap (although not identical). I'm not really sure if some generalization would make sense here though.

pkg/comp-functions/functions/buckets/cloudscalebucket/cloudscalebucket.go
Kidswiss: They are not identical, one creates a `cloudscale` bucket and the other creates an `exoscale` bucket.

TheBigLee: True. Had a closer look and indeed it's not really feasible

TheBigLee (approved):

PR #213, opened 2024-08-07
Add support for PostgreSQL 15 and 16 on Exoscale

TheBigLee:
## Summary

* This PR adds support for PostgreSQL 15 and 16 on Exoscale.

## Checklist

- [ ] Update tests.


apis/exoscale/v1/dbaas_exoscale_postgresql.go
```diff
@@ -58,11 +58,11 @@ type ExoscalePostgreSQLParameters struct {
 type ExoscalePostgreSQLServiceSpec struct {
 	ExoscaleDBaaSServiceSpec `json:",inline"`
 
-	// +kubebuilder:validation:Enum="14"
-	// +kubebuilder:default="14"
+	// +kubebuilder:validation:Enum="14";"15";"16"
```
zugao, line 61: I think we could remove this validation, everything depends on exoscale here, we just pass the value along. @Kidswiss what do you think?

PR #214, opened 2024-08-09
Support RBAC management for instances

TheBigLee:
## Summary
* This PR adds support to specify custom groups and users that should have access to the instance namespace.

## Checklist

- [ ] Update tests.


pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -139,7 +139,7 @@ func createInstanceNamespace(serviceName, compName, claimNamespace, instanceName
 	return svc.SetDesiredKubeObjectWithName(ns, instanceNamespace, namespaceResName)
 }
 
-func createNamespacePermissions(instance string, instanceNs string, namespaceResName string, svc *runtime.ServiceRuntime) error {
+func createNamespacePermissions(instance string, instanceNs string, namespaceResName string, comp Composite, svc *runtime.ServiceRuntime) error {
```
Kidswiss, line 142: We could just accept the `Security` object directly here.

pkg/comp-functions/functions/common/instance_namespace.go
```diff
@@ -158,7 +158,38 @@ func createNamespacePermissions(instance string, instanceNs string, namespaceRes
 		return fmt.Errorf("cannot get claim namespace: %w", err)
 	}
 
-	if org == "" {
+	security := comp.GetSecurity()
```
Kidswiss, line 161: Then you can avoid this line.

Kidswiss (changes requested):
Just a small nit

pkg/comp-functions/functions/common/instance_namespace.go
TheBigLee: True. This makes more sense.

pkg/comp-functions/functions/common/instance_namespace.go
TheBigLee: I've removed it

Kidswiss (approved):

PR #220, opened 2024-08-20
Improve CI Pipeline on PRs

Kidswiss:
## Summary

## Checklist

- [ ] Update tests.


.github/workflows/pr.yml
```diff
@@ -0,0 +1,191 @@
...
+  create-release:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for bug label
+        if: contains(github.event.pull_request.labels.*.name, 'bug')
+        id: bug
+        run: |
+          echo "set=true" >> $GITHUB_OUTPUT
+      - name: Check for bug label
```
zugao, line 104:
```suggestion
      - name: Check for enhancement label
```

zugao (approved):
LGTM just a couple of small things. I wonder if you tried to merge the PR when there are no labels or a wrong labels, what would happen?

Kidswiss:
> I wonder if you tried to merge the PR when there are no labels or a wrong labels, what would happen?

It fails, but there's also an action that checks that at least one of the labels is there.

PR #224, opened 2024-08-23
Fix pointer leaks by using generics

Kidswiss:
## Summary

We re-used the same pointer for every single reconcile of any given service type. This could lead to leaking settings between the instances. Resulting in instances enabling features that aren't actually enabled in the claim belonging to the instance.

This change gets rid of the pointer by leveraging generics and reflect.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/mailgun_alerting.go
+++ b/pkg/comp-functions/functions/common/mailgun_alerting.go
@@ -17,41 +17,40 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func MailgunAlerting(obj client.Object) func(ctx context.Context, svc *runtime.ServiceRuntime) *xfnproto.Result {
-	return func(ctx context.Context, svc *runtime.ServiceRuntime) *xfnproto.Result {
-		log := controllerruntime.LoggerFrom(ctx)
-		log.Info("Starting mailgun-alerting function")
+func MailgunAlerting[T client.Object](ctx context.Context, obj T, svc *runtime.ServiceRuntime) *xfnproto.Result {
+	log := controllerruntime.LoggerFrom(ctx)
+	log.Info("Starting mailgun-alerting function")
 
-		err := svc.GetObservedComposite(obj)
-		if err != nil {
-			return runtime.NewFatalResult(fmt.Errorf("Can't get composite: %w", err))
-		}
-		alertConfig, ok := obj.(Alerter)
-		if !ok {
-			return runtime.NewWarningResult(fmt.Sprintf("Type %s doesn't implement Alerter interface", reflect.TypeOf(obj).String()))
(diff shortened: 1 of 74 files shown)


pkg/comp-functions/functions/common/mailgun_alerting.go
```diff
@@ -17,41 +17,40 @@ import (
...
-		log.Info("Deploying AlertmanagerConfig for mail alerting...")
-		err = deployAlertmanagerConfig(ctx, name, email, instanceNamespace, svc)
-		if err != nil {
-			return runtime.NewFatalResult(fmt.Errorf("Can't deploy AlertmanagerConfig "+name+"-alertmanagerconfig-mailgun for mail alerting: %w", err))
-		}
-
-		log.Info("Finishing mailgun-alerting function with NewNormal")
+func MailgunAlerting[T client.Object](ctx context.Context, obj T, svc *runtime.ServiceRuntime) *xfnproto.Result {
+	log := controllerruntime.LoggerFrom(ctx)
+	log.Info("Starting mailgun-alerting function")
```
wejdross, line 22: this is really cool, much cleaner now

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -178,6 +193,45 @@ func (m Manager) RunFunction(ctx context.Context, req *fnv1beta1.RunFunctionRequ
...
+
+func (m *Manager) getStepName(step any) string {
+	v := reflect.ValueOf(step).Field(0)
+	if v.Kind() != reflect.String {
+		panic(fmt.Sprintf("%s not a string, can't get step name", v.Kind()))
+	}
+
+	return v.String()
+}
+
```
zugao, line 215: Can you please add here the reason as to why we use reflection. Or maybe in any other function you see fit.

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -178,6 +193,45 @@ func (m Manager) RunFunction(ctx context.Context, req *fnv1beta1.RunFunctionRequ
 	return sr.GetResponse()
 }
 
+func (m *Manager) extractSteps(service any) []any {
```
zugao, line 196: I fell like we overcomplicated things here, the generics should have saved us from reflection....

zugao (approved):
LGTM

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: Done.

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: As discussed, generics can't help in this situation unfortunately...

PR #225, opened 2024-08-26
Add support for SPKS and implement sts resizing

TheBigLee:
## Summary

* This PR adds support for the SPKS MariaDB and Redis offerings
* It also implements the sts deletion logic for PVC resizing using composition functions

## Checklist

- [ ] Update tests.


Kidswiss (approved):

zugao:
Did we reuse the scripts or they have been just implemented in this PR? I feel like this PR is full of bombs, have you tested it throughly?

TheBigLee:
> Did we reuse the scripts or they have been just implemented in this PR? I feel like this PR is full of bombs, have you tested it throughly?

We did reuse them from the vshnredis, but have been adjusted slightly to work with this setup.
It has been tested thoroughly. What bombs are you refering to?

wejdross (approved):

pkg/comp-functions/functions/spksmariadb/pvcresize.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	if patch {
+		err = addDeletionJob(svc, comp, newSize, release.GetName())
+		if err != nil {
+			return runtime.NewFatalResult(fmt.Errorf("cannot create the deletion job: %w", err))
+		}
+		// We pause the release at this point to make sure that provider-helm doesn't update the
+		// release until the deletion job removed the sts
+		release.SetAnnotations(map[string]string{
```
Kidswiss, line 140: I'd probably set the annotation via the script as well. There's still potential for a very narrow race condition here.

pkg/comp-functions/functions/spksmariadb/script/recreate.sh
```diff
@@ -0,0 +1,41 @@
...
+  # We try to delete the sts and wait for 5s. On APPUiO it can happen that the
+  # deletion with orphan doesn't go through and the sts is stuck with an orphan finalizer.
+  # So if the delete hasn't returned after 5s we forcefully patch away the finalizer.
+  kubectl -n "$namespace" delete sts "$name" --cascade=orphan --ignore-not-found --wait=true --timeout 5s || true
+  kubectl -n "$namespace" patch sts "$name" -p '{"metadata":{"finalizers":null}}' || true
+  # Upause the release so that the sts is recreated. We pause the release to avoid provider-helm updating the release
+  # before the sts is deleted.
+  # Then we first patch the siye to garbage and afterwards to the right size to enforce an upgrade
+  echo "Triggering sts re-creation"
+  kubectl annotate release "$release" "crossplane.io/paused-"
```
Kidswiss, line 27: Make this a `trap` so it always triggers, even if something else fails. Otherwise we might end up with paused releases.

pkg/comp-functions/functions/spksmariadb/script/recreate.sh
```diff
@@ -0,0 +1,41 @@
...
+release="$RELEASE_NAME"
+
+echo "Checking if the PVC sizes match"
+# Check if delete is necessary
+found=$(kubectl -n "$namespace" get sts "$name" -o json --ignore-not-found)
+
+foundsize=$(echo -En "$found" | jq -r '.spec.volumeClaimTemplates[] | select(.metadata.name=="data") | .spec.resources.requests.storage')
+
+if [[ $foundsize != "$size" ]]; then
+  echo "PVC sizes don't match, deleting sts"
```
Kidswiss, line 17: I would add the pause annotation at this point in the logic. This way it's contained within itself and less likely to run into issues.

Kidswiss (changes requested):
I like the approach.

I've added a few comments about it.

Also, it would be great if this was generic for any release, because right now the code exists more or less identical for two spks services and still with the race condtion for the other services.

pkg/comp-functions/functions/spksmariadb/pvcresize.go
wejdross: I think that setting it in script has exactly the same risk of that race condition. Here it's maybe even smaller as annotation will be set immediately, even before job starts.

pkg/comp-functions/functions/spksmariadb/script/recreate.sh
```diff
@@ -0,0 +1,41 @@
...
+  # deletion with orphan doesn't go through and the sts is stuck with an orphan finalizer.
+  # So if the delete hasn't returned after 5s we forcefully patch away the finalizer.
+  kubectl -n "$namespace" delete sts "$name" --cascade=orphan --ignore-not-found --wait=true --timeout 5s || true
+  kubectl -n "$namespace" patch sts "$name" -p '{"metadata":{"finalizers":null}}' || true
+  # Upause the release so that the sts is recreated. We pause the release to avoid provider-helm updating the release
+  # before the sts is deleted.
+  # Then we first patch the siye to garbage and afterwards to the right size to enforce an upgrade
+  echo "Triggering sts re-creation"
+  kubectl annotate release "$release" "crossplane.io/paused-"
+  kubectl patch release "$release" --type merge -p "{\"spec\":{\"forProvider\":{\"values\":{\"persistence\":{\"size\":\"foo\"}}}}}"
```
wejdross, line 28: do we still need that `foo`?

wejdross (commented):
I'd add PrometheusRule as well to check if annotation `crossplane.io/paused` is set for longer than 15 minutes, so we don't end up with partially broken resources and catch early issues.

pkg/comp-functions/functions/spksmariadb/pvcresize.go
Kidswiss: I wouldn't bet on that, we don't know when crossplane decides to actually reconcile this. It could be that it gets reconciled after the job has spawned. 

That's why I would move it to the script. `kubectl` will return after it sent the request directly to the API Server, which will greatly reduce the risk of this race condition happening.

TheBigLee: As discussed, moving it to the script is not really an option, as both the script and the release will be reconciled at the same time.
I adjusted the script though to check and wait for the annotation on the release.

pkg/comp-functions/functions/spksmariadb/script/recreate.sh
TheBigLee: As mentioned above, this doesn't really work.

pkg/comp-functions/functions/spksmariadb/script/recreate.sh
TheBigLee: Yes. Because provider-helm wouldn't update the release when we just remove the annotation. It needs a change in the values somewhere. This hacky solution looks like the easiest way.

TheBigLee:
> I'd add PrometheusRule as well to check if annotation `crossplane.io/paused` is set for longer than 15 minutes, so we don't end up with partially broken resources and catch early issues.

This is obsolete, as we use `trap` to ensure the value is removed

pkg/comp-functions/functions/spksmariadb/pvcresize.go
```diff
@@ -0,0 +1,260 @@
...
+
+func addStsObserver(svc *runtime.ServiceRuntime, comp *spksv1alpha1.CompositeMariaDBInstance) error {
+	statefulset := &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "mariadb",
+			Namespace: comp.GetName(),
+		},
+	}
+
+	providerConfigRef := comp.GetLabels()["service.syn.tools/cluster"]
```
wejdross, line 162: I'd add here a check if this key exists in map otherwise You'll be working with an empty string

pkg/comp-functions/functions/spksmariadb/pvcresize.go
```diff
@@ -0,0 +1,260 @@
...
+								stsRecreateScript,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	providerConfigRef := comp.GetLabels()["service.syn.tools/cluster"]
```
wejdross, line 246: here as well

pkg/comp-functions/functions/spksredis/pvcresize.go
```diff
@@ -0,0 +1,268 @@
...
+								stsRecreateScript,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	providerConfigRef := comp.GetLabels()["service.syn.tools/cluster"]
```
wejdross, line 254: here as well

pkg/comp-functions/functions/spksredis/pvcresize.go
```diff
@@ -0,0 +1,268 @@
...
+
+func addStsObserver(svc *runtime.ServiceRuntime, comp *spksv1alpha1.CompositeRedisInstance) error {
+	statefulset := &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "redis-node",
+			Namespace: comp.GetName(),
+		},
+	}
+
+	providerConfigRef := comp.GetLabels()["service.syn.tools/cluster"]
```
wejdross, line 166: here as well

wejdross (approved):
I'm ok with that, just one nitpick. LGTM

pkg/comp-functions/functions/spksmariadb/script/recreate.sh
TheBigLee: Done. Please check again

Kidswiss (approved):

wejdross (approved):

PR #226, opened 2024-08-29
fixing timescaledb extension bug

wejdross:
when timescale extension was set during creation it blocked reconciliation because of broken object name


pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
```diff
@@ -334,7 +334,7 @@ func createSgPostgresConfig(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRun
 		},
 	}
 
-	err := svc.SetDesiredKubeObjectWithName(sgPostgresConfig, comp.GetName()+"-pgconf", "pg-conf")
+	err := svc.SetDesiredKubeObjectWithName(sgPostgresConfig, comp.GetName()+"-"+configResourceName, configResourceName)
```
Kidswiss, line 337: This will change the object name, which will probably result in the deletion of the old object.

I would keep the name as is.

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
wejdross: so after visiting this rabbit hole I think it's safe to proceed now

Kidswiss (approved):

PR #229, opened 2024-09-05
Fix operator upgrade restart instance

zugao:
## Summary

* This PR makes sure that whenever the Stackgres Operator is being updated, the postgres clusters will be upgraded only during the maintenance window and not right away.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/restart.go
+++ b/pkg/comp-functions/functions/vshnpostgres/restart.go
@@ -94,7 +94,7 @@ func getPendingRestart(ctx context.Context, svc *runtime.ServiceRuntime) (time.T
 	return time.Time{}, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/comp-functions/functions/vshnpostgres/restart.go
```diff
@@ -89,6 +94,16 @@ func getPendingRestart(ctx context.Context, svc *runtime.ServiceRuntime) (time.T
 	return time.Time{}, nil
 }
 
+// In case the operator was updated and the PostgreSQL clusters requires immediate restart we will postpone it during the maintenance
```
Kidswiss, line 97:
```suggestion
// In case the operator was updated and the PostgreSQL clusters require a security maintenance, we ensure that it only happens during the maintenance window
```

Kidswiss (approved):

PR #232, opened 2024-09-10
Validate against forbidden fields

wejdross:
## Summary

* Stackgres refuses usage of few configuration options and our customers shouldn't be able to set them. This PR ensures there're no possibility to create instance with broken parameters. 

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/controller/webhooks/postgresql.go
+++ b/pkg/controller/webhooks/postgresql.go
@@ -35,6 +35,24 @@ var (
 
 var _ webhook.CustomValidator = &PostgreSQLWebhookHandler{}
 
+var blocklist = map[string]string{
+	"listen_addresses":      "",
+	"port":                  "",
+	"cluster_name":          "",
+	"hot_standby":           "",
+	"fsync":                 "",
+	"full_page_writes":      "",
+	"log_destination":       "",
+	"logging_collector":     "",
+	"max_replication_slots": "",
+	"max_wal_senders":       "",
+	"wal_keep_segments":     "",
+	"wal_level":             "",
+	"wal_log_hints":         "",
+	"archive_mode":          "",
+	"archive_command":       "",
+}
+
 // PostgreSQLWebhookHandler handles all quota webhooks concerning postgresql by vshn.
 type PostgreSQLWebhookHandler struct {
 	DefaultWebhookHandler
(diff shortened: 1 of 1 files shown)


pkg/controller/webhooks/postgresql.go
```diff
@@ -287,3 +298,146 @@ func validateVacuumRepack(vacuum, repack bool) error {
...
+				Detail:   fmt.Sprintf("Error parsing pgConf: %s", err.Error()),
+				Type:     field.ErrorTypeInvalid,
+				BadValue: pgConfBytes,
+			})
+			return fErros
+		}
+	}
+
+	for key, _ := range pgConf {
+		if key == "listen_addresses" {
```
zugao, line 321: Can you please define a list of forbidden pg parameters on top of this file and instead of having multiple if clauses to just have one? I would also consider create a new go file to separate the concerns and make it a bit more readable

zugao (changes requested):

pkg/controller/webhooks/postgresql.go
Kidswiss: I agree, having a map of forbidden keys would be more readable. You can use the maps key for the field name, so you can efficiently do checks. Then use the value of the map to store the specifics of the error.

It would also be easier to extend or change in the future.

Kidswiss (changes requested):

pkg/controller/webhooks/postgresql.go
zugao: The error seems to be almost the same, only the name is added to a field. I would still go with just a list.

pkg/controller/webhooks/postgresql.go
```diff
@@ -287,3 +298,52 @@ func validateVacuumRepack(vacuum, repack bool) error {
...
+				Field:    "spec.parameters.service.postgresqlSettings",
+				Detail:   fmt.Sprintf("Error parsing pgConf: %s", err.Error()),
+				Type:     field.ErrorTypeInvalid,
+				BadValue: pgConfBytes,
+			})
+			return fErros
+		}
+	}
+
+	blocklist := map[string]string{
```
zugao, line 320: why did you create a map if the values are empty? Please change to a list and move it outside the function at the top of the file. Our code is already very confusing, lets not make it a pain to read.

zugao (dismissed):

pkg/controller/webhooks/postgresql.go
Kidswiss: The idea for the map is that we don't have nested loops, so I would keep it as such.

But I agree to move it to the top.

zugao: You meant that there's no good way to search in the list? I forgot how bad go is in that regard... but there are packages that help with finding an object in the list... I would still have gone with the list... it heart my eyes to see empty strings there, it feels like they will be filled during a for loop....

zugao (approved):
LGTM

pkg/controller/webhooks/postgresql.go
wejdross: finding an element in array is the same operation in each language, finding key in map is O(1) . O(n^2) You suggest would hurt my eyes ;p

PR #235, opened 2024-09-25
Improve PostgreSQL ordering, reduce issues

wejdross:
Fix probably a bug in our ObjectStorage spec

## Checklist

- [ ] Update tests.

### DIFF
--- a/apis/v1/objectstorage_types.go
+++ b/apis/v1/objectstorage_types.go
@@ -18,7 +18,6 @@ const (
 type BucketDeletionPolicy string
 
 //go:generate yq -i e ../generated/appcat.vshn.io_objectbuckets.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.security.default={})"
-
 // +kubebuilder:object:root=true
 // +kubebuilder:printcolumn:name="Bucket Name",type="string",JSONPath=".spec.parameters.bucketName"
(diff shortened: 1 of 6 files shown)


wejdross:
@Kidswiss please take a look at this, seems that it was the issue with our services - causing broken deployment of services. After deleting that line everything seems to work.

Kidswiss:
Can you elaborate what "probably a bug" you're fixing here?

wejdross:
Was our intention to add this .spec.parameters.security to ObjectBuckets? I'd say no - so it's bug, but I can be wrong and miss some information - therefore "probably a bug" :laughing:

Kidswiss:
> Was our intention to add this .spec.parameters.security to ObjectBuckets? I'd say no - so it's bug, but I can be wrong and miss some information - therefore "probably a bug" 😆

`.spec.parameters.security` contains the deletion protection settings for the bucket. They are necessary or they can't delete the buckets anymore, if they use ObjectBucket directly.

Kidswiss (approved):
Please test on lab, so we're sure it doesn't remove any existing sgclusters.

zugao:
Does this PR help with creating 5 or more instances at the same time?

wejdross:
> Does this PR help with creating 5 or more instances at the same time?

yes, it solves the issue

zugao (approved):
LGTM, again as Simon said, please test with existing buckets, we don't want those buckets to be removed accidentely.

PR #236, opened 2024-09-26
Fix dependency issue when using encrypted PVCs for VSHNPostgreSQL

TheBigLee:
## Summary

There was a dependency loop that prevented the creation of VSHNPostgreSQL instances that use encrypted PVCs:
* The SGCluster object can't be created before the luks secret for the PVC is generated
* The luks secret can't be created without getting the instance count of the cluster object

This has been resolved by relying on the instance count from the composite and also waiting for the luks secrets to be present before attempting to access the sgcluster object to be able to override the storageClass for the PVCs

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -995,11 +995,11 @@ func (s *ServiceRuntime) areResourcesReady(names []string) bool {
 	return true
 }
 
-// WaitForDependencies takes two arguments, the name of the main resource, which should be deployed after the dependencies.
+// WaitForObservedDependencies takes two arguments, the name of the main resource, which should be deployed after the dependencies.
 // It also takes a list of names for objects to depend on. It does NOT deploy any objects, but check for their existence.
 // If true is returned it is safe to continue with adding your main object to the desired resources.
 // If the main resource already exists in the observed state it will always return true.
-func (s *ServiceRuntime) WaitForDependencies(mainResource string, dependencies ...string) bool {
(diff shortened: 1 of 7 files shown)


wejdross (approved):
Please test it in LAB with:
- fresh installation
- test behavior of already working instance with encryption enabled (if we have any)

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1011,17 +1011,33 @@ func (s *ServiceRuntime) WaitForDependencies(mainResource string, dependencies .
...
+// WaitForDesiredDependencies takes two arguments, the name of the main resource, which should be deployed after the dependencies.
+// It also takes a list of names for objects to depend on. It does NOT deploy any objects, but check for their existence.
+// If true is returned it is safe to continue with adding your main object to the desired resources.
+// If the main resource already exists in the observed state it will always return true.
+func (s *ServiceRuntime) WaitForDesiredDependencies(mainResource string, dependencies ...string) bool {
+	if _, ok := s.req.Desired.Resources[mainResource]; ok {
+		return true
+	}
+
+	if !s.areResourcesReady(dependencies) {
```
Kidswiss, line 1023: Just a note: this will look at the observed state, not the desired one.

But from a logic perspective a desired resource cannot yet be ready because it's only the desired state, not the actual state. However, this might be a good failsafe, if the object is still existing but is missing from the desired by mistake.

Kidswiss (approved):
LGTM, if it was tested on the lab.

PR #237, opened 2024-09-27
Add providerConfig management

Kidswiss:
## Summary

* Each claim/composite pair can specific their own providerConfig name
* It will be injected into each managed resource
* The providerConfigs need to be correctly provisioned beforehand
* Not setting the label will result in the current defaults being used
* Usage will no longer be wrapped into Kube Objects to avoid deploying them to service clusters
  * Also it's best practices to NOT wrap them, it makes the deprovisioning process a lot smoother
* It's possible to set a label for managed resources that should not get the providerConf from the claim/composite
  * Some managed resources have already been labeled, but by all means not all of them. That should be in the scope of another story

## Checklist

- [x] Update tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,8 +218,8 @@ This will:
 
 If you want to print the diff while using the debugger you can simply do this:
 
-* Star the local debugging session as described
-* Run `make render-diff -e DEBUG=Developmen`
+* Start the local debugging session as described
+* Run `make render-diff -e DEBUG=Development`
 
 NOTE: `crank render` has a 60s timeout, so you might run into it, if your debugging takes longer
 
--- a/README.md
+++ b/README.md
@@ -214,6 +214,15 @@ This will:
 * Run `crank` again with the state already downloaded, to avoid any unintended diffs
 * Use `dyff` to generate the diffs between both results
 
+### Generate diff against live version with debugger
+
+If you want to print the diff while using the debugger you can simply do this:
+
+* Star the local debugging session as described
+* Run `make render-diff -e DEBUG=Developmen`
+
+NOTE: `crank render` has a 60s timeout, so you might run into it, if your debugging takes longer
+
 # Run API Server locally
 To run the API server on your local machine you need to register the IDE running instance with kind cluster.
 This can be achieved with the following guide.
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -46,12 +46,14 @@ var (
 )
 
 const (
-	OwnerKindAnnotation    = "appcat.vshn.io/ownerkind"
-	OwnerVersionAnnotation = "appcat.vshn.io/ownerapiversion"
-	OwnerGroupAnnotation   = "appcat.vshn.io/ownergroup"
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -214,6 +214,15 @@ This will:
 * Run `crank` again with the state already downloaded, to avoid any unintended diffs
 * Use `dyff` to generate the diffs between both results
 
+### Generate diff against live version with debugger
+
+If you want to print the diff while using the debugger you can simply do this:
+
+* Star the local debugging session as described
+* Run `make render-diff -e DEBUG=Developmen`
```
wejdross, line 222: typo here

wejdross: even 2

wejdross (changes requested):
Few small typos

README.md
Kidswiss: I've added the missing `t`, what's the second typo?

README.md
```diff
@@ -214,6 +214,15 @@ This will:
 * Run `crank` again with the state already downloaded, to avoid any unintended diffs
 * Use `dyff` to generate the diffs between both results
 
+### Generate diff against live version with debugger
+
+If you want to print the diff while using the debugger you can simply do this:
+
+* Star the local debugging session as described
```
wejdross, line 221: Should be start :D

Kidswiss: Ah... I thought the two typos were on the same line :D

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1127,7 +1142,60 @@ func (s *ServiceRuntime) UsageOfBy(of, by string) error {
...
+			return true, nil
+		}
+		return false, err
+	}
+
+	resources, err := request.GetObservedComposedResources(s.req)
+	if err != nil {
+		return false, err
+	}
+	res := resources[resource.Name(name)]
```
wejdross, line 1175: I'd check here for empty string or just if the value is set

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1127,7 +1142,60 @@ func (s *ServiceRuntime) UsageOfBy(of, by string) error {
...
+	// we need to clean the objectmeta, or Crossplane will struggle with adding
+	// it to the `resourceRefs` array.
+	usage.ObjectMeta = metav1.ObjectMeta{
+		Name:            usage.GetName(),
+		Annotations:     usage.Annotations,
+		Labels:          usage.Labels,
+		OwnerReferences: usage.OwnerReferences,
+	}
+
+	if len(usage.Spec.ManagementPolicies) == 0 || usage.Spec.ManagementPolicies[0] != xpv1.ManagementActionObserve {
```
wejdross, line 1189: also here, if managementpolicies for some reason will be nil, then You will receive:
```
panic: runtime error: index out of range [0] with length 0
```

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: If a resource with this name doesn't exist, then this function would not reach this.

Because we check for `ErrNotFound` on line 1165 already.

wejdross (approved):

PR #239, opened 2024-10-03
adding delayed removal of xobjectbucket

wejdross:
## Summary

* this PR adds logic to delay deletion of xobjectbucket so we're sure billing run at least once

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/controller/webhooks/xobjectbuckets.go
+++ b/pkg/controller/webhooks/xobjectbuckets.go
@@ -3,6 +3,7 @@ package webhooks
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	appcatv1 "github.com/vshn/appcat/v4/apis/v1"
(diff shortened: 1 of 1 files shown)


Kidswiss:
Did you test, that Crossplane correctly removes the `XObjectBucket` after 61 minutes?

pkg/controller/webhooks/xobjectbuckets.go
```diff
@@ -55,6 +56,15 @@ func (p *XObjectbucketDeletionProtectionHandler) ValidateDelete(ctx context.Cont
 		return nil, fmt.Errorf("object is not valid")
 	}
 
+	creationTimestamp := bucket.GetCreationTimestamp()
+
+	now := time.Now()
+	age := now.Sub(creationTimestamp.Time)
+
+	if age < 61*time.Minute {
+		return nil, fmt.Errorf("XObjectBucket is too young to be deleted")
```
Kidswiss, line 65: NIT: Maybe print how many minutes remaining until it should be deleted.

wejdross:
> Did you test, that Crossplane correctly removes the `XObjectBucket` after 61 minutes?

I tested if xobjectbucket was gone after 61 seconds :D

pkg/controller/webhooks/xobjectbuckets.go
wejdross: sure

Kidswiss:
> > Did you test, that Crossplane correctly removes the `XObjectBucket` after 61 minutes?
> 
> I tested if xobjectbucket was gone after 61 seconds :D

Could you still try it with a larger amount of time? 1 minute is within a single Crossplane reconcile loop.

Also, something interesting to test as well: does it still delete it, if Crossplane gets restarted before the timer runs out?

pkg/controller/webhooks/xobjectbuckets.go
zugao: the phrasing "too young" really does not help. Either print the minutes as suggested above or just something like "XObjectBucket has been created less than 1 hour".

zugao (approved):
LGTM

wejdross:
> Could you still try it with a larger amount of time? 1 minute is within a single Crossplane reconcile loop.
> 
> Also, something interesting to test as well: does it still delete it, if Crossplane gets restarted before the timer runs out?

1. it removes it correctly on small time intervals
2. it removes it correctly on longer intervals like 3 minutes
3. it removes it correctly during longer time intervals with many concurrent restarts of crossplane and providers (just to make some mess)

Kidswiss (approved):
Great!

PR #240, opened 2024-10-04
Add billing to services

zugao:
## Summary

* Added a label to each of our services to allow easy scrape of billing metrics 

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/vshn_minio.go
+++ b/apis/vshn/v1/vshn_minio.go
@@ -2,7 +2,6 @@ package v1
 
 import (
 	"fmt"
-
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -246,4 +245,15 @@ func (v *VSHNMinio) GetPDBLabels() map[string]string {
 
 func (v *VSHNMinio) GetSecurity() *Security {
 	return &v.Spec.Parameters.Security
+}
+
+func (v *VSHNMinio) GetWorkloadPodTemplateLabelsManager() PodTemplateLabelsManager {
+	if v.GetInstances() == 1 {
+		return &DeploymentManager{}
+	}
+	return &StatefulSetManager{}
+}
+
+func (v *VSHNMinio) GetWorkloadName() string {
+	return v.GetName()
 }
--- a/pkg/comp-functions/functions/common/interfaces.go
+++ b/pkg/comp-functions/functions/common/interfaces.go
@@ -18,6 +18,8 @@ type InfoGetter interface {
 	GetSecurity() *vshnv1.Security
 	InstanceNamespaceInfo
 	GetPDBLabels() map[string]string
+	GetWorkloadPodTemplateLabelsManager() vshnv1.PodTemplateLabelsManager
+	GetWorkloadName() string
 }
 
 // InstanceNamespaceInfo provides all the necessary information to create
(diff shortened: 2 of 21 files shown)


apis/vshn/v1/common_types.go
```diff
@@ -261,3 +263,41 @@ func (a *TimeOfDay) AddTime(d time.Duration) TimeOfDay {
 	a.SetTime(a.GetTime().Add(d))
 	return *a
 }
+
+type PodTemplateLabelsManager interface {
+	SetPodTemplateLabels(map[string]string)
+	GetPodTemplateLabels() map[string]string
+	client.Object
+	GetObject() client.Object
+}
```
wejdross, line 272: there's interfaces.go file, please put that interafce in there, preferably into InfoGetter as it affects all of our services

wejdross (changes requested):

wejdross:
build is also failing and needs to be fixed

pkg/comp-functions/functions/common/interfaces.go
```diff
@@ -18,6 +18,8 @@ type InfoGetter interface {
 	GetSecurity() *vshnv1.Security
 	InstanceNamespaceInfo
 	GetPDBLabels() map[string]string
+	GetServiceKubeType() vshnv1.PodTemplateLabelsManager
+	GetServiceKubeTypeName() string
 }
```
TheBigLee, line 23: I would change the name of those functions as "Service" is misleading here

apis/vshn/v1/common_types.go
zugao: Unfortunately I cannot use _intefraces.go_ file as it will generate an `import cycle not allowed` error. I am using the definition of the interface directly in the API therefore I cannot import from the comp-function package.

TheBigLee (commented):
Please also fix the failing build action

pkg/comp-functions/functions/common/interfaces.go
zugao: Can you suggest a better name?

zugao: I think I found a solution :)

apis/vshn/v1/vshn_minio.go
```diff
@@ -247,3 +246,14 @@ func (v *VSHNMinio) GetPDBLabels() map[string]string {
 func (v *VSHNMinio) GetSecurity() *Security {
 	return &v.Spec.Parameters.Security
 }
+
+func (v *VSHNMinio) GetWorkloadPodTemplateLabelsManager() PodTemplateLabelsManager {
+	if v.GetInstances() == 1 {
+		return &DeploymentManager{}
+	}
+	return &StatefulSetManager{}
+}
```
wejdross, line 255: single instance minio is deployment and for multiple it's statefulset?

wejdross (approved):
This code won't work if we have multiple deployments/statefullsets in single instance namespace, but that was never in acceptance criteria, therefore I'll aproove that.

apis/vshn/v1/vshn_minio.go
zugao: yes

zugao:
> This code won't work if we have multiple deployments/statefullsets in single instance namespace, but that was never in acceptance criteria, therefore I'll aproove that.

This is exactly the reason why this PR exists, to filter other deployments/statefulsets. Currently we have 2 deployments for nextcloud, one for the service itself and one for the metrics and with current code we count nextcloud twice into billing. I understand that in the future other deployments and statefulsets might reside into  the instance namespace that HAVE to be counted towards billing but this is not the case right now and we should never develop for something in the future as we don't know how it will look like from requirements point of view.

Kidswiss (approved):

PR #244, opened 2024-10-16
Add deletionProtection for managed crossplane resources

TheBigLee:
## Summary

* This PR adds deletion protection for various managed crossplane resources

## Checklist

- [ ] Update tests.


apis/kubernetes/v1alpha1/doc.go
```diff
@@ -14,9 +14,8 @@ See the License for the specific language governing permissions and
...
-
-// Hub marks this type as a conversion hub.
-func (g *Object) Hub() {
-
-}
+// Package v1alpha1 contains the v1alpha1 group Object resources of the Kubernetes provider.
+// +kubebuilder:object:generate=true
+// +groupName=kubernetes.crossplane.io
+// +versionName=v1alpha1
+package v1alpha1
```
zugao, line 21: why are we moving back from v1alpha2 to v1alpha2?

Kidswiss: SPKS is not yet on the latest provider-kubernetes

pkg/comp-functions/functions/vshnmariadb/user_management.go
```diff
@@ -61,7 +61,8 @@ func addUser(comp common.Composite, svc *runtime.ServiceRuntime, username string
 				"crossplane.io/external-name": username,
 			},
 			Labels: map[string]string{
-				runtime.ProviderConfigIgnoreLabel: "true",
+				runtime.ProviderConfigIgnoreLabel:      "true",
+				"appcat.vshn.io/webhook-allowdeletion": "true",
```
Kidswiss, line 65: Having the name of this label as a const in the `runtime` package would be nice.

Kidswiss (approved):
LGTM

apis/kubernetes/v1alpha1/doc.go
TheBigLee: we are not. we are just adding it temporarily as SPKS still uses the old provider-kubernetes

pkg/comp-functions/functions/vshnmariadb/user_management.go
TheBigLee: Good point. I've added a new const and replaced all labels

zugao (approved):
We need to test it in dev cluster before pushing to prod

PR #247, opened 2024-10-22
Add initcontainer to wait for PostgreSQL

Kidswiss:
## Summary
It was possible that Nextcloud started without PostgreSQL being ready. This new initcontainer will make sure that PostgreSQL is reachable before attempting to start the main Nextcloud container.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM, it's good that we prevent from deploying nextcloud before postgres is available.

PR #249, opened 2024-10-25
Add MariaDB Scaling

Kidswiss:
## Summary

With this MariaDB can scale to a 3 node Galera cluster.
    
* Added ProxySQl statefulset
* Updated default version of MariaDB
* Updated Galera helm chart
* Added sensible PDBs to ProxySQL
## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_mariadb.go
+++ b/apis/vshn/v1/dbaas_vshn_mariadb.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
+
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -23,7 +24,7 @@ type VSHNMariaDB struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	// Spec defines the desired state of a VSHNMariaDB.
-	Spec VSHNMariaDBSpec `json:"spec"`
+	Spec VSHNMariaDBSpec `json:"spec,omitempty"`
 
 	// Status reflects the observed state of a VSHNMariaDB.
 	Status VSHNMariaDBStatus `json:"status,omitempty"`
@@ -78,15 +79,25 @@ type VSHNMariaDBParameters struct {
 
 	// Security defines the security of a service
 	Security Security `json:"security,omitempty"`
+
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Enum=1;3;
+
+	// Instances configures the number of MariaDB instances for the cluster.
+	// Each instance contains one MariaDB server.
+	// These serves will form a Galera cluster.
+	// An additional ProxySQL statefulset will be deployed to make failovers
+	// as seamless as possible.
+	Instances int `json:"instances,omitempty"`
 }
 
 // VSHNMariaDBServiceSpec contains MariaDB DBaaS specific properties
 type VSHNMariaDBServiceSpec struct {
-	// +kubebuilder:validation:Enum="10.4";"10.5";"10.6";"10.9";"10.10";"10.11";"11.0";"11.1";"11.2";
-	// +kubebuilder:default="11.2"
+	// +kubebuilder:validation:Enum="10.4";"10.5";"10.6";"10.9";"10.10";"10.11";"11.0";"11.1";"11.2";"11.3";"11.4";"11.5";
+	// +kubebuilder:default="11.5"
 
 	// Version contains supported version of MariaDB.
-	// Multiple versions are supported. The latest version "11.2" is the default version.
+	// Multiple versions are supported. The latest version "11.5" is the default version.
 	Version string `json:"version,omitempty"`
 
 	// MariadbSettings contains additional MariaDB settings.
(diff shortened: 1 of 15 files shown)


apis/vshn/v1/dbaas_vshn_mariadb.go
```diff
@@ -78,12 +83,22 @@ type VSHNMariaDBParameters struct {
...
 
 // VSHNMariaDBServiceSpec contains MariaDB DBaaS specific properties
 type VSHNMariaDBServiceSpec struct {
-	// +kubebuilder:validation:Enum="10.4";"10.5";"10.6";"10.9";"10.10";"10.11";"11.0";"11.1";"11.2";
-	// +kubebuilder:default="11.2"
+	// +kubebuilder:validation:Enum="10.4";"10.5";"10.6";"10.9";"10.10";"10.11";"11.0";"11.1";"11.2";"11.3";"11.4";"11.5";
+	// +kubebuilder:default="11.5"
 
 	// Version contains supported version of MariaDB.
 	// Multiple versions are supported. The latest version "11.2" is the default version.
```
zugao, line 104:
```suggestion
	// Multiple versions are supported. The latest version "11.5" is the default version.
```

pkg/comp-functions/functions/vshnmariadb/mariadb_deploy.go
```diff
@@ -396,3 +411,64 @@ func getReleaseSecretValue(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNMariaDB
 	}
 	return m["release"]
 }
+
+// To make scaling as seamless as possible we create an additional service.
+// This service will either point to a single db instance, or to the proxysql cluster.
+func createMainService(comp *vshnv1.VSHNMariaDB, svc *runtime.ServiceRuntime, secretName string) error {
```
zugao, line 417: Technically speaking the customer can also use the svc directly from galera cluster bypassing proxysql, is that right?

zugao (approved):
LGTM

The ProxySQL settings will be managed in a different PR right?

Kidswiss:
> LGTM
> 
> The ProxySQL settings will be managed in a different PR right?

Do you mean the POC with the mysql client? Because right now all the settings are managed, just via the config file.

zugao:
> Do you mean the POC with the mysql client? Because right now all the settings are managed, just via the config file.

Ah yes, we agreed to first start with config file, all good.

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -482,6 +482,14 @@ func KubeOptionAddRefs(refs ...xkube.Reference) KubeObjectOption {
...
 }
 
+// KubeOptionAllowDeletion adds the allow deletion label to the object
+func KubeOptionAllowDeletion(obj *xkube.Object) {
+	if obj.ObjectMeta.Labels == nil {
+		obj.ObjectMeta.Labels = map[string]string{}
+	}
+	obj.ObjectMeta.Labels[WebhookAllowDeletionLabel] = "true"
+}
+
```
wejdross, line 492: why do we have this here? I can't find references in code

wejdross (approved):

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: It's used multiple times in `proxysql.go`:
https://github.com/vshn/appcat/pull/249/files#diff-6c00f19154044c901c94120e2e51fa40f7ccb6e422335f9172257d054447d96dR198
https://github.com/vshn/appcat/pull/249/files#diff-6c00f19154044c901c94120e2e51fa40f7ccb6e422335f9172257d054447d96dR258
https://github.com/vshn/appcat/pull/249/files#diff-6c00f19154044c901c94120e2e51fa40f7ccb6e422335f9172257d054447d96dR414
https://github.com/vshn/appcat/pull/249/files#diff-6c00f19154044c901c94120e2e51fa40f7ccb6e422335f9172257d054447d96dR476
https://github.com/vshn/appcat/pull/249/files#diff-6c00f19154044c901c94120e2e51fa40f7ccb6e422335f9172257d054447d96dR501

wejdross: file was hidden :see_no_evil: now it makes more sense :D

pkg/comp-functions/functions/vshnmariadb/mariadb_deploy.go
Kidswiss: Yes, netpols only apply to pods and not the svcs... so there's no easy way to block this.
https://kubernetes.io/docs/concepts/services-networking/network-policies/#what-you-can-t-do-with-network-policies-at-least-not-yet

But on the other hand, it's probably only a question of time until we find the first app that doesn't like proxysql, just like with pgbouncer, so having a direct option for that might be a good thing.

pkg/comp-functions/functions/vshnmariadb/proxysql.go
```diff
@@ -0,0 +1,510 @@
...
+        {writer_hostgroup=2,backup_writer_hostgroup=4,reader_hostgroup=3,offline_hostgroup=1,active=1,max_writers=1,writer_is_also_reader=1,max_transactions_behind=100}
+    )
+
+    proxysql_servers =
+    (
+        { hostname = "proxysql-0.proxysqlcluster.{{ .Namespace }}.svc", port = 6032, weight = 1 },
+        { hostname = "proxysql-1.proxysqlcluster.{{ .Namespace }}.svc", port = 6032, weight = 1 }
+    )`
+)
+
```
wejdross, line 69: can't we just move it to separate file and then template it with text/template? 
#readibility

wejdross (approved):

pkg/comp-functions/functions/vshnmariadb/proxysql.go
Kidswiss: Good idea!

zugao (approved):
LGTM

PR #250, opened 2024-10-28
Remove custom pg exporter queries

zugao:
## Summary

* After upgrading stackgres operator to the latest version some of our custom pg metrics we export does not work anymore.. We have found out that the pg_exporter has been updated in stackgres as we don't require any of the metrics I removed in this PR

## Checklist

- [x] Update tests.


wejdross (approved):

Kidswiss (approved):
Please test before rolling out

zugao:
I tested with @wejdross and it seems the default config map was successfully recreated gfor each instances without any restarts. During maintenance the instances will pick up the new file.

PR #256, opened 2024-10-30
Implement webhook logic for split control-plane - service cluster setup

TheBigLee:
## Summary

* This PR adds support for connecting the webhook controller to another cluster.

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/objectbucket.go
+++ b/pkg/comp-functions/functions/vshnpostgres/objectbucket.go
@@ -25,6 +25,9 @@ func EnsureObjectBucketLabels(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, 
 	if err != nil {
 		return runtime.NewWarningResult("cannot get xobjectbucket")
 	}
+	labels := bucket.GetLabels()
+	labels["appcat.vshn.io/ignore-provider-config"] = "true"
+	bucket.SetLabels(labels)
 
 	err = svc.SetDesiredComposedResourceWithName(bucket, "pg-bucket")
 	if err != nil {
--- a/pkg/comp-functions/functions/vshnpostgres/objectbucket.go
+++ b/pkg/comp-functions/functions/vshnpostgres/objectbucket.go
@@ -25,9 +25,12 @@ func EnsureObjectBucketLabels(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, 
 	if err != nil {
 		return runtime.NewWarningResult("cannot get xobjectbucket")
 	}
-	labels := bucket.GetLabels()
-	labels["appcat.vshn.io/ignore-provider-config"] = "true"
-	bucket.SetLabels(labels)
+
+	labels := map[string]string{
+		"appcat.vshn.io/ignore-provider-config": "true",
+	}
+
+	svc.AddLabels(bucket, labels)
 
 	err = svc.SetDesiredComposedResourceWithName(bucket, "pg-bucket")
 	if err != nil {
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -446,7 +446,7 @@ func ComposedOptionProtects(resName string) ComposedResourceOption {
 // adds it to the desired composed resources. It takes options to manipulate the resulting kubec object before applying.
 func (s *ServiceRuntime) SetDesiredKubeObject(obj client.Object, objectName string, opts ...KubeObjectOption) error {
 
-	kobj, err := s.putIntoObject(false, obj, objectName, objectName)
+	kobj, err := s.putIntoObject(obj, objectName, objectName)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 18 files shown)


pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -560,6 +560,18 @@ func (s *ServiceRuntime) SetDesiredKubeObserveObject(obj client.Object, objectNa
...
+func (s *ServiceRuntime) SetDesiredKubeObserveObjectWithLabels(obj client.Object, objectName string, labels map[string]string, refs ...xkube.Reference) error {
+
+	kobj, err := s.putIntoObjectWithLabels(true, obj, objectName, objectName, labels, refs...)
+	if err != nil {
+		return err
+	}
+
+	return s.SetDesiredComposedResourceWithName(kobj, objectName)
+}
+
```
wejdross, line 574: I see here good pattern for builder.

SetDesiredKubeObject().WithName().WithLabels().WithReferences()

does anyone else agree? Should we open a ticket to refactor that piece of code?

pkg/controller/webhooks/deletionprotection_test.go
```diff
@@ -127,7 +127,7 @@ func Test_checkManagedObject(t *testing.T) {
 	assert.NoError(t, err)
 
 	// Then expect parent
-	compInfo, err = checkUnmanagedObject(context.TODO(), pvc, c, logr.Discard())
+	compInfo, err = checkUnmanagedObject(context.TODO(), pvc, c, c, logr.Discard())
```
wejdross, line 130: why do You pass the same client `c` twice?

pkg/comp-functions/runtime/function_mgr.go
Kidswiss: I'd rather go with functional options like we already have with `SetDesiredKubeObject`. IMHO makes it easier to use than having a builder.

pkg/controller/webhooks/deletionprotection_test.go
TheBigLee: Because for this test, we don't use a control-cluster <-> service-cluster split. So both clients are the same.

pkg/comp-functions/functions/vshnpostgres/objectbucket.go
```diff
@@ -19,12 +19,17 @@ func EnsureObjectBucketLabels(ctx context.Context, comp *vshnv1.VSHNPostgreSQL,
...
+		"appcat.vshn.io/ignore-provider-config": "true",
+	}
+
 	bucket := &appcatv1.XObjectBucket{}
 
 	err = svc.GetDesiredComposedResourceByName(bucket, "pg-bucket")
 	if err != nil {
 		return runtime.NewWarningResult("cannot get xobjectbucket")
 	}
+	bucket.SetLabels(labels)
```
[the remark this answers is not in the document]
TheBigLee: Fixed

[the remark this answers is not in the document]
zugao: I would add a function in runtime to manage this.

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -475,6 +475,26 @@ func (s *ServiceRuntime) SetDesiredKubeObjectWithName(obj client.Object, objectN
...
+		}
+		for val, key := range labels {
+			current[val] = key
+		}
+		obj.SetLabels(current)
+	}
+}
+
+// KubeOptionObserveOnly sets the object to only observe.
+func KubeOptionObserveOnly(obj *xkube.Object) {
```
zugao, line 493: There's already a function that does just that `KubeOptionObserve(obj *xkube.Object)`

zugao (changes requested):
Looks good but some small changes are needed

pkg/controller/webhooks/deletionprotection.go
```diff
@@ -145,3 +151,40 @@ func isDeletionProtected(obj client.Object) bool {
...
+		if err != nil {
+			return cpClient, err
+		}
+
+		// Read the file into a byte slice
+		kubeconfig := make([]byte, stat.Size())
+		_, err = bufio.NewReader(file).Read(kubeconfig)
+		if err != nil && err != io.EOF {
+			return cpClient, err
+		}
```
wejdross, line 177: even though I like C way of reading a file, I'd go with os.ReadFile() :)

wejdross (commented):
small things

pkg/comp-functions/runtime/function_mgr.go
TheBigLee: I've removed mine

pkg/controller/webhooks/deletionprotection.go
TheBigLee: Good point. Changed the code to use `os.ReadFile()`

wejdross (approved):
my suggestions are applied, LGTM

zugao (approved):
LGTM

PR #259, opened 2024-11-07
Fix typo in the usage protecting the instance NS

Kidswiss:
## Summary

Crossplane resource names are always lower case. If the name doesn't match exactly, then the usage won't be applied properly.

This lead to a race condition when deprovisioning MariaDB with users.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM

PR #261, opened 2024-11-15
Prevent empty BackupNames from being referenced

Kidswiss:
## Summary

After an instance has been restored and some of the fields get cleaned the cluster kube object can still have a non-nil `initialData` field containing empty values.

Previously we only checked if the whole field is nil, but we also need to check if the BackupName is empty or not.

## Checklist

- [x] Update tests.


wejdross (approved):

zugao (approved):
LGTM

PR #262, opened 2024-11-15
Re-write alerts so they work with cluster monitoring

Kidswiss:
## Summary

The current user-workload alerts will not properly work with the cluster workload monitoring, because:

* A specific `"syn": "true"` label is missing
* They don't filter for the namespace. So each alert would trigger for each instance that is running.

The second part needed a bit more involved refactoring, as we have to pass the specific namespace for each instance and can't just match in a generic manner.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/nonsla/alerting.go
+++ b/pkg/comp-functions/functions/common/nonsla/alerting.go
@@ -5,7 +5,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// ServiceRule is a func definition to get a specific rule based on a container name s
+// ServiceRule is a func definition to get a specific rule based on a container name s and namespace name n
 type ServiceRule func(s, n string) promV1.Rule
 
 // alert non-exportable alert type to be used only in this package
@@ -18,7 +18,10 @@ type Alerts struct {
 	// alerts are generic alerts defined for all services of appcat
 	alerts []alert
 	// alertContainerName is the container name to be used for alert names alert expression
-	alertContainerName, namespace string
+	alertContainerName string
+
+	// Alertdefinitions contains all the ServiceRules
+	alertDefinitions map[alert]ServiceRule
 }
 
 const (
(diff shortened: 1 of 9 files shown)


pkg/comp-functions/functions/common/nonsla/alerting.go
```diff
@@ -5,7 +5,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// ServiceRule is a func definition to get a specific rule based on a container name s
+// ServiceRule is a func definition to get a specific rule based on a container name s and namespace name n
```
zugao, line 8:
```suggestion
// ServiceRule is a func definition to get a specific rule based on a container name s and namespace name
```

pkg/comp-functions/functions/common/nonsla/alerting.go
```diff
@@ -60,12 +64,13 @@ var (
 				},
 				Expr: intstr.IntOrString{
 					Type:   intstr.String,
-					StrVal: "label_replace( bottomk(1, (kubelet_volume_stats_available_bytes{job=\"kubelet\",metrics_path=\"/metrics\"} / kubelet_volume_stats_capacity_bytes{job=\"kubelet\",metrics_path=\"/metrics\"}) < 0.15 and kubelet_volume_stats_used_bytes{job=\"kubelet\",metrics_path=\"/metrics\"} > 0 and predict_linear(kubelet_volume_stats_available_bytes{job=\"kubelet\",metrics_path=\"/metrics\"}[6h], 4 * 24 * 3600) < 0  unless on(namespace, persistentvolumeclaim) kube_persistentvolumeclaim_access_mode{access_mode=\"ReadOnlyMany\"} == 1 unless on(namespace,persistentvolumeclaim) kube_persistentvolumeclaim_labels{label_excluded_from_alerts=\"true\"}== 1) * on(namespace) group_left(label_appcat_vshn_io_claim_namespace)kube_namespace_labels, \"name\", \"$1\", \"namespace\",\"vshn-" + namespace + "-(.+)-.+\")",
```
zugao, line 63: was this namespace a bug until now?

zugao (approved):
Looks ok but yeah check what you get with crossplane diff.

pkg/comp-functions/functions/common/nonsla/alerting.go
Kidswiss: no :D

pkg/comp-functions/functions/common/nonsla/alerting.go
Kidswiss: I don't think so, the user monitoring is based on namespacing. So each of these rules would only evaluate for its own namespace anyway.

PR #263, opened 2024-11-15
Allow rotation of the restart ops

Kidswiss:
## Summary

Without the annotation Crossplane won't be able to remove the old ops and get stuck in `Synced: false` state.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM but here crossplane will only remove SGDbOps not pods.

Kidswiss:
> LGTM but here crossplane will only remove SGDbOps not pods.

You're right, but the end result is a pod :)

I adjusted the wording though.

PR #264, opened 2024-11-15
Allow to disable PgBouncer in VSHNPostgreSQL

wejdross:
## Summary

* Adding new parameter allowing users to disable PgBouncer in our VSHNPostgreSQL

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_postgresql.go
+++ b/apis/vshn/v1/dbaas_vshn_postgresql.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
+
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	sgv1 "github.com/vshn/appcat/v4/apis/stackgres/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -146,6 +147,10 @@ type VSHNPostgreSQLServiceSpec struct {
 
 	// PgBouncerSettings passes additional configuration to the pgBouncer instance.
 	PgBouncerSettings *sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni `json:"pgBouncerSettings,omitempty"`
+
+	// Disable connection pooling service PgBouncer. All connections will go straight to PostgreSQL instance.
+	// +kubebuilder:default=false
+	DisablePgBouncer bool `json:"DisablePgBouncer,omitempty"`
 
 	// +kubebuilder:default=true
 	// This is default option if neither repack or vacuum are selected
--- a/crds/vshn.appcat.vshn.io_vshnkeycloaks.yaml
+++ b/crds/vshn.appcat.vshn.io_vshnkeycloaks.yaml
@@ -9775,6 +9775,10 @@ spec:
                             service:
                               description: Service contains PostgreSQL DBaaS specific properties
                               properties:
+                                DisablePgBouncer:
+                                  default: false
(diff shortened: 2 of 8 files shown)


apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -147,6 +148,10 @@ type VSHNPostgreSQLServiceSpec struct {
 	// PgBouncerSettings passes additional configuration to the pgBouncer instance.
 	PgBouncerSettings *sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni `json:"pgBouncerSettings,omitempty"`
 
+	// Disable connection pooling. This means that PgBouncer will be disabled and all connections will be made directly to the PostgreSQL instance.
+	// +kubebuilder:default=false
+	DisableConnectionPooling bool `json:"disableConnectionPooling,omitempty"`
```
zugao, line 153: I think this setup is confusing. Having `PgBouncerSettings` and `DisableConnectionPooling` in the same object and with different names might confuse the end user. The best way in my opinion would be to put those fields in an object names `ConnectionPooling`. This will probably be a pain in the ass as it will break the current API. What we could also do, we could just name them the bool field `DisablePgBouncer`. @Kidswiss what's your take on that?

wejdross: I thought about that, but struct managing pgBouncer Settings is called: `*sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni` which imho should be reserved just for Ini file settings

zugao: I don't understand what your comment has to do with the issue I am raising. Why do we have to care about `*sgv1.SGPoolingConfigSpecPgBouncerPgbouncerIni` internal struct ?

wejdross: there's also 
```

// SGPoolingConfigSpec defines model for SGPoolingConfigSpec.
type SGPoolingConfigSpec struct {
	// Connection pooling configuration based on PgBouncer.
	PgBouncer *SGPoolingConfigSpecPgBouncer `json:"pgBouncer,omitempty"`
}
```
which sounds perfectly, but then it would be possible to configure pgbouncer ini in few places, which is also weird unfortunate

therefore I decided to add additional field

zugao: You can also just create your own struct `ConnectionPooling` with `DisableConnectionPooling` and `PgBouncerSettings ` as fields.

wejdross: but I can't just move `PgBouncerSettings` to another struct as such change would impact all our running instances, I really don't think we should introduce such api changes in simple boolean field. I can add it to the `PgBouncerSettings` - it has nothing to do with pgbouncer ini file, but let's say it's the closest to the subject. Also if someone wants to disable PgBouncer it's rather unlike to set any PgBouncer settings

zugao: Have you carefully read my first comment?

> I can add it to the `PgBouncerSettings`

No, you cannot. You would have the same problem as you described.

Kidswiss: I'd go with just renaming the field to `DisablePgBouncer`. Introducing even more levels in our claims might also get very confusing at one point.

Kidswiss (approved):

PR #265, opened 2024-11-15
Add new custom pg exporter config

Kidswiss:
## Summary

StackGres' default exporter configuration doesn't play well with large databases.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/pgqexporter_config.go
+++ b/pkg/comp-functions/functions/vshnpostgres/pgqexporter_config.go
@@ -1,0 +1,55 @@
+package vshnpostgres
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+
+	"github.com/crossplane/function-sdk-go/proto/v1beta1"
+	xkube "github.com/vshn/appcat/v4/apis/kubernetes/v1alpha2"
+	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+/*
+	This code ensures we have minimalistic postgresql-exporter configuration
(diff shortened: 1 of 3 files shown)


pkg/comp-functions/functions/vshnpostgres/pgqexporter_config.go
```diff
@@ -0,0 +1,55 @@
...
+
+/*
+	This code ensures we have minimalistic postgresql-exporter configuration
+	It prevents high memory usage for big databases
+*/
+//go:embed scripts/queries.yml
+var queries string
+
+func PgExporterConfig(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) *v1beta1.Result {
+	err := svc.GetObservedComposite(comp)
```
zugao, line 24: Do you disable somewhere the current exporter?

pkg/comp-functions/functions/vshnpostgres/scripts/queries.yml
```diff
@@ -0,0 +1,880 @@
+pg_replication:
```
zugao, line 1: Where is this coming from? How can we ensure this is updated in time?

pkg/comp-functions/functions/vshnpostgres/pgqexporter_config.go
Kidswiss: We don't disable anything, we just override the default configs from stackgres.

pkg/comp-functions/functions/vshnpostgres/scripts/queries.yml
Kidswiss: This is from the original stackgres exporter config.

I did a `curl localhost:9187/metrics | grep -o ".*{" | uniq -c | sort -bgr` on the container to identify all the really big metrics and just removed them from the configs.

As for updating them... We could probably observe the original confing in the compfunction and remove the broken queries then dynamically 🤔

zugao (approved):
LGTM

PR #267, opened 2024-11-19
Use a default client if controlPanel client is nil

Kidswiss:
## Summary

XObjectBuckets and mysql provder-sql objects did not have a valid cpClient specified. This causes nilpointers if any of these should get deleted.

Resulting in errors like these:
```
Error from server (InternalError): Internal error occurred: failed calling webhook "users.mysql.vshn.appcat.vshn.io": failed to call webhook: Post "https://webhook-service.syn-appcat.svc:443/validate-mysql-sql-crossplane-io-v1alpha1-user?timeout=10s": EOF
```

These are Crossplane managed resources and always reside on the control plane.

So they will always have the same clients for local and control plane connections.

If the cpClient isn't specified we default to the other client. This change will make it easier in the future to add managed resources as we don't have to specify both clients seperately (and potentially forget them...).

## Checklist

- [x] Update tests.


zugao (approved):
LGTM

PR #269, opened 2024-11-21
Use an array instead of a string for the FQDN parameter

TheBigLee:
## Summary

* This PR will change the FQD parameter from a string to an array, so that the end user can specify multiple endpoints.

## Checklist

- [x] Update tests.

### DIFF
--- a/apis/vshn/v1/vshn_nextcloud.go
+++ b/apis/vshn/v1/vshn_nextcloud.go
@@ -93,7 +93,7 @@ type VSHNNextcloudServiceSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
 
-	// FQDN contains the FQDNs which will be used for the ingress.
+	// FQDN contains the FQDNs array, which will be used for the ingress.
 	// If it's not set, no ingress will be deployed.
 	// This also enables strict hostname checking for this FQDN.
 	FQDN []string `json:"fqdn"`
--- a/apis/vshn/v1/vshn_nextcloud.go
+++ b/apis/vshn/v1/vshn_nextcloud.go
@@ -91,11 +91,12 @@ type VSHNNextcloudServiceSpec struct {
 	// Collabora contains settings to control the Collabora integration.
 	Collabora CollaboraSpec `json:"collabora,omitempty"`
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
 
-	// FQDN contains the FQDN which will be used for the ingress.
+	// FQDN contains the FQDNs which will be used for the ingress.
 	// If it's not set, no ingress will be deployed.
 	// This also enables strict hostname checking for this FQDN.
-	FQDN string `json:"fqdn"`
+	FQDN []string `json:"fqdn"`
 
 	// RelativePath on which Nextcloud will listen.
 	// +kubebuilder:default="/"
--- a/pkg/comp-functions/functions/vshnnextcloud/collabora_test.go
+++ b/pkg/comp-functions/functions/vshnnextcloud/collabora_test.go
@@ -56,11 +56,4 @@ func Test_addCollabora(t *testing.T) {
 	res = DeployCollabora(ctx, comp, svc)
 
 	assert.True(t, res.Severity == v1beta1.Severity_SEVERITY_NORMAL && res.Message == "Collabora not enabled")
-
-	comp.Spec.Parameters.Service.Collabora.Enabled = true
-	comp.Spec.Parameters.Service.Collabora.FQDN = ""
-	res = DeployCollabora(ctx, comp, svc)
-
-	assert.True(t, res.Severity == v1beta1.Severity_SEVERITY_WARNING && res.Message == "Collabora FQDN is not a valid DNS name: "+comp.Spec.Parameters.Service.Collabora.FQDN)
-
 }
--- a/pkg/comp-functions/functions/vshnnextcloud/deploy.go
+++ b/pkg/comp-functions/functions/vshnnextcloud/deploy.go
@@ -257,14 +257,6 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 		}
 	}
 
-	if len(comp.Spec.Parameters.Service.FQDN) == 0 {
-		return nil, fmt.Errorf("FQDN array is empty, but requires at least one entry: %w", errors.New("empty fqdn"))
-	}
-
-	if err := validateFQDNs(comp.Spec.Parameters.Service.FQDN); err != nil {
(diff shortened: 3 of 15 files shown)


apis/vshn/v1/vshn_nextcloud.go
```diff
@@ -91,11 +91,12 @@ type VSHNNextcloudServiceSpec struct {
...
 	Collabora CollaboraSpec `json:"collabora,omitempty"`
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
 
-	// FQDN contains the FQDN which will be used for the ingress.
+	// FQDN contains the FQDNs which will be used for the ingress.
 	// If it's not set, no ingress will be deployed.
 	// This also enables strict hostname checking for this FQDN.
-	FQDN string `json:"fqdn"`
+	FQDN []string `json:"fqdn"`
```
wejdross, line 99: remember to inform customer before deployment :P

apis/vshn/v1/vshn_nextcloud.go
```diff
@@ -91,11 +91,12 @@ type VSHNNextcloudServiceSpec struct {
 	// Collabora contains settings to control the Collabora integration.
 	Collabora CollaboraSpec `json:"collabora,omitempty"`
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
 
-	// FQDN contains the FQDN which will be used for the ingress.
+	// FQDN contains the FQDNs which will be used for the ingress.
```
wejdross, line 96:
```suggestion
	// FQDN contains the FQDNs array, which will be used for the ingress.
```

`make generate` afterwards

pkg/comp-functions/functions/vshnnextcloud/deploy.go
```diff
@@ -256,10 +256,20 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
...
 	}
 
+	if len(comp.Spec.Parameters.Service.FQDN) == 0 {
+		return nil, fmt.Errorf("FQDN array is empty, but requires at least one entry: %w", errors.New("empty fqdn"))
+	}
+
+	trustedDomain := []string{
+		comp.GetName() + "." + comp.GetInstanceNamespace() + ".svc.cluster.local",
+	}
+	trustedDomain = append(trustedDomain, comp.Spec.Parameters.Service.FQDN...)
```
wejdross, line 266: I'd make here verification, similar to: https://github.com/vshn/appcat/blob/master/pkg/comp-functions/functions/vshnnextcloud/collabora.go#L56

or even now I'm thinking that we could move this input verification to webhook, so we won't receive ticket to debug broken fqdn

wejdross (changes requested):

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: I've added the validation

wejdross (approved):

wejdross:
please fix Summary section before merging

pkg/comp-functions/functions/vshnnextcloud/deploy.go
```diff
@@ -428,3 +443,12 @@ func setBackgroundJobMaintenance(t vshnv1.TimeOfDay, nextcloudConfig string) str
 	backgroundJobHour := t.GetTime().Add(40 * time.Minute).Add(time.Hour).Hour()
 	return strings.Replace(nextcloudConfig, "%maintenance_value%", strconv.Itoa(backgroundJobHour), 1)
 }
+
+func validateFQDNs(fqdns []string) error {
+	for _, fqdn := range fqdns {
+		if !valid.IsDNSName(fqdn) {
```
Kidswiss, line 449: Idea: this is something we could check in a webhook, so it doesn't break silently for the user.

Kidswiss (approved):
Just one idea, rest LGTM

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: I've refactored the code and moved this validation to the webhook

pkg/controller/webhooks/nextcloud.go
```diff
@@ -41,3 +47,36 @@ func SetupNextcloudWebhookHandlerWithManager(mgr ctrl.Manager, withQuota bool) e
 		}).
 		Complete()
 }
+
+func (n *NextcloudWebhookHandler) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
Kidswiss, line 51: We should also validate on update, not only on create.

Kidswiss (changes requested):

wejdross (approved):

pkg/controller/webhooks/nextcloud.go
TheBigLee: Good point. Added that part as well

pkg/comp-functions/functions/vshnnextcloud/collabora_test.go
```diff
@@ -56,11 +56,4 @@ func Test_addCollabora(t *testing.T) {
 	res = DeployCollabora(ctx, comp, svc)
 
 	assert.True(t, res.Severity == v1beta1.Severity_SEVERITY_NORMAL && res.Message == "Collabora not enabled")
-
```
zugao, line 59: well I would add a test here check on the list of FQDN.

zugao: Why this had to be removed? don't you work with nextcloud fqdn?

pkg/comp-functions/functions/vshnnextcloud/deploy.go
```diff
@@ -256,10 +257,20 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 		}
 	}
 
+	if len(comp.Spec.Parameters.Service.FQDN) == 0 {
```
zugao, line 260: Can we add that into directly in the the webhook?

TheBigLee: Done

Kidswiss: > Can we add that into directly in the the webhook?

Already covered here: https://github.com/vshn/appcat/pull/269/files#diff-d41d9990ae83874d678984536634bbcc08146417526791f235c09d1bdc9f0d7eR94

No need for a webhook here...

Kidswiss (approved):

pkg/comp-functions/functions/vshnnextcloud/deploy.go
TheBigLee: True, I've removed the whole section

PR #272, opened 2024-11-26
Don't deploy billing prometheus rule for tests

Kidswiss:
## Summary

This will not deploy any billing data for instances where the claim resides in the `vshn-test` namespace.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/billing.go
+++ b/pkg/comp-functions/functions/common/billing.go
@@ -5,12 +5,13 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"text/template"
+
 	xfnproto "github.com/crossplane/function-sdk-go/proto/v1beta1"
 	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
-	"text/template"
 )
 
 var rawExpr = "vector({{.}})"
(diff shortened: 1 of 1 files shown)


pkg/comp-functions/functions/common/billing.go
```diff
@@ -22,6 +23,11 @@ func CreateBillingRecord(ctx context.Context, svc *runtime.ServiceRuntime, comp
 	log := controllerruntime.LoggerFrom(ctx)
 	log.Info("Enabling billing for service", "service", comp.GetName())
 
+	if comp.GetClaimNamespace() == "vshn-test" {
```
wejdross, line 26: do we want to parametrize it via component, so it's not that visible to the whole world?

Kidswiss: That's a better argument :D

However then we have yet another parameter in the already very bloated component.

zugao: I would still move it there, we will know that it exists instead of forgetting about it. There's no escape from bloating the component. And this is not the solution to un-bloat it. I would really encourage to create a parameter there.

Kidswiss: I've made it configurable now.

But making a value that could be a constant value configurable just so we don't forget about it also seems wrong. It's not like we remember all the stuff that's in the component as well.

zugao: I would go to the component to search for smth first instead of the go code.

zugao (approved):

pkg/comp-functions/functions/common/billing.go
Kidswiss: The right place for that would be the docs, but noone ever reads them, so I also stopped documenting...

zugao: that's also true

wejdross: my idea was to set a global commodore parameter, this way we would keep namespace name for ourselfs

PR #273, opened 2024-11-27
Add support for remote control plane to SLI exporter

Kidswiss:
## Summary

This adds support for remote control planes for the SLI exporter. It now
takes an additional `KUBECONFIG` env var to connect to the control plane.

Additionally to determine the state of the maintenance of the service
cluster, it connects to the local service cluster to reconcile on
`UpgradeJob` objects.

Also the SLI-Prober doesn't need any configuration about the services
anymore. It will detect these on startup.

## Checklist

- [x] Update tests.

### DIFF
--- a/cmd/sliexporter.go
+++ b/cmd/sliexporter.go
@@ -2,12 +2,14 @@ package cmd
 
 import (
 	"os"
-	"strconv"
 	"time"
 
+	managedupgradev1beta1 "github.com/appuio/openshift-upgrade-controller/api/v1beta1"
 	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
+	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
 	"github.com/vshn/appcat/v4/pkg"
+	"github.com/vshn/appcat/v4/pkg/common/utils"
 	maintenancecontroller "github.com/vshn/appcat/v4/pkg/sliexporter/maintenance_controller"
 	"github.com/vshn/appcat/v4/pkg/sliexporter/probes"
 	vshnkeycloakcontroller "github.com/vshn/appcat/v4/pkg/sliexporter/vshnkeycloak_controller"
@@ -16,16 +18,20 @@ import (
 	vshnpostgresqlcontroller "github.com/vshn/appcat/v4/pkg/sliexporter/vshnpostgresql_controller"
 	vshnrediscontroller "github.com/vshn/appcat/v4/pkg/sliexporter/vshnredis_controller"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 )
 
 type sliProber struct {
-	scheme                                                                                                                      *runtime.Scheme
-	metricsAddr, probeAddr                                                                                                      string
-	leaderElect, enableVSHNPostgreSQL, enableVSHNRedis, enableVSHNMinio, enableMaintenanceStatus, enableKeycloak, enableMariaDB bool
+	scheme                                    *runtime.Scheme
+	metricsAddr, probeAddr, serviceKubeConfig string
+	leaderElect                               bool
 }
 
 var s = sliProber{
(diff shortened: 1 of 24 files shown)


pkg/sliexporter/maintenance_controller/controller.go
```diff
@@ -90,8 +93,16 @@ func (r *MaintenanceReconciler) jobState(job managedupgradev1beta1.UpgradeJob) s
...
+
+	// For external reconcile triggers we can't register a kind via `For()`
+	// instead we have to name the reconciler something
+	// usually the name is the lowercase of the kind
+	// so that's what I've used here.
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&managedupgradev1beta1.UpgradeJob{}).
+		Named("upgradejob").
+		// This is the magic sauce, it makes the reconciler reconcile on events happening on the serviceCluster
+		WatchesRawSource(source.Kind(sc.GetCache(), &managedupgradev1beta1.UpgradeJob{}, &handler.TypedEnqueueRequestForObject[*managedupgradev1beta1.UpgradeJob]{})).
```
wejdross, line 106:
```suggestion
		WatchesRawSource(source.Kind(sc.GetCache(), &managedupgradev1beta1.UpgradeJob{}, &handler.TypedEnqueueRequestForObject[*managedupgradev1beta1.UpgradeJob]{})).
		
			return ctrl.NewControllerManagedBy(mgr).
		Named("upgradejob").
		// This is the magic sauce, it makes the reconciler reconcile on events happening on the serviceCluster
		WatchesRawSource(source.Kind(
			sc.GetCache(),
			&managedupgradev1beta1.UpgradeJob{},
			&handler.TypedEnqueueRequestForObject[*managedupgradev1beta1.UpgradeJob]{},
		)).
		Complete(r)
```

wejdross (approved):

apis/vshn/v1/vshn_nextcloud.go
```diff
@@ -188,6 +188,19 @@ type CollaboraSpec struct {
 	FQDN string `json:"fqdn,omitempty"`
 }
 
+type addon interface {
```
zugao, line 191: You forgot to cleanup after our short session about addons.

cmd/sliexporter.go
```diff
@@ -39,21 +45,17 @@ var SLIProberCMD = &cobra.Command{
...
-	SLIProberCMD.Flags().BoolVar(&s.enableVSHNRedis, "vshn-redis", getEnvBool("APPCAT_SLI_VSHNREDIS"),
-		"Enable probing of VSHNRedis instances")
-	SLIProberCMD.Flags().BoolVar(&s.enableVSHNMinio, "vshn-minio", getEnvBool("APPCAT_SLI_VSHNMINIO"),
-		"Enable probing of VSHNMinio instances")
-	SLIProberCMD.Flags().BoolVar(&s.enableMaintenanceStatus, "vshn-track-oc-maintenance-status", getEnvBool("APPCAT_SLI_TRACK_OC_MAINTENANCE_STATUS"),
-		"Enable oc maintenance status observer. Will set the labels 'maintenance' accordingly.")
-	SLIProberCMD.Flags().BoolVar(&s.enableKeycloak, "vshn-keycloak", getEnvBool("APPCAT_SLI_VSHNKEYCLOAK"), "Enable probing of VSHNKeycloak instances")
-	SLIProberCMD.Flags().BoolVar(&s.enableMariaDB, "vshn-mariadb", getEnvBool("APPCAT_SLI_VSHNMARIADB"), "Enable probing of VSHNMariaDB instances")
+	SLIProberCMD.Flags().StringVar(&s.serviceKubeConfig, "service-kubeconfig", os.Getenv("SERVICE_KUBECONFIG"),
+		"Kubeconfig for the service cluster itself, usually only for debugging purpose. The sliproper should run on the service clusters and thus use the in-cluster-config.")
```
zugao, line 58:
```suggestion
		"Kubeconfig for the service cluster itself, usually only for debugging purpose. The sliprober should run on the service clusters and thus use the in-cluster-config.")
```

pkg/sliexporter/sli_reconciler/reconciler.go
```diff
@@ -31,7 +31,8 @@ type Reconciler struct {
 	nn                 types.NamespacedName
 	startupGracePeriod time.Duration
 	fetchProberFor     func(context.Context, Service) (probes.Prober, error)
-	client             client.Client
+	// TODO: multiple clients
```
zugao, line 34: TODO?

zugao (commented):
Some small nitpicks but I don't see where you connect to the control plane, where do you set `KUBECONFIG`? I see only `SERVICE_ KUBECONFIG` which is not even mandatory

apis/vshn/v1/vshn_nextcloud.go
Kidswiss: ah, shit :D

pkg/sliexporter/sli_reconciler/reconciler.go
Kidswiss: We can crape that, I ended up not using multiple clients at all.

Kidswiss:
> Some small nitpicks but I don't see where you connect to the control plane, where do you set `KUBECONFIG`? I see only `SERVICE_ KUBECONFIG` which is not even mandatory

`KUBECONFIG` is read in the `ctrl.GetConfigOrDie()` function. As seen here: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.3/pkg/client/config#GetConfigOrDie

zugao:
As pointed out in the docs PR, does this PR also addresses the problem of XRs from different service clusters? We should make sure this works, otherwise it's gonna be a bug which will have to fix later on.

Kidswiss:
> As pointed out in the docs PR, does this PR also addresses the problem of XRs from different service clusters? We should make sure this works, otherwise it's gonna be a bug which will have to fix later on.

Yes I need to add that logic, somehow I missed that completely. Might end up having multiple k8s clients after all.

Kidswiss:
> As pointed out in the docs PR, does this PR also addresses the problem of XRs from different service clusters? We should make sure this works, otherwise it's gonna be a bug which will have to fix later on.

I came up with a simple logic to match all control plane composites to a local service cluster, could you have another look?

pkg/sliexporter/sli_reconciler/reconciler.go
```diff
@@ -105,3 +121,25 @@ func (r *Reconciler) Reconcile(ctx context.Context) (ctrl.Result, error) {
...
+	if !ok {
+		return false, fmt.Errorf("resource does not implement common.Composite")
+	}
+
+	ns := &corev1.Namespace{}
+
+	err := r.scClient.Get(ctx, types.NamespacedName{Name: comp.GetInstanceNamespace()}, ns)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.l.Info("instance doesn't have a instancenamespace on this cluster, skipping")
```
zugao, line 138:
```suggestion
			r.l.Info("instance doesn't have a instance namespace on this cluster, skipping")
```

zugao (approved):
LGTM

PR #274, opened 2024-11-28
Set the correct version for the SGCluster

TheBigLee:
## Summary

We need to ensure that the SGCluster always contains the "correct" version and that we don't override it with just a major version. The reason is, that Stackgres doesn't allow changing the version number in the `SGCluster` object. Since stackgres changes the version field during a minor upgrade, it won't be possible anymore to "revert" this change to a major version only setting.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/maintenance.go
+++ b/pkg/comp-functions/functions/vshnpostgres/maintenance.go
@@ -14,6 +14,7 @@ import (
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 )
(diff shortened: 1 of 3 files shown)


pkg/comp-functions/functions/vshnpostgres/maintenance.go
```diff
@@ -178,3 +184,21 @@ func createMaintenanceSecret(instanceNamespace, sgNamespace, resourceName string
...
+	err := svc.GetObservedKubeObject(observedCluster, "cluster")
+	if errors.IsNotFound(err) {
+		// Cluster doesn't exist yet. So let's ignore it here
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("cannot get observed cluster object: %w", err)
+	}
+
+	pgVersion := observedCluster.Spec.Postgres.Version
```
zugao, line 199: shouldn't you take from the status?

pkg/comp-functions/functions/vshnpostgres/maintenance.go
```diff
@@ -129,6 +130,11 @@ func addSchedules(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runtime
 		},
 	}...)
 
+	err = setPsqlMinorVersion(svc, cluster)
```
zugao, line 133: what if you set the version to null? Will the operator figure it out itself?

pkg/comp-functions/functions/vshnpostgres/maintenance.go
Kidswiss: The problem comes from trying to change this field, so it makes sense that we take this value.

pkg/comp-functions/functions/vshnpostgres/maintenance.go
Kidswiss: Right now the version field is not required, so not setting it might be a way to go.

However, if they ever decide to make it required, it would break again. Then we'd have to go back to this solution.

Kidswiss (approved):

zugao (approved):

PR #275, opened 2024-11-28
Add possibility to disable TLS in VSHNPostgreSQL

wejdross:
## Summary

* Short and easy implementation of TLS switch in VSHNPostgreSQL
* compatibile with SLI exporter
* switch happens without any pod distruption and takes place immediately


apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -226,6 +229,13 @@ type VSHNPostgreSQLEncryption struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+// VSHNPostgreSQLTLS contains TLS specific parameters
+type VSHNPostgreSQLTLS struct {
+	// Enabled specifies if the instance should use TLS for the instance.
```
Kidswiss, line 234: You could also add here that there will be no interruption in the service, if this value changes.

Kidswiss (approved):
Just one little nitpick, rest LGTM

zugao (approved):
LGTM

PR #277, opened 2024-11-29
adding missing reference in vshnpostgresql claim

wejdross:
## Summary

*Without this setting some applications might refuse to work with "Server does not support TLS connections"
*caused by .spec.parameters.service.tls = nil


Kidswiss (approved):
Yeah now it should be correct.

PR #278, opened 2024-12-02
add TLS fix

wejdross:
## Summary

* Fixing TLS settings
* For context: there is a difference if we initialize nested map inside some struct and single field. I had to add `properties` entry to generator to make it work.
* Currently our XVSHNPostgreSQL receives new struct inside `spec.parameters.service` and has a form:
```        
        repackEnabled: true
        serviceLevel: besteffort
 ++  tls:
 ++      enabled: true
        vacuumEnabled: false
```
* test cases I did:
** new instance
** existing instance during migration from `master` branch
** existing instances with previous fixes and partially applied fields


zugao:
I don't understand what is the problem.

wejdross:
Our Appcat's comp-function is reading field `spec.parameters.service.tls.enabled` and based on that it sets correct value for SGcluster's `.spec.ssl.enabled`. But this field must be present, otherwise it'll default to false inside comp function because of `nil`. 

This PR ensures that after I deploy this change to clusters every xrd of XVSHNPostgreSQL would update it's spec and set new field `.tls.enabled: true`. Without this change after deployment of my previous PR, instances that were not created with tls settings in claim resulted in `ssl.enabled: false` because of nil value for TLS settings in xrd.

zugao (approved):
LGTM

zugao:
> Our Appcat's comp-function is reading field `spec.parameters.service.tls.enabled` and based on that it sets correct value for SGcluster's `.spec.ssl.enabled`. But this field must be present, otherwise it'll default to false inside comp function because of `nil`.

I guess if we had used `disabled: false` instead of `enabled: true` as default we would not have this problem now.

zugao:
Don't forget to squash your commits!

PR #279, opened 2024-12-04
ensure correct TLS settings for nested VSHNPostgreSQLs

wejdross:
## Summary

* Propagating TLS settings for nested PostgreSQLs


Kidswiss (approved):
Have you tested it on the lab?

wejdross:
@Kidswiss yeah, it's still deployed on LAB, so You can check it as well:
```
ka get xvshnpostgresqls.vshn.appcat.vshn.io -A --no-headers -o yaml | grep -i tls -A 1

We don't want to see any enabled: false
```
As a test case I created keycloak and nextcloud, afterwards TLS settings in backing PostgreSQL were set correctly

PR #280, opened 2024-12-05
Create AddOn API

zugao:
## Summary

* This PR changes API for how we add add-ons
* This PR fixes the problem with billing that the instance id of an add-on is the same as the parent service.

## Checklist

- [x] Update tests.


wejdross:
Can You explain this PR better? How is that going to help with billing? Why introducing Addons? - To generate  prometheus rule based on that? Why naming addon as 'office' instead of 'collabora'?

PR #283, opened 2024-12-12
Switch instances monitoring from cluster to user-workload monitoring

TheBigLee:
## Summary

This PR will switch the monitoring of the instances from cluster monitoring to user-workload monitoring.
This is required to have alerting and monitoring work for the end users

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/billing.go
+++ b/pkg/comp-functions/functions/common/billing.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+
 	xfnproto "github.com/crossplane/function-sdk-go/proto/v1beta1"
 	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 )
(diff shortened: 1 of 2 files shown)


pkg/comp-functions/functions/common/billing.go
```diff
@@ -75,6 +77,13 @@ func CreateBillingRecord(ctx context.Context, svc *runtime.ServiceRuntime, comp
 	}
 
 	p := &v1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{
+				// This label is required, the ensure that the rule is added to Prometheus and
+				// not the Thanos Rules in the Openshift User Workload Monitoring
+				"openshift.io/prometheus-rule-evaluation-scope": "leaf-prometheus",
```
Kidswiss, line 84: Do I understand correctly, that this is the magic sauce that will make the billing data go to the right place in mimir?

Kidswiss (approved):

pkg/comp-functions/functions/common/billing.go
TheBigLee: Openshift user-workload monitoring has an additional thanos-ruler rolled out (see https://docs.openshift.com/container-platform/4.17/observability/monitoring/monitoring-overview.html) and the prometheus rules are deployed to the thanos-rules by default. This leads to the issue that prometheus itself doesn't know the the recording rule and hence we can't forward the metrics from this recording rule to our mimir. This flag, ensures that our prometheus rule is rolled out to Prometheus itself rather than to the thanos-ruler

zugao (approved):
LGTM

PR #285, opened 2024-12-13
fixing name for nextcloud deployment in collabora install script

wejdross:
## Summary

* The Nextcloud deployment name changed with the latest nextcloud helm chart version and  had to update collabora install script argument to ensure proper installation

## Checklist

- [x] Update tests.


TheBigLee (approved):

pkg/comp-functions/functions/vshnnextcloud/collabora.go
```diff
@@ -540,7 +540,7 @@ func createInstallCollaboraJob(comp *vshnv1.VSHNNextcloud, svc *runtime.ServiceR
 							Command: []string{
 								"bash",
 								"-cefx",
-								fmt.Sprintf("oc exec -i deployments/%s -- /install-collabora.sh \"%s\" \"%s\" \"%s\"", comp.GetName(), svc.Config.Data["isOpenshift"], comp.GetName(), comp.Spec.Parameters.Service.Collabora.FQDN),
+								fmt.Sprintf("oc exec -i deployments/%s -- /install-collabora.sh \"%s\" \"%s\" \"%s\"", comp.GetName()+"-nextcloud", svc.Config.Data["isOpenshift"], comp.GetName(), comp.Spec.Parameters.Service.Collabora.FQDN),
```
zugao, line 543: please update and use `comp.GetWorkloadName()`

wejdross: This funtion returns GetName()
```
func (v *VSHNNextcloud) GetWorkloadName() string {
	return v.GetName()
}
```

why do we even keep such method?

wejdross: I just commited updated code to GetWorkloadName(), but I'm really unsure if that's the way we should go

zugao: TLDR: so that we have one single place for getting the name of the workload so that we reduce potential bugs in the future.

During a feature request we needed the workload names of all our services. Later the implementation changed and we were not relying on on the workload names. We still agreed to keep the functions anyway for the reason of this PR you opened. Unfortunately if you take a look at all our service api implementation you will realise that not all services have their workload name  `comp.GetName()`. You have to see these methods as constants for each service and most of them luckily have the same name as  `comp.GetName()`. To keep our code bug free, whenever there are constants we need to put them in our APIs.

zugao (approved):

PR #288, opened 2024-12-27
Add support for devcontainers

Kidswiss:
## Summary

* Adds support for devcontainers to streamline the dev ux

## Checklist

- [x] Update tests.

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+mkdir -p $HOME/.kube
+kubectl completion bash > /home/vscode/.kube/completion.bash.inc
+printf "
+source /usr/share/bash-completion/bash_completion
+source $HOME/.kube/completion.bash.inc
+complete -F __start_kubectl k
+" >> $HOME/.bashrc
+
+printf "
+source <(kubectl completion zsh)
+complete -F __start_kubectl k
+" >> $HOME/.zshrc
+
+make setup-kindev
+
+cp .kind/.kind/kind-config ~/.kube/config
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,23 @@
+FROM mcr.microsoft.com/devcontainers/go:dev-1.23-bookworm
+
+ARG TARGETOS
+ARG TARGETARCH
+
+ENV HELMVERS="v3.16.4"
+ENV YQVERSION="v4.44.6"
+ENV KUBECOLORVERSION="0.4.0"
+
+# Install kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl && chmod +x ./kubectl && mv ./kubectl /usr/local/bin
+
+# Install helm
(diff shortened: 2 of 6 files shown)


.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,19 @@
+mkdir -p $HOME/.kube
```
wejdross, line 1: `#!/bin/bash`

since by default we use Bash it makes sense to add shebang to avoid issues

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,19 @@
...
+kubectl completion bash > /home/vscode/.kube/completion.bash.inc
+printf "
+source /usr/share/bash-completion/bash_completion
+source $HOME/.kube/completion.bash.inc
+alias k=kubectl
```
wejdross, line 6: small request from my side:
```
alias ka='kubectl --as cluster-admin'
alias kwp='watch -n 0.1 kubectl get pod'
alias kwo='watch -n 0.1 kubectl get object'
alias kwoa='watch -n 0.1 kubectl --as cluster-admin get object'
```

wejdross: alias of `ll` is also missing in bash spec

wejdross (approved):
I like it a lot, it'll grow with time for sure with some customizations and aliases, but I'm ok with that, I think it might make even bigger sense to make dev container with necessary tooling etc. for components, so we have correct jsonnets, commodore etc. that should shorten onboarding drastically

wejdross:
the only thing is that rebuilding is enforced and it takes time to achieve, on metered connections it can take ages, maybe we could think of some caching in kindev?

.devcontainer/postCreateCommand.sh
Kidswiss: You can add your own aliases via dotfiles. I wouldn't clutter the defaults too much, as everybody wants different things.

Here's an example of my dotfiles: https://git.earthnet.ch/simon.beck/devcontainerdotfiles/src/branch/main/zshrc

Kidswiss:
> the only thing is that rebuilding is enforced and it takes time to achieve, on metered connections it can take ages, maybe we could think of some caching in kindev?

What do you mean by "rebuilding is enforced"? Once the container is built you can simply start it.

wejdross:
> > the only thing is that rebuilding is enforced and it takes time to achieve, on metered connections it can take ages, maybe we could think of some caching in kindev?
> 
> What do you mean by "rebuilding is enforced"? Once the container is built you can simply start it.

but any interruption, change to dotfiles, oom, whatever will result in rebuilding whole container, this is something I don't like, I can't  find any docs, how to shutdown gracefully (keep container in background) and shutdown IDE for exmaple.

Kidswiss:
> > > the only thing is that rebuilding is enforced and it takes time to achieve, on metered connections it can take ages, maybe we could think of some caching in kindev?
> > 
> > 
> > What do you mean by "rebuilding is enforced"? Once the container is built you can simply start it.
> 
> but any interruption, change to dotfiles, oom, whatever will result in rebuilding whole container, this is something I don't like, I can't find any docs, how to shutdown gracefully (keep container in background) and shutdown IDE for exmaple.

You can `docker stop` the container. In vscode if you do the `open in devcontainer` it will simply start the container again, without any rebuilding.

As to your other things, yes if it gets interrupted during the execution of the `postCreateCommand.sh` you might end up with a half-provisioned kindev. If it gets interrupted, you can do `cd .kind` and then `make vshnall`. It should then behave like normal kindev and continue where it left of. Or run the `postCreateCommand.sh` again by calling `.devcontainer/postCreateCommand.sh`.

Also, you don't have to rebuild the whole thing if you change your dotfiles. The dotfiles are stored in `~/dotfiles` and are a simple git repo. So if you change your dotfiles, you can just do something like this:

```
~/dotfiles
git pull
./install.sh
```

And your new things will be applied.

Rebuilding is only really required if something fundamentally changes in the `Dockerfile`. Or if you break the container somehow :D

PR #290, opened 2025-01-07
Add image repository parameter, to be able to override the registry

TheBigLee:
## Summary

Currently, the registry is hardcoded to docker.io. However, we want to be able to use a different registry in some cases (eg. to avoid rate-limiting). Using this parameter allows us to change the registry for the nextcloud image to another registry.

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnnextcloud/collabora.go
+++ b/pkg/comp-functions/functions/vshnnextcloud/collabora.go
@@ -536,7 +536,7 @@ func createInstallCollaboraJob(comp *vshnv1.VSHNNextcloud, svc *runtime.ServiceR
 					Containers: []corev1.Container{
 						{
 							Name:  comp.GetName() + "-install-collabora",
-							Image: "quay.io/appuio/oc:v4.13",
+							Image: svc.Config.Data["oc_image"],
 							Command: []string{
 								"bash",
 								"-cefx",
--- a/pkg/comp-functions/functions/vshnnextcloud/deploy.go
+++ b/pkg/comp-functions/functions/vshnnextcloud/deploy.go
@@ -223,7 +223,7 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 		extraInitContainers = []map[string]any{
 			{
 				"name":  "dbchecker",
-				"image": "docker.io/busybox",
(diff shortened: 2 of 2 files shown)


wejdross (approved):
there's also 
```
Image: "quay.io/appuio/oc:v4.13",
```
in collabora.go line 539 which we might want to migrate

zugao:
> there's also
> 
> ```
> Image: "quay.io/appuio/oc:v4.13",
> ```
> 
> in collabora.go line 539 which we might want to migrate

I would fix it as well in this PR

Kidswiss:
> > there's also
> > ```
> > Image: "quay.io/appuio/oc:v4.13",
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in collabora.go line 539 which we might want to migrate
> 
> I would fix it as well in this PR

quay.io doesn't have that restrictive rate limits though.

But might be handy in the future if we make it configurable.

Kidswiss (approved):

zugao:
> > > there's also
> > > ```
> > > Image: "quay.io/appuio/oc:v4.13",
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > in collabora.go line 539 which we might want to migrate
> > 
> > 
> > I would fix it as well in this PR
> 
> quay.io doesn't have that restrictive rate limits though.
> 
> But might be handy in the future if we make it configurable.

I agree it doesn't but it is pinned to oc 4.13. This will fail some time in the future.

TheBigLee:
> > > > there's also
> > > > ```
> > > > Image: "quay.io/appuio/oc:v4.13",
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > in collabora.go line 539 which we might want to migrate
> > > 
> > > 
> > > I would fix it as well in this PR
> > 
> > 
> > quay.io doesn't have that restrictive rate limits though.
> > But might be handy in the future if we make it configurable.
> 
> I agree it doesn't but it is pinned to oc 4.13. This will fail some time in the future.

I've parameterized the oc image as well now

zugao (approved):

PR #296, opened 2025-01-15
Make connection details split cluster compatible

Kidswiss:
## Summary

With this change we ensure that we never reference an instance namespace directly in the connection details' `WriteConnectionSecretToReference`. Instead, we hard code it to the Crossplane namespace. Then there's a new function in the runtime that will check each managed resource for connection details, if it's set, it will automatically create a secret in the instance namespace on the service cluster via the provider-kubernetes.
    
There's an annotation that can be set on the managed resources to disable this functionality for specific resources.

This is necessary, because the instance namespaces don't exist on the control clusters. As the `WriteConnectionSecretToReference` is a property of the managed resources, it's not aware about the instance namespace. That's why these changes are necessary. The references are automatically converted to point to the Crossplane namespace, while simultaneously deploying a new object that will write the secrets to the instance namespaces. This ensures that it will behave the same way as before.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnmariadb/mariadb_deploy.go
+++ b/pkg/comp-functions/functions/vshnmariadb/mariadb_deploy.go
@@ -19,6 +19,7 @@ import (
 	"github.com/vshn/appcat/v4/pkg/comp-functions/functions/common/maintenance"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	corev1 "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -260,7 +261,7 @@ func newRelease(ctx context.Context, svc *runtime.ServiceRuntime, values map[str
 			SkipPartOfReleaseCheck: true,
 		},
 	}
-	rel, err := common.NewRelease(ctx, svc, comp, values, cd...)
+	rel, err := common.NewRelease(ctx, svc, comp, values, comp.GetName()+"-release", cd...)
 	rel.Spec.ForProvider.Chart.Name = comp.GetServiceName() + "-galera"
 
 	return rel, err
@@ -433,6 +434,11 @@ func createMainService(comp *vshnv1.VSHNMariaDB, svc *runtime.ServiceRuntime, se
 		targetPort = 6033
 	}
 
+	serviceType := corev1.ServiceTypeClusterIP
+	if comp.Spec.Parameters.Network.ServiceType == "LoadBalancer" {
+		serviceType = corev1.ServiceTypeLoadBalancer
+	}
+
 	service := corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      serviceName,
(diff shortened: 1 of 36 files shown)


Kidswiss:
I still need to do some cleanup, but the code is ready for review.

pkg/comp-functions/functions/vshnmariadb/mariadb_deploy.go
```diff
@@ -473,3 +497,27 @@ func getMariaDBRootPassword(secretName string, svc *runtime.ServiceRuntime) ([]b
 	}
 	return secret.Data["mariadb-root-password"], nil
 }
+
+// we effectively have to allow all traffic that the service can be
+// accessed via the loadbalancer.
+func addLoadbalancerNetpolicy(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNMariaDB) error {
```
zugao, line 503: why this is the case? why the traffic has to be allowed from other instance namespaces?

pkg/comp-functions/functions/vshnpostgres/loadBalancer.go
```diff
@@ -104,3 +110,27 @@ func updateConnectionSecretWithLoadBalancerIP(ctx context.Context, svc *runtime.
 	svc.SetConnectionDetail("LOADBALANCER_IP", []byte(ip))
 
 }
+
+// we effectively have to allow all traffic that the service can be
+// accessed via the loadbalancer.
+func addLoadbalancerNetpolicy(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNPostgreSQL) error {
```
zugao, line 116: I think you can reuse this method across all services. Just pass `InfoGetter` resource instead of comp

pkg/comp-functions/functions/vshnpostgres/maintenance.go
```diff
@@ -202,3 +204,20 @@ func setPsqlMinorVersion(svc *runtime.ServiceRuntime, desiredCluster *stackgresv
 
 	return nil
 }
+
+func addStackgresCredentialsObserver(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNPostgreSQL, sgNamespace string) error {
```
zugao, line 208: In my PR of major upgrade I created a kube client to get this secret :(

zugao (commented):
Can you please add the reasoning behind this change in the description of the PR. It's clear what is the goal, I don't understand why do we need to do that.

pkg/comp-functions/functions/vshnmariadb/mariadb_deploy.go
Kidswiss: Otherwise, the external loadbalancer connection is also blocked. I'll open a  followup to do research if we can still somehow allow external connections while block internal connections.

pkg/comp-functions/functions/vshnpostgres/loadBalancer.go
Kidswiss: True, will do that.

pkg/comp-functions/functions/vshnpostgres/maintenance.go
Kidswiss: Please don't create a kube client, use observers. Kube clients will open a whole new world of PITA for non-converged setups...

zugao: here it is my comment :D

Kidswiss:
> Can you please add the reasoning behind this change in the description of the PR. It's clear what is the goal, I don't understand why do we need to do that.

I've added more context in the description, is the why now clearer?

pkg/comp-functions/functions/vshnpostgres/maintenance.go
zugao: Remember this thing, you will have to rebase on my PR or on master if I am able to deploy it soon

PR #297, opened 2025-01-15
Add new VSHNForgejo Service

wejdross:
## Summary

* This MR aims to provide base provisioning of VSHNForgejo

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/common/release.go
```diff
@@ -108,5 +109,9 @@ func NewRelease(ctx context.Context, svc *runtime.ServiceRuntime, comp InfoGette
 		},
 	}
 
+	if strings.HasPrefix(svc.Config.Data["chartRepository"], "oci://") {
+		release.Spec.ForProvider.Chart.URL = fmt.Sprintf("%s:%s", svc.Config.Data["chartRepository"], svc.Config.Data["chartVersion"])
+	}
+
```
wejdross, line 115: small workaround to make Forgejo chart pulling working

apis/vshn/v1/dbaas_vshn_Codey.go
```diff
@@ -0,0 +1,181 @@
+package v1
```
Kidswiss, line 1: Please don't use capital leters in filenames, that can lead to weird issues.

apis/vshn/v1/dbaas_vshn_Forgejo.go
```diff
@@ -0,0 +1,295 @@
+package v1
```
Kidswiss, line 1: Same here.

apis/vshn/v1/dbaas_vshn_Codey.go
```diff
@@ -0,0 +1,181 @@
...
+// Workaround to make nested defaulting work.
+// kubebuilder is unable to set a {} default
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_codeys.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.default={})"
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_codeys.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.size.default={})"
+//go:generate yq -i e ../../generated/vshn.appcat.vshn.io_codeys.yaml --expression "with(.spec.versions[]; .schema.openAPIV3Schema.properties.spec.properties.parameters.properties.service.default={})"
+
+// +kubebuilder:object:root=true
+
+// Codey is the API for creating Codey instances.
+type Codey struct {
```
Kidswiss, line 19: According to the design documentation the Codey claim should be called `Instance` and live under `codey.io/v1` and not under appcat.

apis/vshn/v1/dbaas_vshn_Codey.go
```diff
@@ -0,0 +1,181 @@
...
+	// MemoryRequests defines the requests amount of memory in units of bytes for an instance.
+	MemoryRequests string `json:"memoryRequests,omitempty"`
+
+	// MemoryLimits defines the limits amount of memory in units of bytes for an instance.
+	MemoryLimits string `json:"memoryLimits,omitempty"`
+
+	// Disk defines the amount of disk space for an instance.
+	Disk string `json:"disk,omitempty"`
+
+	// Plan is the name of the resource plan that defines the compute resources.
```
Kidswiss, line 86: We should probably make plan an enum, so users can't choose invalid plans.

A default would be nice as well.

apis/vshn/v1/dbaas_vshn_Codey.go
```diff
@@ -0,0 +1,181 @@
...
+	Plan string `json:"plan,omitempty"`
+}
+
+// CodeyStatus reflects the observed state of a Codey.
+type CodeyStatus struct {
+	NamespaceConditions         []Condition `json:"namespaceConditions,omitempty"`
```
Kidswiss, line 92: Let's get rid of these `*Conditions` fields in the status. They were a bad solution from the PnT days.

apis/vshn/v1/dbaas_vshn_Codey.go
```diff
@@ -0,0 +1,181 @@
...
+func (v *Codey) GetPDBLabels() map[string]string {
+	return map[string]string{}
+}
+
+func (v *Codey) GetSize() VSHNSizeSpec {
+	return v.Spec.Parameters.Size
+}
+
+func (v *Codey) GetBillingName() string {
+	return "appcat-" + v.GetServiceName()
```
Kidswiss, line 168: Is this correct for codey? Will this have an `appcat-` prefix as well?

apis/vshn/v1/dbaas_vshn_Forgejo.go
```diff
@@ -0,0 +1,295 @@
...
+	ServiceLevel VSHNDBaaSServiceLevel `json:"serviceLevel,omitempty"`
+
+	// Version contains supported version of Forgejo.
+	// Multiple versions are supported. Defaults to 10.0.0 if not set.
+	// +kubebuilder:default="10.0.0"
+	Version string `json:"version,omitempty"`
+}
+
+// VSHNForgejoSizeSpec contains settings to control the sizing of a service.
+type VSHNForgejoSizeSpec struct {
```
Kidswiss, line 102: This size spec is identical to the codey one, you could re-use some stuff here.

apis/vshn/v1/dbaas_vshn_Forgejo.go
```diff
@@ -0,0 +1,295 @@
...
+	// Disk defines the amount of disk space for an instance.
+	Disk string `json:"disk,omitempty"`
+
+	// Plan is the name of the resource plan that defines the compute resources.
+	Plan string `json:"plan,omitempty"`
+}
+
+// VSHNForgejoStatus reflects the observed state of a VSHNForgejo.
+type VSHNForgejoStatus struct {
+	NamespaceConditions         []Condition `json:"namespaceConditions,omitempty"`
```
Kidswiss, line 125: Same here, get rid of the `*Conditions`.

hack/bootstrap/template/api.txt
```diff
@@ -139,12 +161,12 @@ type {{.Name}}TLSSpec struct {
 
 // {{.Name}}Status reflects the observed state of a {{.Name}}.
 type {{.Name}}Status struct {
-	NamespaceConditions         []v1.Condition `json:"namespaceConditions,omitempty"`
-	SelfSignedIssuerConditions  []v1.Condition `json:"selfSignedIssuerConditions,omitempty"`
-	LocalCAConditions           []v1.Condition `json:"localCAConditions,omitempty"`
-	CaCertificateConditions     []v1.Condition `json:"caCertificateConditions,omitempty"`
-	ServerCertificateConditions []v1.Condition `json:"serverCertificateConditions,omitempty"`
-	ClientCertificateConditions []v1.Condition `json:"clientCertificateConditions,omitempty"`
+	NamespaceConditions         []Condition `json:"namespaceConditions,omitempty"`
```
Kidswiss, line 164: Let's remove those here as well.

pkg/comp-functions/functions/common/release.go
Kidswiss: Do OCI images require the version to be part of the url?

pkg/comp-functions/functions/vshnforgejo/deploy.go
```diff
@@ -0,0 +1,204 @@
...
+					"HOST":    "'{\"size\":100, \"recent_ratio\":0.25, \"ghost_ratio\":0.5}'",
+				},
+				"cron": map[string]any{
+					"ENABLED": true,
+					"archive_cleanup": map[string]any{
+						"SCHEDULE":   "@hourly",
+						"OLDER_THAN": "2h",
+					},
+				},
+				"database": map[string]any{
```
Kidswiss, line 97: I'd suggest to set sqlite to WAL mode. It will increase the sqlite performance quite a bit: https://forgejo.org/docs/latest/admin/database-preparation/#sqlite

```
"SQLITE_JOURNAL_MODE": "WAL",
```

pkg/comp-functions/functions/vshnforgejo/deploy.go
```diff
@@ -0,0 +1,204 @@
...
+					"DB_TYPE": "sqlite3",
+				},
+				"image": map[string]any{
+					"tag": comp.Spec.Parameters.Service.Version,
+				},
+				"indexer": map[string]any{
+					"ISSUE_INDEXER_TYPE":   "bleve",
```
Kidswiss, line 104: We should probably document somewhere that we enable this by default. The bleve search index will be the same size as all the code.

pkg/comp-functions/functions/vshnforgejo/deploy.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+		"redis-cluster": map[string]any{
+			"enabled": false,
+		},
+		"strategy": map[string]any{
+			"type": "Recreate",
+		},
+	}
+
+	for _, host := range comp.Spec.Parameters.Service.FQDN {
```
Kidswiss, line 176: Maybe the `unstructured` package could help make this more readable?

pkg/controller/webhooks/forgejo.go
```diff
@@ -0,0 +1,100 @@
...
+	if warning != nil || err != nil {
+		return warning, err
+	}
+
+	forgejo, ok := obj.(*vshnv1.VSHNForgejo)
+	if !ok {
+		return nil, fmt.Errorf("provided manifest is not a valid VSHNForgejo object")
+	}
+
+	if forgejo.Spec.Parameters.Service.FQDN == nil {
```
Kidswiss, line 61: You should be able to do that via CRD validation: https://book.kubebuilder.io/reference/markers/crd-validation

gitea.values.yaml
```diff
@@ -0,0 +1,18 @@
+gitea:
```
Kidswiss, line 1: Still needed?

Kidswiss (approved):
Some things to check.

apis/vshn/v1/dbaas_vshn_Codey.go
zugao: these conditions should be gone

pkg/comp-functions/functions/vshnforgejo/deploy.go
wejdross: 1. OCI image refused to be found without that tweak, I just saw that it's similarily solved in provider-helm/examples example
2. added that

tobru (changes requested):
Tested in the lab cluster, a few things:

- Please name the initial admin user `forgejo_admin`. The current name `gitea_admin` is perplexing.
- Why is `.spec.parameters.service.fqdn` an array? The spec ind https://kb.vshn.ch/app-catalog/service/forgejo-codey/architecture.html#_forgejo_by_vshn wants just a string
- I configured `.spec.parameters.security.deletionProtection` to `true`, but I could delete the claim anyway. Why?
- The instance isn't really deleted? The instance namespace is still there after deletion of the claim and the Forgejo instance happily running.
- Why is the kind of Codey called `Codey`? According to the spec, it should be named `Instance`. See https://kb.vshn.ch/app-catalog/service/forgejo-codey/architecture.html#_codey
- The Codey spec doesn't allow configuring the fqdn, see the spec docs: https://kb.vshn.ch/app-catalog/service/forgejo-codey/architecture.html#_codey
- The plan in Codey should be at `.spec.parameters.size.plan`, not at `.spec.parameters.size` - So we have it the same like other services allowing to define a plan.

wejdross:
> * perplexing



> Tested in the lab cluster, a few things:
> 
> * Please name the initial admin user `forgejo_admin`. The current name `gitea_admin` is perplexing.
> * Why is `.spec.parameters.service.fqdn` an array? The spec in https://kb.vshn.ch/app-catalog/service/forgejo-codey/architecture.html#_forgejo_by_vshn wants just a string
> * I configured `.spec.parameters.security.deletionProtection` to `true`, but I could delete the claim anyway. Why?
> * The instance isn't really deleted? The instance namespace is still there after deletion of the claim and the Forgejo instance happily running.
> * Why is the kind of Codey called `Codey`? According to the spec, it should be named `Instance`. See https://kb.vshn.ch/app-catalog/service/forgejo-codey/architecture.html#_codey
> * The Codey spec doesn't allow configuring the fqdn, see the spec docs: https://kb.vshn.ch/app-catalog/service/forgejo-codey/architecture.html#_codey
> * The plan in Codey should be at `.spec.parameters.size.plan`, not at `.spec.parameters.size` - So we have it the same like other services allowing to define a plan.

1. renamed user admin account
2. for nextcloud we had to migrate sonner than we thought and it was a headache, also helm chart expects an array, so we not reuse that
3. No idea, need debug <----
4. Codey was renamed to Instance - already deployed on LAB
5. FQDN no longer exists in API it's now generated from `.metadata.name`
6. Plan was reworked

tobru:
Just tested again:

> renamed user admin account

Seems to still be called `gitea_admin`

>  for nextcloud we had to migrate sonner than we thought and it was a headache, also helm chart expects an array, so we not reuse that

OK, let's stick with the array. Docs need to be updated accordingly.

> No idea, need debug

Right now, instances seem to get properly deleted again.

> Codey was renamed to Instance - already deployed on LAB

Thanks, looks good

> FQDN no longer exists in API it's now generated from .metadata.name

I now get:

```
The Instance "my-codey-prod" is invalid: spec.parameters.service.fqdn: Required value
```

> Plan was reworked

Thanks, looks good

apis/vshn/v1/dbaas_vshn_Forgejo.go
Kidswiss: The file still contains a capital letter.

Kidswiss (changes requested):
You forgot to adjust the name of the forgejo api file.

pkg/comp-functions/functions/vshnforgejo/deploy.go
wejdross: This is already in our arhcitecture document, but I'll add information that REPO indexing consumes 6 times more disk space than repo itself.

https://docs.gitea.com/administration/config-cheat-sheet#indexer-indexer

pkg/comp-functions/functions/vshnforgejo/deploy.go
wejdross: have You got a snippet?

wejdross:
LAB was restored to stable state and Codey/Forgejo in my humble opinion work now exactly as expected. Please check.

tobru:
Feedback from a users perspective testing `codey.io/v1/Instance`:

- The generated Ingress is not fully correct: It should use the claim name to generate the hostname and have a dot to separate the "app" subdomain. E.g. my claim is called "tobrucodey", I'd expect a hostname called tobrucodey.app.codey.ch. Right now it's called "tobrucodey-zxtj4-app.codey.ch"
- Shouldn't the composite better be named `XCodeyInstance` rather than `XInstance`? Does fell like an upcoming clash in the future to me.

Other than that, LGTM. Code review by Team Schedar still required before merging and deploying.

wejdross:
> Feedback from a users perspective testing `codey.io/v1/Instance`:
> 
> * The generated Ingress is not fully correct: It should use the claim name to generate the hostname and have a dot to separate the "app" subdomain. E.g. my claim is called "tobrucodey", I'd expect a hostname called tobrucodey.app.codey.ch. Right now it's called "tobrucodey-zxtj4-app.codey.ch

that's intentional to avoid clashes in names if multiple users create their instance using for example "test"

> * Shouldn't the composite better be named `XCodeyInstance` rather than `XInstance`? Does fell like an upcoming clash in the future to me.

that doesn't make any difference, I can rename that, but then I'd rather have CodeyInstances and xCodeyInstances, which was critized in previous review :D it's just a naming convention we have in schedar co we have claims wahtever.api/version and xrd xwhatever.api/version

pkg/comp-functions/functions/vshnforgejo/deploy.go
Kidswiss: We use it quite a lot already: https://github.com/search?q=repo%3Avshn%2Fappcat%20unstructured&type=code

It's basically helper functions to work with such `map[string]any` objects.

here's the doc: https://pkg.go.dev/k8s.io/apimachinery@v0.30.2/pkg/apis/meta/v1/unstructured

tobru:
> that's intentional to avoid clashes in names if multiple users create their instance using for example "test"

Then leave it for now, but still the `.` is needed for `instancename.app.codey.ch`. Am I able to add a custom domain name now or not? I've lost track.

> that doesn't make any difference, I can rename that, but then I'd rather have CodeyInstances and xCodeyInstances, which was critized in previous review :D it's just a naming convention we have in schedar co we have claims wahtever.api/version and xrd xwhatever.api/version

@Kidswiss What's your take here? We could also name it `CodeyInstance`, I could live with that.

wejdross:
> Then leave it for now, but still the `.` is needed for `instancename.app.codey.ch`. 
- done

> Am I able to add a custom domain name now or not? I've lost track.
- no, it's autogenerated

that's a good catch with a dot, I just pushed fix for tht in component-appcat, should be visible in LAB cluster in ~10 minutes

Kidswiss:
> > that doesn't make any difference, I can rename that, but then I'd rather have CodeyInstances and xCodeyInstances, which was critized in previous review :D it's just a naming convention we have in schedar co we have claims wahtever.api/version and xrd xwhatever.api/version
> 
> @Kidswiss What's your take here? We could also name it `CodeyInstance`, I could live with that.

On a technical level it really doesn't matter, it won't clash the same way `Instance` won't clash. As long as we won't have any other kind `Instance` within the same group (codey.io).

However, we might want to think about not going with such a generic name. For example if we have other services where we decide to have the kind `Instance`, then users will have to use the FQDN to get the correct ones.

For example:

We have`instances.codey.io` and some other `instances.miniflux.servala.io`. Now a customer creates a miniflux instance and then wants to list it via kubectl. If they do a `kubectl get instances` they won't see their miniflux. Because it will get the alphabetically first kind `Instance` which would be `instnaces.codey.io`. They'll have to do `kubectl get instances.miniflux.servala.io`. 

So I'd personally go with specific kinds: `CodeyInstance` or just `Codey`.

pkg/comp-functions/functions/vshnforgejo/deploy.go
```diff
@@ -0,0 +1,244 @@
...
+	if svc.Config.Data["ingress_annotations"] != "" {
+		annotations := map[string]string{}
+
+		err := yaml.Unmarshal([]byte(svc.Config.Data["ingress_annotations"]), &annotations)
+		if err != nil {
+			// do nothing and use default annotation which sets issuer to staging
+			svc.Log.Error(fmt.Errorf("cannot unmarshal ingress annotations"), "error", err)
+		}
+
+		values["ingress"].(map[string]any)["annotations"] = annotations
```
Kidswiss, line 228: This one could also be handled with `unstructured` or `SetNestedObjectValue`

Kidswiss (approved):
Just one more thing from my side.

wejdross:
> > > that doesn't make any difference, I can rename that, but then I'd rather have CodeyInstances and xCodeyInstances, which was critized in previous review :D it's just a naming convention we have in schedar co we have claims wahtever.api/version and xrd xwhatever.api/version
> > 
> > 
> > @Kidswiss What's your take here? We could also name it `CodeyInstance`, I could live with that.
> 
> On a technical level it really doesn't matter, it won't clash the same way `Instance` won't clash. As long as we won't have any other kind `Instance/XInstace` within the same group (codey.io).
> 
> However, we might want to think about not going with such a generic name. For example if we have other services where we decide to have the kind `Instance`, then users will have to use the FQDN to get the correct ones.
> 
> For example:
> 
> We have`instances.codey.io` and some other `instances.miniflux.servala.io`. Now a customer creates a miniflux instance and then wants to list it via kubectl. If they do a `kubectl get instances` they won't see their miniflux. Because it will get the alphabetically first kind `Instance` which would be `instnaces.codey.io`. They'll have to do `kubectl get instances.miniflux.servala.io`.
> 
> So I'd personally go with specific kinds: `CodeyInstance` or just `Codey`.

So, I'll just go with codeyinstances.codey.io

tobru (approved):
Still some open comments, otherwise LGTM

PR #298, opened 2025-01-17
Set image registry and repository

zugao:
## Summary

* We have to make sure we can configure registries and repositories for our helm charts in the component. We might want to change the default docker hub registry.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnkeycloak/deploy.go
+++ b/pkg/comp-functions/functions/vshnkeycloak/deploy.go
@@ -332,9 +332,6 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 			"--spi-events-listener-jboss-logging-success-level=info",
 			"--spi-events-listener-jboss-logging-error-level=warn",
 		},
-		"image": map[string]any{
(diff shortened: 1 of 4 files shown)


pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -398,6 +395,12 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 		"podAnnotations":     podAnnotations,
 	}
 
+	if svc.Config.Data["imageRegistry"] != "" {
```
Kidswiss, line 398: Maybe we should fall back to the current hardcoded value, if the value in data is empty?

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnkeycloak/deploy.go
zugao: I hardcoded it into the component.

Kidswiss (approved):

wejdross (approved):

PR #303, opened 2025-02-04
Set SSLMode for provider-sqls provider-config

TheBigLee:
## Summary

We need to pass the correct ssl mode to the provider-config for provider-sql in order to be able to connect to the postgresql server depending on the ssl mode of the server itself.

## Checklist

- [x] Update tests.


Kidswiss (approved):

Kidswiss (approved):

wejdross (approved):
I assume You tested it already if it creates roles correctly?

zugao (approved):

PR #305, opened 2025-02-05
Generalize Ingress creation, use wildcard cert with OCP apps domain

ThisIsntTheWay:
## Summary

* Generalize ingress creation logic
* Create separate ingresses that bundle FQDNs in the following way:
  * FQDN is part of an OCP clusters default apps domain and only has one subdomain (`sub1.apps...`)
  * FQDN is either _not_ part of said domain or has 2 or more subdomains (`sub2.sub1.apps...`)
    * Bundled ingresses will have "letsencrypt" or "wildcard" added to their name according to their TLS config
  * Services that only have one FQDN will only create a single ingress with no extra name according to those rules
* Use wildcard certs for the ingress bundling eligible FQDNs (empty TLS config)

Given an apps domain `example.com`:
```bash
❯ k get ing -o custom-columns="NAME:.metadata.name,HOSTS:.spec.rules[*].host"
NAME                               HOSTS
nextcloud-collabora-code-ingress   k-collabora.example.com
nextcloud-letsencrypt-ingress      next.nextcloud.example.com,past.nextcloud.example.com,presentcloud.my-domain.ch,nextcloud.my-domain.ch,future.tensecloud.my-domain.ch
nextcloud-wildcard-ingress         nextcloud.example.com,2-nextcloud.example.com

❯ k get ing nextcloud-letsencrypt-ingress  -o yaml | yq .spec.tls
- hosts:
    - next.nextcloud.example.com
    - past.nextcloud.example.com
    - presentcloud.my-domain.ch
    - nextcloud.my-domain.ch
    - future.tensecloud.my-domain.ch
  secretName: nextcloud-ingress-cert

❯ k get ing nextcloud-wildcard-ingress -o yaml | yq .spec.tls
- {}
```

This affects the following services:
- Nextcloud
- Keycloak
- Collabora
- Forgejo

Related PR in `component-appcat`: https://github.com/vshn/component-appcat/pull/623

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/common/ingress.go
+++ b/pkg/comp-functions/functions/common/ingress.go
@@ -1,0 +1,261 @@
+package common
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+	"gopkg.in/yaml.v2"
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+// IngressConfig contains general information for generating an Ingress obect
+type IngressConfig struct {
+	AdditionalAnnotations  map[string]string // Optional
+	AdditionalIngressNames []string          // Optional
+	FQDNs                  []string
+	ServiceConfig          IngressRuleConfig
+	TlsCertBaseName        string
+}
+
+// IngressRuleConfig describes an ingress rule configuration
+type IngressRuleConfig struct {
+	RelPath           string // Optional, defaults to "/"
+	ServiceNameSuffix string // Optional
+	ServicePortName   string // Has preference over ServicePortNumber
+	ServicePortNumber int32
+}
+
+// Checks if an FQDN is part of a reference FQDN, e.g. an OpenShift Apps domain; "*nextcloud*.apps.cluster.com".
+// Returns true if yes and FQDN is not a 2nd level subdomain (i.e. *sub2.sub1*.apps.cluster.com)
+func IsSingleSubdomainOfRefDomain(fqdn string, reference string) bool {
+	if !strings.Contains(fqdn, reference) || reference == "" {
+		return false
+	}
+
+	noSuffix, _ := strings.CutSuffix(fqdn, reference)
+	return len(strings.Split(noSuffix, ".")) == 2 // Handles prefixed dot of reference domain
+}
+
+// Obtain ingress annotations and optionally extend them using additionalAnnotations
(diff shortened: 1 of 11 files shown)


pkg/comp-functions/functions/common/ingress.go
```diff
@@ -0,0 +1,264 @@
...
+		})
+	}
+
+	return ingresses, nil
+}
+
+// Generate an Ingress containing a single FQDN using a TLS config as such:
+// FQDN is one subdomain ON defaultAppsDomain (e.g. sub1.apps.cluster.com) -> Empty TLS config (uses wildcard cert on OCP).
+// FQDN does not statisfy the former -> TLS config using a Let's Encrypt certificate.
+func GenerateIngress(comp InfoGetter, svc *runtime.ServiceRuntime, ingressConfig IngressConfig) (*netv1.Ingress, error) {
```
zugao, line 209: I would definitely add some unit tests for this function, also for others as well

pkg/comp-functions/functions/vshnkeycloak/ingress.go
```diff
@@ -33,8 +33,32 @@ func AddIngress(_ context.Context, comp *vshnv1.VSHNKeycloak, svc *runtime.Servi
...
+		ServiceConfig: common.IngressRuleConfig{
+			RelPath:           comp.Spec.Parameters.Service.RelativePath,
+			ServiceNameSuffix: "keycloakx-http",
+			ServicePortName:   "https",
+		},
+		TlsCertBaseName: "keycloak",
+	})
+	if err != nil {
+		return runtime.NewWarningResult(fmt.Sprintf("cannot generate ingress: %s", err))
+	} else {
```
zugao, line 48: I would remove here the else clause

zugao (changes requested):
This PR will touch existing services. Make sure you check the crossplane-diff of this change.

pkg/comp-functions/functions/vshnkeycloak/ingress.go
ThisIsntTheWay: Yeah makes no sense given that the if returns anyway if true, must've had tunnel vision.

pkg/comp-functions/functions/common/ingress.go
ThisIsntTheWay: Added such in 4b1aa1e0076e0f506f8bb799ebbe5dd0f8dba1ba

Kidswiss (changes requested):
LGTM

zugao (approved):
LGTM

Kidswiss (approved):

PR #310, opened 2025-02-12
Add Forgejo backup

ThisIsntTheWay:
## Summary

* Added: backup functionality to `VSHNForgejo`

## Checklist


pkg/comp-functions/functions/common/backup/backup.go
```diff
@@ -52,6 +52,9 @@ func AddK8upBackup(ctx context.Context, svc *runtime.ServiceRuntime, comp common
 }
 
 func createObjectBucket(ctx context.Context, comp common.InfoGetter, svc *runtime.ServiceRuntime) error {
+	if comp.GetName() == "" {
```
zugao, line 55: this should not be possible. The `comp` argument is the composite itself which cannot exist without a name.

pkg/comp-functions/functions/common/backup/backup.go
```diff
@@ -52,6 +52,9 @@ func AddK8upBackup(ctx context.Context, svc *runtime.ServiceRuntime, comp common
...
+		return fmt.Errorf("could not get composition name")
```
zugao, line 56: It's composite and not composition.

pkg/comp-functions/functions/common/release.go
```diff
@@ -80,7 +80,7 @@ func NewRelease(ctx context.Context, svc *runtime.ServiceRuntime, comp InfoGette
 
 	release := &xhelmv1.Release{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: comp.GetName(),
+			Name: resName,
```
zugao, line 83: why this has to change? This is in common package and will disrupt existing releases from other services, we have to be very carefully here.

pkg/comp-functions/functions/vshnforgejo/backup.go
```diff
@@ -0,0 +1,80 @@
...
+
+	err = updateRelease(svc, comp)
+	if err != nil {
+		return runtime.NewWarningResult(fmt.Sprintf("cannot update release with backup configuration: %s", err))
+	}
+
+	return nil
+}
+
+func updateRelease(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNForgejo) error {
```
zugao, line 45: This method looks fine but I notice we copy paste it already 3 or 4 times. Would be nice if we could move it to common package.

pkg/comp-functions/functions/vshnforgejo/script/backup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+forgejo dump --type tar -t /tmp/backup -V -f -
```
zugao, line 3: Add a new line here.

zugao (changes requested):
Some small changes

pkg/comp-functions/functions/common/backup/backup.go
ThisIsntTheWay: This actually did return an empty string on a few occasions.  
With an empty string, it creates a bucket with a bad policy `-bucket` which minio complains about and then the whole pipeline gets stuck until that bucket is deleted.

pkg/comp-functions/functions/vshnforgejo/backup.go
ThisIsntTheWay: Postponing refactoring this for now. Especially VSHNRedis has a more convoluted `updateRelease()` than the other services and would need some extra attention.

pkg/comp-functions/functions/common/backup/backup.go
zugao: Could you debug this part? Try to reproduce and see why it's empty

pkg/comp-functions/functions/vshnforgejo/script/backup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
wejdross, line 1:
```suggestion
#!/bin/bash -xef
```

pkg/comp-functions/functions/vshnforgejo/script/backup.sh
wejdross, line 3: please use absolute path here

wejdross (commented):
small things from me, I had mostly the same thoughts as Gabi

pkg/comp-functions/functions/common/release.go
ThisIsntTheWay: I remember observing `GetDesiredComposedResourceByName(comp.GetName() + "release")` returning _not found_, which still happens when I use `comp.GetName()`, but the release actually gets updated just fine...  
I attributed the broken pipeline to _this_ before I realized that `createObjectBucket` created a bad bucket with no composite name which was the actual culprit.  
I'll revert this.

pkg/comp-functions/functions/common/backup/backup.go
Kidswiss: I agree, this should not be empty at this point. Maybe we missed getting the composite somewhere?

pkg/comp-functions/functions/vshnforgejo/backup.go
Kidswiss: AFAIK VSHNRedis is still mostly PnT right? We should really migrate that at one point...

pkg/comp-functions/functions/vshnforgejo/script/backup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -xef
+
+/usr/local/bin/forgejo dump --type tar -t /tmp/backup -V -f -
```
Kidswiss, line 3: Maybe we should have `/tmp/backup` as an `emptyDir`? This will make it a bit cleaner. Also not sure if OpenShift/APPUiO likes having massive data in a non-emptyDir folder.

Kidswiss (changes requested):
All-in-All LGTM.

Just the empty comp name and some thoughts about the backup tmp dir.

pkg/comp-functions/functions/vshnforgejo/script/backup.sh
ThisIsntTheWay: This is actually already implemented:  
https://github.com/vshn/appcat/blob/23c494401b25b1efc9170906127c53b8a4e0dbe7/pkg/comp-functions/functions/vshnforgejo/deploy.go#L162-L169

Kidswiss: Oops, missed that one :D

pkg/comp-functions/functions/common/backup/backup.go
ThisIsntTheWay: I believe to have found the issue (`ServiceRuntime.desiredResources` is null), but not the cause.

The first run is pretty much guaranteed to have an error like this:
```log
2025-02-25T09:21:29.823Z | ERROR | FUNCTIONS | vshnforgejo/deploy.go:28 | Cannot bootstrap instance namespace | {"resource": "forgejo-test--d98hp", "claimNamespace": "test", "claimName": "forgejo-test-1-prap", "error": "cannot create rbac rules for forgejo instance: cannot get claim namespace: not found"}
```
(This also happens with other services, such as NextCloud)

In the case of Forgejos `AddBackup()`: Once it called `svc.GetDesiredComposite(comp)` during the **first** run, `comp` is just this (produced using json.Marshal):
```json
{"metadata":{"creationTimestamp":null},"spec":{"parameters":{"service":{"fqdn":null},"size":{"requests":{}},"scheduling":{},"backup":{"retention":{}},"maintenance":{},"security":{},"monitoring":{}},"writeConnectionSecretToRef":{}},"status":{"schedules":{"maintenance":{}}}}
```

Initially, I thought the cuplrit was `svc.SetDesiredKubeObject()` (or similar) never getting called.  
However, in the case of NextCloud, despite `DeployNextcloud()` returning early and thus not calling this function, `AddIngress()` can actually use `comp.GetName()` just fine:  
https://github.com/vshn/appcat/blob/23c494401b25b1efc9170906127c53b8a4e0dbe7/pkg/comp-functions/functions/vshnnextcloud/ingress.go#L55

So basically, the cause must be something else...

Kidswiss: Can you change `svc.GetDesiredComposite(comp)` where it fails to `svc.GetObservedComposite(comp)`?

Might be possible that the desired state of the composite isn't set at that point.

ThisIsntTheWay: Yep, that fixed it.  
Should we keep the empty string check? I think it wouldn't hurt as this helps preventing bad buckets from generating.

Kidswiss: Shouldn't hurt to keep it, so fine by me.

zugao (approved):
LGTM

Kidswiss (approved):
LGTM

PR #311, opened 2025-02-19
Fix postgres webhook to allow major upgrade when non HA

zugao:
## Summary

* When services has explicitly set instances to 1 then allow major upgrade

## Checklist

- [ ] Update tests.


ThisIsntTheWay (approved):
Looks good

wejdross (commented):
LGTMT

PR #312, opened 2025-02-20
Temporary full deletion protection

TheBigLee:
## Summary

## Checklist

- [ ] Update tests.


zugao (changes requested):
I would not set exists to true to all objects in the cluster. This can be disruptive and dangerous. It should be sufficient to just block out objects with the labels of ownerKind, ownerVersion, onwerGroup and ownerName. Line 54, 60, 66 and 72 would leave without changes

PR #314, opened 2025-02-26
Fix deletionProtection for user secrets

TheBigLee:
When creating a user using the userManagment feature in VSHNPostgreSQL and VSHNMariaDB the secret objectis created without the `allowDeletion` label for the deletionProtection webhook.

This prevents us, from deleting user secrets when a user is removed.

## Summary

## Checklist

- [ ] Update tests.


Kidswiss (approved):
LGTM

zugao (approved):
From logical perspective everything seems good but I would use 2 different sets of functions `AddCredentialsSecretWithoutProtection()` and `AddCredentialsSecretWithProtection()` instead of passing the arguments. It's more legible.

TheBigLee:
> From logical perspective everything seems good but I would use 2 different sets of functions `AddCredentialsSecretWithoutProtection()` and `AddCredentialsSecretWithProtection()` instead of passing the arguments. It's more legible.

I see your point. However, there are currently two functions involved `AddCredentialSecrets()` which uses `AddGenericSecret()`
I would have to duplicate both of them, which seems a bit pointless. WTYT?

zugao:
`I would have to duplicate both of them, which seems a bit pointless. WTYT?` If it's not feasable, create a new type bool and define 2 constants - AllowDeletion, NotAllowDeletion. WIll be easier to read the code....

TheBigLee:
> `I would have to duplicate both of them, which seems a bit pointless. WTYT?` If it's not feasable, create a new type bool and define 2 constants - AllowDeletion, NotAllowDeletion. WIll be easier to read the code instead of passing false or true....

I've added two constants as suggested

PR #316, opened 2025-02-27
Use separate observer for root password

Kidswiss:
## Summary

If another object is referenced via the connectionDetails for a kube object, it can happen that the kube object can't be deleted, if the referenced object is deleted first.
    
In this specific case, we used to reference the secret that StackGres creates to expose the root password via the connectionDetails of the `cluster` kube object. However that secret is managed by StackGres. So it will be created by StackGres once a sgcluster object is created.
    
This sgcluster object is managed by the `cluster` kube object, which leads to the race condition.
    
It can happen that StackGres is faster with removing the secret than provider-kubernetes in remove the `cluster` kube object. When that happens the `cluster` kube object doesn't get deleted, because provider-kubernetes throws observation warnings. These warnings prevent the deletion of the actual sgcluster object and the wrapping `cluster` kube object.

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/connection_details.go
+++ b/pkg/comp-functions/functions/vshnpostgres/connection_details.go
@@ -11,6 +11,7 @@ import (
 	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


Kidswiss:
IMHO that's a bug in provider-kubernete's behaviour. I'll open a ticket upstream about this.

zugao (approved):
LGTM

Kidswiss:
Upstream bug: https://github.com/crossplane-contrib/provider-kubernetes/issues/345

PR #321, opened 2025-03-07
Also check for non empty value when waiting for connection details

Kidswiss:
## Summary

Even thought the field in a connectionDetail map might already be populated, it might happen that it still contains an empty value.

This commit adds an additional check to the waiting logic to also make sure that the value is not empty.

This will fix a racecondition where a pod can start with an empty value on the very first provisioning.
## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -1080,7 +1080,7 @@ func (s *ServiceRuntime) WaitForDesiredDependencies(mainResource string, depende
 }
 
 // WaitForObservedDependenciesWithConnectionDetails does the same as WaitForDependencies but additionally also checks the given list of fields against the
-// available connection details.
+// available connection details. It checks whether the field exists and has a non-empty value.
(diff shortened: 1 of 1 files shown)


pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1080,7 +1080,7 @@ func (s *ServiceRuntime) WaitForDesiredDependencies(mainResource string, depende
 }
 
 // WaitForObservedDependenciesWithConnectionDetails does the same as WaitForDependencies but additionally also checks the given list of fields against the
-// available connection details.
+// available connection details. It checks whether the fiel exists and has a non-empty value.
```
zugao, line 1083:
```suggestion
// available connection details. It checks whether the field exists and has a non-empty value.
```

zugao (approved):
LGTM

wejdross (approved):

Kidswiss:
Tested on the lab and locally. I was not able to observe issues with Keycloak provisioning anymore.

PR #323, opened 2025-03-11
Protect the certificate by the cluster

Kidswiss:
## Summary

The cluster object references the certificates as connection details. To make sure the deprovisioning works, we now deploy additional usages that ensure the certs will be removed after the cluster object.

This can sometimes lead to raceconditions during the de-provisioning as `provider-kubernetes` will not reconcile the object anymore if the referenced object doesn't exist anymore.

See upstream bug: https://github.com/crossplane-contrib/provider-kubernetes/issues/345

## Checklist

- [x] Update tests.


zugao (approved):
LGTM

PR #326, opened 2025-03-13
adding diff check for CRDS/ directory

wejdross:
## Summary
As part of improvement of our workflow we need to set check for crds/ directory to break if there's not generated and uncommited content. This branch ensures that. 

## Checklist
- [x] Categorize the PR by setting a good title and adding one of the labels: `bug`, `enhancement`, `documentation`, `change`, `breaking`, `dependency`
- [x] Update tests.


.github/workflows/diff.yml
```diff
@@ -6,25 +6,53 @@ jobs:
...
-             https://git.vshn.net/api/v4/projects/58084/trigger/pipeline
+             https://git.vshn.net/api/v4/projects/58084/trigger/pipeline | jq .web_url)
+          echo "$response"
+          echo "::set-output name=response::$response"
+      
+      - name: Update PR description
+        run: |
+          pr_number=${{ steps.get_pr.outputs.pr_number }}
+          pr_description=${{ steps.get_pr.outputs.pr_description }}
+          new_line="[ ] Checked out Your new diff at: \"${{ steps.trigger_diff.outputs.response }}\""
```
Kidswiss, line 54: Would this add a new line every time we push to the PR?

Maybe it's a bit better structured if we instead add comments?

wejdross:
hello world

TheBigLee (approved):

PR #327, opened 2025-03-17
Add non-converged setup support for the API server

TheBigLee:
## Summary

* This PR will add support for non-converged setups to our APIServer

## Checklist

- [x] Update tests.


pkg/apiserver/appcat/appcat.go
```diff
@@ -34,6 +35,10 @@ func New() restbuilder.ResourceHandlerProvider {
 		if err != nil {
 			return nil, err
 		}
+		err = providerkubernetesv1alpha1.SchemeBuilder.AddToScheme(c.Scheme())
```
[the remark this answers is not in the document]
TheBigLee: Fixed

pkg/apiserver/vshn/k8up/k8up.go
```diff
@@ -39,11 +39,15 @@ type ConcreteSnapshotHandler struct {
...
 
 // Get returns the snapshot with the given ID.
-func (c *ConcreteSnapshotHandler) Get(ctx context.Context, id, instanceNamespace string) (*k8upv1.Snapshot, error) {
+func (c *ConcreteSnapshotHandler) Get(ctx context.Context, id, instanceNamespace string, scClient client.Client) (*k8upv1.Snapshot, error) {
 
 	snapshot := &k8upv1.Snapshot{}
-
-	err := c.client.Get(ctx, client.ObjectKey{Namespace: instanceNamespace, Name: id}, snapshot)
+	var err error
+	if scClient != nil {
```
zugao, line 46: at this point in code is it not yet clear if you are in converged or non-converged mode?

pkg/apiserver/vshn/mariadb/vshnmariadb.go
```diff
@@ -39,3 +47,47 @@ func (c *concreteMariaDBProvider) ListVSHNMariaDB(ctx context.Context, namespace
...
+	err = c.client.Get(ctx, client.ObjectKey{Name: secretRef.Name, Namespace: secretRef.Namespace}, &secret)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	kubeconfig := secret.Data[secretRef.Key]
+
+	return kubeconfig, nil
+}
+
```
zugao, line 71: Please ad a comment and describe what this method does.

pkg/apiserver/vshn/mariadb/vshnmariadb.go
```diff
@@ -39,3 +47,47 @@ func (c *concreteMariaDBProvider) ListVSHNMariaDB(ctx context.Context, namespace
 
 	return instances, nil
 }
+
```
zugao, line 50: Please ad a comment and describe what this method does.

pkg/apiserver/vshn/nextcloud/vshnnextcloud.go
```diff
@@ -41,3 +50,68 @@ func (c *concreteNextcloudProvider) ListVSHNnextcloud(ctx context.Context, names
 
 	return instances, nil
 }
+
+func (k *concreteNextcloudProvider) GetKubeConfig(ctx context.Context, instance vshnv1.VSHNNextcloud) ([]byte, error) {
```
zugao, line 54: This method seems to be duplicated

zugao (changes requested):
Ok so first, there is a lot of duplicated code, I think you can reuse one single implementation across multiple service types (GetKubeClient() and GetKubeConf()). Second, this code smells badly:
```
	if scClient != nil {
		// do
	} else {
		// else
	}
```

You need to have this info way before calling the leaf ends, find a way to understand what client to use right after the program starts.

pkg/apiserver/vshn/k8up/k8up.go
TheBigLee: The APIServer itself doesn't know and doesn't care if it's in converged or non-converged mode

pkg/apiserver/vshn/nextcloud/vshnnextcloud.go
TheBigLee: I've refactored this whole part and created a single method for all services

Kidswiss (approved):

pkg/apiserver/vshn/nextcloud/backup.go
```diff
@@ -37,6 +38,15 @@ func New() restbuilder.ResourceHandlerProvider {
 			return nil, err
 		}
 
+		cache, err := ristretto.NewCache(&ristretto.Config[string, []byte]{
```
Kidswiss, line 41: This call is duplicated several times. It might make more sense to keep it in the common.go file.

Kidswiss (changes requested):
Do we really need a complicated caching library dependency for this?

The amount of service clusters for each controlplane will be relatively static and small. So a simple hashmap would probably suffice.

zugao:
> The amount of service clusters for each controlplane will be relatively static and small. So a simple hashmap would probably suffice.

Also a good idea. I will leave to Bigli to decide whether to change the code or not. I believe it should be ok to keep the cache as it is. We might needed in the future.

pkg/apiserver/vshn/nextcloud/backup.go
TheBigLee: Fixed

TheBigLee:
> Do we really need a complicated caching library dependency for this?
> 
> The amount of service clusters for each controlplane will be relatively static and small. So a simple hashmap would probably suffice.

I've removed the whole caching logic as it's irrelevant considering all other API calls we have.

Kidswiss (approved):

PR #328, opened 2025-03-17
Fix ingress using wrong service names

ThisIsntTheWay:
## Summary

If a Forgejo or Nextcloud service is created without having its service name in the name, then the helm chart will add `forgejo/nextcloud` to the service name.  

For example, a `VSHNForgejo` with name `my-forgejo-instance` will create this service: `my-forgejo-instance-pcdj9-http`.  
On the other hand, one with the name `my-cool-instance` creates this service: `my-cool-instance-cgrjz-forgejo-http`.
Same with Nextcloud.

The ingress creation logic currently does not account for that.

This PR fixes this issue for:
- Forgejo & Codey
- Nextcloud

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnforgejo/deploy.go
+++ b/pkg/comp-functions/functions/vshnforgejo/deploy.go
@@ -173,11 +173,16 @@ func addForgejo(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.V
 		},
 	}
 
+	svcNameSuffix := "http"
+	if !strings.Contains(comp.GetName(), "forgejo") {
+		svcNameSuffix = "forgejo-" + svcNameSuffix
+	}
+
 	svc.Log.Info("Adding ingress")
 	ingressConfig := common.IngressConfig{
 		FQDNs: comp.Spec.Parameters.Service.FQDN,
 		ServiceConfig: common.IngressRuleConfig{
-			ServiceNameSuffix: "http",
+			ServiceNameSuffix: svcNameSuffix,
 			ServicePortNumber: 3000,
 		},
 		TlsCertBaseName: "forgejo",
--- a/pkg/comp-functions/functions/vshnnextcloud/ingress.go
+++ b/pkg/comp-functions/functions/vshnnextcloud/ingress.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	xfnproto "github.com/crossplane/function-sdk-go/proto/v1"
 	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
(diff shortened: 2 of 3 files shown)


pkg/comp-functions/functions/vshnforgejo/deploy.go
```diff
@@ -173,11 +173,16 @@ func addForgejo(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.V
 		},
 	}
 
+	svcNameSuffix := "http"
+	if !strings.Contains(comp.GetName(), "forgejo") {
```
zugao, line 177: Can't you use GetWorkloadName() function?

pkg/comp-functions/functions/vshnnextcloud/ingress.go
```diff
@@ -23,9 +24,15 @@ func AddIngress(_ context.Context, comp *vshnv1.VSHNNextcloud, svc *runtime.Serv
 		return runtime.NewFatalResult(fmt.Errorf("FQDN array is empty, but requires at least one entry, %w", errors.New("empty fqdn")))
 	}
 
+	var svcNameSuffix string
+	if !strings.Contains(comp.GetName(), "nextcloud") {
```
zugao, line 28: This will recreate the ingress object is that right?

ThisIsntTheWay: Nope, it just updates it

pkg/comp-functions/functions/vshnforgejo/deploy.go
ThisIsntTheWay: Reposting from chat (just for transparency):  
For Forgejo this would work, but in the case of VSHNextCloud, the collabora service name doesn't use the `nextcloud` suffix, so the ingress would still point at the wrong svc.

Kidswiss: It might not work everywhere, but having a function like `getHelmEscapedName()` or something on the composites to account for the suffixing might be an idea.

This suffix stuff has bitten us before with minio and mariadb as well... It seems like a popular pattern in the helm world, for whatever reason...

zugao: Agree with Simon, let introduce another function that would help here.

ThisIsntTheWay: After discussing this with @Kidswiss, we've decided to leave this as is.  
If we were to add such a function, we would basically end up recreating `GetWorkloadName()` as most helm charts have a `blabla.fullname` helper func which basically produces the some output as GetWorkloadName.

zugao (approved):
LGTM

PR #329, opened 2025-03-18
Add TLS for SPKS Redis 7

Kidswiss:
## Summary

* The comp function deploys certificates with the correct SANs
* It makes sure that the StatefulSet gets scaled down properly if the tls state changes
* Add logic to disable connection detail handling for all desired resources, as this will clash with SPK's implementation
* To make the scaling observation faster, it will write to the composite's status as long as the sts is still scaling down and hasn't reached 0 replicas

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/spks/spksredis/tls.go
```diff
@@ -0,0 +1,210 @@
...
+
+	err := svc.SetDesiredKubeObject(sts, comp.GetName()+"-sts-observer", runtime.KubeOptionObserve)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// isSTSScaling will return true, if the release it set to 0 replicas and
```
zugao, line 171:
```suggestion
// isSTSScaling will return true, if the release is set to 0 replicas and
```

pkg/comp-functions/functions/spks/spksredis/tls.go
```diff
@@ -0,0 +1,210 @@
...
+// the sts hasn't fully scaled down yet.
+// This is necessary to ensure that we don't have mixed tls/non-tls members in the cluster
+// at the same time.
+func isSTSScaling(obsRelease *xhelm.Release, comp *spksv1alpha1.CompositeRedisInstance, svc *runtime.ServiceRuntime) bool {
```
zugao, line 175: I wonder what is the consequence of leaving helm to normally rollout the changes? What happens when there are mixed pods with tls/non-tls?

pkg/comp-functions/functions/spks/spksredis/tls.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return false
+}
+
+// updateScaleTimestamp will update a field in the status called `scaleTimeStamp`.
+// This will instantly trigger a new reconcile.
+// Making the switch from TLS to non-TLS much faster on SPK. Because SPKS has a
+// default reconcile period of 10 minutes.
+func updateScaleTimestamp(svc *runtime.ServiceRuntime, comp *spksv1alpha1.CompositeRedisInstance) error {
```
zugao, line 207: I hope this will not interfere in heavy used clusters such as prod. With how many instances have you tested this feature? (not just redis 7)

zugao (approved):
LGTM but I didn't entered in too much details as I pretty much lack the overall context here. Some typos and curiosities from my side

pkg/comp-functions/functions/spks/spksredis/tls.go
Kidswiss: If you leave it to helm, or rather the sts controller, it will break the cluster and get stuck.

The STS controller will start with the highest numbered pod and restart it with the new settings. The pod will then attempt to connect to the other instances via TLS, which will fail. After some time the pod will force switch to be a master and at that point the Redis cluster won't be consistent anymore.

That's why we scale it manually first.

pkg/comp-functions/functions/spks/spksredis/tls.go
Kidswiss: It's currently only implement in Redis 7 and it will only trigger if you switch from tls to no-tls or vice-versa. It won't trigger on newly provisioned instances either.

pkg/comp-functions/functions/spks/spksredis/tls.go
zugao: perfect, thank you

PR #330, opened 2025-03-20
Add appcat release management in maintenance

zugao:
## Summary

* Add release management in the maintenance of each service instance for our CI/CD pipeline.

We already assume that the composition revisions have _metadata.appcat.vshn.io/revision_ label set. Going through all service type composition revision, pick revisoins with the highest revision number and get the above revision label to pass it to the claim's _compositionRevisionSelector_.

## Checklist

- [x] Update tests.

### DIFF
--- a/cmd/maintenance.go
+++ b/cmd/maintenance.go
@@ -2,11 +2,13 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/vshn/appcat/v4/pkg/auth/stackgres"
+	"github.com/vshn/appcat/v4/pkg/maintenance/release"
 	"net/http"
 	"time"
-
-	"github.com/vshn/appcat/v4/pkg/auth/stackgres"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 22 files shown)


cmd/maintenance.go
```diff
@@ -74,64 +96,72 @@ func newMaintenanceCMD() *cobra.Command {
...
-		m = maintenance.NewForgejo(kubeClient, getHTTPClient())
+		m = maintenance.NewForgejo(kubeClient, getHTTPClient(), vh, log)
 	default:
 
 		panic("service name is mandatory")
 	}
 
-	return m.DoMaintenance(cmd.Context())
+	if err = errors.Join(
+		m.DoMaintenance(ctx),
```
zugao, line 150: I have decided to continue releasing the latest version even though maintenance might fail

pkg/comp-functions/functions/vshnpostgres/maintenance.go
```diff
@@ -53,18 +53,6 @@ var (
 				"get",
 			},
 		},
-		{
-			APIGroups: []string{
-				"vshn.appcat.vshn.io",
-			},
-			Resources: []string{
-				"vshnpostgresqls",
```
zugao, line 61: this should be unnecessary as we don't expect vshnpostgresqls to be in the instance namespace, therefore it has been moved to the cluster role that is defined in the component

pkg/maintenance/release/release.go
```diff
@@ -0,0 +1,198 @@
...
+func (vh *DefaultVersionHandler) getLatestRevisionLabel(ctx context.Context) (string, error) {
+	cr, err := vh.getLatestRevision(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch latest revision: %w", err)
+	}
+
+	revision, exists := cr.GetLabels()["metadata.appcat.vshn.io/revision"]
+	if !exists || revision == "" {
+		vh.log.Info("Label metadata.appcat.vshn.io/revision is missing from composition revision")
+		return "", nil
```
zugao, line 91: Will not return an error at the moment. It will break all our maintenance jobs.

cmd/maintenance.go
Kidswiss: Makes sense.

If we ever have a bug in the main maintenance we could never fix it automagically otherwise.

Maintenance is broken -> we release new version that fixes it -> it never gets applied because maintance is broken.

PR #332, opened 2025-03-21
Check Codey FQDN uniqueness

ThisIsntTheWay:
## Summary

* Checks if a to-be-created  `CodeyInstance` will use an unique FQDN, deny its creation otherwise

## Checklist

- [ ] Update tests.

### DIFF
--- a/pkg/controller/webhooks/codey.go
+++ b/pkg/controller/webhooks/codey.go
@@ -1,0 +1,133 @@
+package webhooks
+
+import (
+	"context"
+	"fmt"
+
+	codey "github.com/vshn/appcat/v4/apis/codey"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+//+kubebuilder:webhook:verbs=create;update,path=/validate-codey-io-v1-codeyinstance,mutating=false,failurePolicy=fail,groups=codey.io,resources=codeyinstances,versions=v1,name=codeyinstance.codey.io,sideEffects=None,admissionReviewVersions=v1
+
+//+kubebuilder:rbac:groups=codey.io,resources=codeyinstances,verbs=get;list;watch;patch;update
+//+kubebuilder:rbac:groups=codey.io,resources=codeyinstances/status,verbs=get;list;watch;patch;update
+
+const (
+	codeyUrlSuffix = ".app.codey.ch"
+)
+
+var (
+	codeyGK = schema.GroupKind{Group: "codey.io", Kind: "CodeyInstance"}
+	codeyGR = schema.GroupResource{Group: codeyGK.Group, Resource: "codeyinstance"}
+)
+
+var _ webhook.CustomValidator = &CodeyInstanceWebhookHandler{}
+
+type CodeyInstanceWebhookHandler struct {
+	DefaultWebhookHandler
+}
+
+// SetupCodeyInstanceWebhookHandlerWithManager registers the validation webhook with the manager.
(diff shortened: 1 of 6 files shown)


pkg/controller/webhooks/codey.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := isCodeyFqdnUnique(codeyFqdn); err != nil {
+		return nil, fmt.Errorf("failed FQDN validation: %v", err)
+	}
+
+	return p.DefaultWebhookHandler.ValidateUpdate(ctx, oldObj, newObj)
+}
+
+// Checks if a given FQDN is already in use by some CodeyInstance in the cluster
+func isCodeyFqdnUnique(fqdn string) error {
+	config, err := rest.InClusterConfig()
```
Kidswiss, line 99: `DefaultWebhookHandler`, which is embedded into `CodeyInstanceWebhookHandler` already has a `client.Client`, which will have the right configurations. Please use that one instead of ad-hoc creating one. It will probably generate issues down the road with split control-plane and service-clusters.

pkg/controller/webhooks/codey.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, namespace := range namespaceList.Items {
+		ingressList, err := clientset.NetworkingV1().Ingresses(namespace.Name).List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return fmt.Errorf("failed listing ingresses in namespace %s: %v", namespace.Name, err)
+		}
+
+		for _, ingress := range ingressList.Items {
+			for _, rule := range ingress.Spec.Rules {
+				if rule.Host == fqdn {
+					return fmt.Errorf("codey FQDN '%s' already present in namespace '%s'", fqdn, namespace.Name)
```
Kidswiss, line 124: Printing the namespace here could potentially leak sensitive information like customer names.

Just print, that it's already in. Also check out the `field.Error` helper functions, they produce nice errors with specific fields that the user should take a look at.

pkg/controller/webhooks/codey.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		panic(err.Error())
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	namespaceList, err := clientset.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "appcat.vshn.io/ownerkind=XVSHNForgejo",
```
Kidswiss, line 109: This will not only consider codey instances, but vshnforgejo ones as well... But I guess that's fine as URLs should be globally unique anyway.

Kidswiss (changes requested):

pkg/controller/webhooks/codey.go
ThisIsntTheWay: True, but I coudln't find other ways of only identifying Codey namespaces

pkg/controller/webhooks/codey.go
ThisIsntTheWay: Nice tip with `field.Error`, using this now produces something like:  
`metadata.name: Invalid value: "codey-instance": produces a codey FQDN (codey-instance.app.codey.ch) that is already in use`

Kidswiss (approved):

pkg/controller/webhooks/codey.go
```diff
@@ -0,0 +1,121 @@
...
+		return fmt.Errorf("failed listing ingresses: %v", err)
+	}
+
+	for _, ingress := range ingressList.Items {
+		for _, rule := range ingress.Spec.Rules {
+			if rule.Host == fqdn {
+				return field.Invalid(
+					field.NewPath("metadata", "name"),
+					strings.Split(fqdn, ".")[0],
+					fmt.Sprintf("produces a codey FQDN (%s) that is already in use", fqdn),
```
TheBigLee, line 114: nit: would phrase this differently: Problaby like `The FQDN %s is already in use. Please choose another FQDN.

TheBigLee (approved):

pkg/controller/webhooks/codey.go
ThisIsntTheWay: Good idea, though I think it'd also be good to retain the "produces" part to make it clear that `.metadata.name` is the 'cause' of the error.

PR #335, opened 2025-03-26
Make any image configurable

Kidswiss:
## Summary

## Checklist

- [ ] Update tests.


zugao (approved):
I absolutely agree that we need to move the images and registries outside appcat, to the component. I would even go further and say that we should validate the Data map of composition in one single palce, preferably when the appcat starts. it should not allow to run if important data is missing. This might even lead to data/service loss in edge cases.

Kidswiss:
> This might even lead to data/service loss in edge cases.

Yeah, such race conditions were the root cause for incidents in the past. As Crossplane now supports arbirtrary pseudo k8s objects for inputs we can actually switch to specific objects for each service individually.

PR #337, opened 2025-03-27
Fix nilpointer

Kidswiss:
## Summary

This fixes a nilpointer introduced by the refactored certificate handling.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/sliexporter/probes/postgresql_test.go
+++ b/pkg/sliexporter/probes/postgresql_test.go
@@ -28,7 +28,7 @@ func TestPostgreSQL_Probe(t *testing.T) {
 	password := "bar"
 	db := "buzz"
 
-	res, err := pool.Run("postgres", "12.3", []string{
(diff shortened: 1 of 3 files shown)


pkg/sliexporter/probes/postgresql_test.go
```diff
@@ -28,7 +28,7 @@ func TestPostgreSQL_Probe(t *testing.T) {
 	password := "bar"
 	db := "buzz"
 
-	res, err := pool.Run("postgres", "12.3", []string{
+	res, err := pool.Run("quay.io/lib/postgres", "12.3", []string{
```
Kidswiss, line 31: We ran into docker hub rate limit, so I changed to the quay mirror.

wejdross (approved):

pkg/sliexporter/probes/postgresql_test.go
zugao, line 31: we have to externalise these images...

zugao (commented):
LGTM

PR #341, opened 2025-04-04
Improve responsiveness with high poll interval

Kidswiss:
## Summary

This change will make Crossplane react much quicker to changes in PostgreSQL that will trigger a restart.
    
The reason this will take long if the poll interval for Crossplane is set to a high value is because of the asnc restart method we employ.
    
If a restart is needed for PostgreSQL then Stackgres will set a`PendingRestart` condition to true on the sgcluster object. With the default poll rate of 1 minute it will then take about 1 minute for the restart to actually trigger.
    
With a poll rate of 1h it would then take 1 hour for the restart to trigger, because Crossplane will then only poll the latest state of the sgcluster object every 60 minutes.
    
The default poll rate does not apply to Managed Resources that are not marked as ready. Then it will reconcile much quicker.
    
So if there's a change between the observed and desired state of the `profile` object we explicitly set the managed resource of the sgcluster to unready. This will then trigger the actualy restart even quicker than before.

A higher poll rate in Crossplane will massively reduce the load on the Kubernetes API.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
+++ b/pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
@@ -194,6 +194,8 @@ func createStackgresObjects(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, sv
 		return err
 	}
 
+	setClusterUnreadyIfProfilesNotEqual(svc)
+
 	return nil
 }
 
@@ -302,6 +304,7 @@ func createSgInstanceProfile(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, s
 		err = fmt.Errorf("cannot create sgInstanceProfile: %w", err)
 		return err
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


Kidswiss:
We could even backport this to the Redis TLS logic recently implemented for SPKS.

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
```diff
@@ -671,3 +687,31 @@ func createCopyJob(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) err
...
 	return nil
 }
+
+func setClusterUnreadyIfProfilesNotEqual(svc *runtime.ServiceRuntime) {
+	desProfile := &sgv1.SGInstanceProfile{}
+	obsProfile := &sgv1.SGInstanceProfile{}
+
+	changed := false
+
+	err := svc.GetObservedKubeObject(obsProfile, "profile")
```
zugao, line 697: Isn't this the other way around? on first cycle the desired is missing and not the observed?

pkg/comp-functions/functions/vshnpostgres/restart.go
```diff
@@ -16,7 +16,7 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 )
 
-var sgDbOpsRestartRetention time.Duration = 30 * 24 * time.Hour
+var sgDbOpsRestartRetention time.Duration = 10 * time.Minute
```
zugao, line 19: Can you add a comment here? why this is now much lower?

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
Kidswiss: You mixed it up :) The desired state gets set during the comp function in `createSgInstanceProfile(ctx, comp, svc)` and the observed state will only be available after the first cycle, sometimes even after a few, depending on how fast provider-kubernetes is.

pkg/comp-functions/functions/vshnpostgres/restart.go
Kidswiss: It's actually explained in the commit message:

On clusters with many instance there can be thousands of old sgdbops restarts that noone cares about and just generate load.

pkg/comp-functions/functions/vshnpostgres/postgresql_deploy.go
zugao: ah I see.. chicken egg problem :D who's first?

zugao (approved):
LGTM

PR #342, opened 2025-04-08
Switch backups to use the default region

Kidswiss:
## Summary

With this change it's now possible to select the default region/zone for the backups.
    
This enables us to deliver true off-site backups, so in case of a whole datacenter not being available anymore it's possible to restore.

This PR also fixes an annoying issue with Exoscale IAM name clashes. It was previously not possible to provision the buckets for nested services where both of them need a bucket (currently only Nextcloud). Also due to our billing re-creating the same instance in the same claim-namespace also failed.

## Checklist

- [x] Update tests.


zugao (approved):
How the region is populated?

PR #348, opened 2025-04-29
Add a Minor and Patch update strategy

Kidswiss:
## Summary

This commit adds a new helm uptade strategy that also allows updates for minor version.

It will then update to the latest minor version with the latest patch.

## Checklist

- [ ] Update tests.


zugao (approved):
LGTM

pkg/maintenance/helm/helmpatcher_test.go
```diff
@@ -458,3 +458,162 @@ func getResult(n, s, ct string) Result {
...
+				getResult("6.8", "active", "image"),
+				getResult("7.0", "active", "image"),
+				getResult("7.1.4", "active", "image"),
+				getResult("7.2.4", "inactive", "image"),
+				getResult("7.1.12-alpha", "active", "image"),
+				getResult("7.3.11", "active", "image"),
+				getResult("7.0.14-alpha", "active", "image"),
+				getResult("7.0.14%alpa", "active", "image"),
+			},
+			expectedVer: "7.3.11",
```
zugao, line 546: hmm I think you duplicated here the test. it should stay the same old version?

Kidswiss: You're right, it should actually return the same string as in the values.

Probably a copy paste error from when the first strategy was implemented, I've now adjusted the test so that there's actually really no newer version in the list.

PR #353, opened 2025-05-07
disabling majorVersionUpgrades

wejdross:
## Summary

* We need to disable MajorVersion Upgrade feature, as it's not yet production ready

## Checklist

- [x] Update tests.


Kidswiss:
AFAIK we used to have a webhook that prevented the user from changing the pg version after it was created. We should also re-add that one as well or they are able to break the instances...

Also we need to add a note to the docs that this feature is currently disabled.

Kidswiss (approved):
Should be fine until stackgres delivers the new features.

PR #354, opened 2025-05-12
Forgejo/Codey: Set compute resources according to plans

ThisIsntTheWay:
## Summary

* Forgejo and Codey instances will now use compute resources according to their selected plans (CPU, Memory limits and PVC size).  
* Fix a bug in the webhooks Codey FQDN check logic where the (temporary) ACME solver ingress was triggering false positives.
  * The ACME solver ingress uses the same hostname as the wildcard/let's encrypt ingress.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/functions/vshnforgejo/deploy.go
+++ b/pkg/comp-functions/functions/vshnforgejo/deploy.go
@@ -201,10 +201,14 @@ func addForgejo(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.V
 	svc.Log.Info("Got plan", "plan", plan)
 
 	// If this is a codey abstraction, we need to refer to codey plans
-	svc.Log.Info("This comps annotations", "annotations", comp.Annotations)
 	if comp.Annotations["metadata.appcat.vshn.io/abstraction"] == "codey" {
-		svc.Log.Info("This is a codey abstraction, using codey plans")
-		svc.Config.Data["plans"] = comp.Annotations["metadata.appcat.vshn.io/codeyplans"]
+		v, ok := comp.Annotations["metadata.appcat.vshn.io/codeyplans"]
+		if !ok {
+			return fmt.Errorf("this is a codey abstraction but has no codey plans")
+		}
+
+		svc.Log.Info("This is a codey abstraction, using codey plans", "codeyplans", v)
+		svc.Config.Data["plans"] = v
 	}
 
 	resources, err := utils.FetchPlansFromConfig(ctx, svc, plan)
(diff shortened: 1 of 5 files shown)


pkg/comp-functions/functions/vshnforgejo/deploy.go
```diff
@@ -189,12 +195,43 @@ func addForgejo(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.V
...
+	// Resources
+	svc.Log.Info("Fetching and setting resources")
+	plan := comp.Spec.Parameters.Size.GetPlan(svc.Config.Data["defaultPlan"])
+	svc.Log.Info("Got plan", "plan", plan)
+
+	// If this is a codey abstraction, we need to refer to codey plans
+	svc.Log.Info("This comps annotations", "annotations", comp.Annotations)
+	if comp.Annotations["metadata.appcat.vshn.io/abstraction"] == "codey" {
+		svc.Log.Info("This is a codey abstraction, using codey plans")
+		svc.Config.Data["plans"] = comp.Annotations["metadata.appcat.vshn.io/codeyplans"]
```
ThisIsntTheWay, line 207: This might be a point of discussion;  
One possible alternative to this would be to ingest the codey plans CM (doesn't exist yet but would be a small change), but this would require the use of a k8s client.  
I felt that this would be kind of overkill (on each reconciliation construct a client and then fetch CM) and doing it this way would be more efficient.

Kidswiss: > but this would require the use of a k8s client.

This will open a whole new can of worms :) Basically for this reason mentioned in the contribution guide:
> A Function MUST NOT assume it is deployed in any particular way, for example that it is running as a Kubernetes Pod in the same cluster as Crossplane.

https://github.com/crossplane/crossplane/blob/main/contributing/specifications/functions.md

But what if we invert the logic? Instead of teaching vshnforgejo all the codey plans, we could just make codey set the resources according to a plan. This would completely decouple that part.

PnT compositions support so called map transforms: https://docs.crossplane.io/latest/guides/function-patch-and-transform/#map-transforms

You could use that to map `mini` and `micro` plan to respective sizes for the vshnforgejo composite. It might even be possible to patch the whole size object in one go.

ThisIsntTheWay: This has now been superseded by the suggestion above:  
Forgejo no longer "knows" about Codey and the Codey composition now uses PnTs map transform funcs instead to dynamically set `parameters.size.<field>` based on the chosen plan.  
:point_right: https://github.com/vshn/component-appcat/pull/716/commits/998f060132def72149e6bb5fc52321cf9eabec5d#diff-336cf350699936f0e090a6ba0abc4d920b41e133a11c6e6625ac27dc79bf1555

PR #356, opened 2025-05-14
Helm Chart updates May 2025

Kidswiss:
## Summary

- With the latest version of the keycloak chart, the port definition isn't necessary anymore, as it's already in the chart and will generate a duplicate error while applying.
- The latest Redis chart seems to work fine with older versions as well, so we will allow a broader selection of open source Redis versions

## Checklist

- [x] Update tests.


mdnix (approved):
LGTM

PR #357, opened 2025-05-19
Add missing roles for APIServer

TheBigLee:
## Summary

* The APIServer needs additional permissions to get the kubeconfig on
a non-converged setup. It needs to read providerconfigs.kubernetes as
well as secrets.

## Checklist

- [ ] Update tests.

### DIFF
--- a/config/apiserver/role.yaml
+++ b/config/apiserver/role.yaml
@@ -26,6 +26,14 @@ rules:
   - update
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - admissionregistration.k8s.io
   resources:
   - mutatingwebhookconfigurations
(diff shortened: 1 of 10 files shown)


zugao (approved):
LGTM

config/apiserver/role.yaml
```diff
@@ -82,3 +82,19 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
Kidswiss, line 85: `make generate` will kill this again. You need to add magic comments for the generator:

```
// +kubebuilder:rbac:groups="kubernetes.crossplane.io",resources= providerconfigs,verbs=get;list;watch
```

Kidswiss (changes requested):

config/apiserver/role.yaml
TheBigLee: Oh good point. Fixed

PR #360, opened 2025-05-21
Make AppCat realtime compatible

Kidswiss:
## Summary

Crossplane will soon make the realtime mode the default. See https://github.com/crossplane/crossplane/issues/6453

The way we serializse the `Object`s leaves a `creationDate: null` in the inner manifests.

However, this will cause a reconcile loop when the realtime mode is enabled. While not apparante there might also be a reduction in reconciles with Crossplane in poll mode.

In either case, the values that get removed by this commit are garbage and can cause issues down the line.

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -438,6 +438,12 @@ func (s *ServiceRuntime) SetDesiredComposedResourceWithName(obj xpresource.Manag
 		return err
 	}
 
+	// We need to remove some fields to make realtime compositions work.
+	// CreationTimestamp is set to nil due to our marshalling, but this will cause the object to go into an endless loop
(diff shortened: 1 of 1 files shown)


zugao (commented):
Why not set the `CreationTimestamp` instead?

Kidswiss:
> Why not set the `CreationTimestamp` instead?

CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. **Clients may not set this value.** It is represented in RFC3339 form and is in UTC. **Populated by the system. Read-only.** Null for lists. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata

ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#objectmeta-v1-meta

PR #362, opened 2025-05-21
Fix wrong mailgun secret ref for patching secret

TheBigLee:
## Summary

* This PR fixes a bug we introduced by referencing the correct secret for the mailgun credentials.

## Checklist

- [ ] Update tests.


zugao (approved):
This should resolve the immediate problem but it will probably fall short for split control plane

TheBigLee:
> This should resolve the immediate problem but it will probably fall short for split control plane

Yes. This has to be properly implemented in a follow-up.

PR #363, opened 2025-05-21
Move Redis from P&T to comp-functions

mdnix:
## Summary

* Move logic from P&T to comp-functions for redis
* Use generic service in component-appcat

## Checklist

- [x] Update tests.


pkg/comp-functions/functions/vshnredis/redis_deploy.go
```diff
@@ -5,27 +5,283 @@ import (
...
+		"architecture":     "standalone",
+
+		"global": map[string]any{
+			"security": map[string]any{
+				"allowInsecureImages": true,
+			},
+		},
+
+		"image": map[string]any{
+			"repository": "bitnami/redis",
```
zugao, line 137: this has to go into defaults of our component-appcat

pkg/comp-functions/functions/vshnredis/redis_deploy.go
```diff
@@ -5,27 +5,283 @@ import (
...
+			"tag":        comp.Spec.Parameters.Service.Version,
+		},
+
+		"auth": map[string]any{
+			"enabled":                   true,
+			"existingSecret":            secret,
+			"existingSecretPasswordKey": passwordKey,
+		},
+
+		"tls": map[string]any{
```
zugao, line 147: I am not sure whether TLS is mandatory all the time, isn't it configurable in the claim?

pkg/comp-functions/functions/common/tls.go
```diff
@@ -230,3 +230,273 @@ func CreateTLSCerts(ctx context.Context, ns string, serviceName string, svc *run
...
+			RenewBefore: &metav1.Duration{
+				Duration: time.Duration(2400 * time.Hour),
+			},
+			Subject: &cmv1.X509Subject{
+				Organizations: []string{
+					"vshn-appcat-client",
+				},
+			},
+			IsCA: false,
+			//DNSNames: []string{
```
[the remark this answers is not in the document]
Kidswiss: AFAIK the client cert need at least some name. It doesn't have to resolve though. It could probably be just a value like `client`. Needs testing though.

pkg/comp-functions/functions/common/tls.go
```diff
@@ -230,3 +230,273 @@ func CreateTLSCerts(ctx context.Context, ns string, serviceName string, svc *run
...
+				},
+			},
+			IsCA: false,
+			PrivateKey: &cmv1.CertificatePrivateKey{
+				Algorithm: cmv1.RSAKeyAlgorithm,
+				Encoding:  cmv1.PKCS1,
+				Size:      4096,
+			},
+			Usages: []cmv1.KeyUsage{"server auth", "client auth"},
+			//DNSNames: []string{
```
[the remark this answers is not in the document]
mdnix: I temporarily commented this part out because in the case of redis, we initially created the certificates using these two DNSNames:
[https://github.com/vshn/component-appcat/blob/master/component/vshn\_redis.jsonnet#L432](https://github.com/vshn/component-appcat/blob/master/component/vshn_redis.jsonnet#L432)

When I added the `CreateMTLSCerts` helper function, my intention was for it to be reusable across different services in the future. That’s why it includes:

```go
DNSNames: []string{
	serviceName + "." + ns + ".svc.cluster.local",
	serviceName + "." + ns + ".svc",
},
```

This approach is similar to `CreateTLSCerts` for our other existing services.

However, the issue here with Redis: using `CreateMTLSCerts` along with additional SANs (and those provided by the function) would result in new certificate requests, potentially breaking existing Redis services. To avoid this during testing, I simply commented it out and passed the TLS options manually when creating a new Redis instance via the component functions:

```go
tlsOpts := &common.TLSOptions{
	AdditionalSans: []string{
		fmt.Sprintf("redis-headless.vshn-redis-%s.svc.cluster.local", comp.GetName()),
		fmt.Sprintf("redis-headless.vshn-redis-%s.svc", comp.GetName()),
	},
}
```

One idea I had was to modify `CreateMTLSCerts` so that if `AdditionalSans` is specified via `TLSOptions`, it would only use those DNS names. If `AdditionalSans` isn’t provided, it would fall back to the defaults.

pkg/comp-functions/functions/vshnredis/redis_deploy.go
mdnix: I would go even further and say I can simply omit this. The imageRegistry and tag will be handled similarly to how it's done for MariaDB.

pkg/comp-functions/functions/vshnredis/redis_deploy.go
mdnix: You're right, added it back in

zugao (approved):
LGTM, during tests try to see what is the impact to the existing instances. Will they be wiped out, restarted, blocked or nothing will happen?

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnredis/redis_url_test.go
```diff
@@ -1,6 +1,6 @@
 package vshnredis
 
-import (
+/* import (
```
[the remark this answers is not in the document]
mdnix: Removed it since I consolidated everything in deploy

TheBigLee (approved):
LGTM

PR #364, opened 2025-05-22
Allow deletion of Appcats, if claim namespace is deleted

wejdross:
## Summary

* If claim namespace has deletionTimestamp, ignore deletion protection of appcats and clean resources
* Lack of this feature caused us a lot of ops overhead, as it's unintuitive for customers
* impovement: faster testing of Appcats

## Checklist

- [x] Update tests.

### DIFF
--- a/pkg/controller/webhooks/default_handler.go
+++ b/pkg/controller/webhooks/default_handler.go
@@ -8,6 +8,7 @@ import (
 	"github.com/vshn/appcat/v4/pkg/common/quotas"
 	"github.com/vshn/appcat/v4/pkg/common/utils"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/functions/common"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -124,6 +125,17 @@ func (r *DefaultWebhookHandler) ValidateDelete(ctx context.Context, obj runtime.
 	comp, ok := obj.(common.Composite)
 	if !ok {
 		return nil, fmt.Errorf("provided manifest is not a valid " + r.gk.Kind + " object")
+	}
+
+	// get details of namespace
+	namespace := corev1.Namespace{}
+	err := r.client.Get(ctx, client.ObjectKey{Name: comp.GetNamespace()}, &namespace)
+	if err != nil {
+		return nil, apierrors.NewInternalError(fmt.Errorf("Error getting namespace: " + err.Error()))
+	}
+
+	if namespace.DeletionTimestamp != nil {
+		return nil, nil
 	}
 
 	allErrs = GetClaimDeletionProtection(comp.GetSecurity(), allErrs)


Kidswiss (approved):
LGTM

The only gripe I have with this is, that the deletion protection gets weakened quite drastically.

We should definitely document that the deletion protection doesn't protect against deletions of the claim namespace.

mdnix (approved):
LGTM, and I agree with @Kidswiss. Since this change modifies behavior it’s important that we document it properly

lieneluksika:
Will not be merged because of differing opinions. We will have an ADR to list all the possible options and make an informed decision.

TheBigLee:
I would also not go with the finalizer approach. As @Kidswiss mentioned, the namespace is already doomed and not recoverable anymore.
I'm also for the webhook solution and agree that we need to carefully design this webhook. It should only trigger for claim namespaces (TBD how to implement that exactly) and then check if that namespace has any instances that have `deletionProtection` enabled and block the deletion in that case.

We should also keep in mind, that on OCP the end-users usually don't interact with namespaces directly but with `projects`. We should also evaluate how OCP handles the deletion of a `project` if the namespace can't be deleted. In the worst case, we end up with `projects` in an infinite termination state and users ending up removing the finalizers on the projects, leaving orphan namespaces behind.

zugao:
> Will not be merged because of differing opinions. We will have an ADR to list all the possible options and make an informed decision.

Having an ADR for that is an overkill in my opinion. We probably already have a consensus here, and it is a loose webhook. We just need an implementation ticket and somebody to refine it after a short talk with the colleagues.

> I would not go the finalizer route, that is a reactive measure and the namespace is doomed already. Also with another finalizer we have the same situation as now, where a customer can break the Crossplane and potentially the cluster by removing it.

I agree, the finalizer is not a good option. I mentioned it along the webhook only because I wanted to emphasize the fact that we have alternatives to the solution proposed in this PR.

> I discussed this with @mdnix. If we go that route it will be quite the rabbit hole. We need to be very careful with that webhook and configure it in a way that we don't impact the cluster stability if it's not available for some reason.

I agree it can be disruptive, but we don't need to make it bulletproof (block all namespaces, block claim namespace when webhook is not available). As you mentioned in your requirement list we could make it quite loose and still protect our claim namespace.

> The problem with the current approach is: the customer can break the system pretty badly by removing finalizers from the claim namespace. We can get phantom claims that are still in the system which makes Crossplane freak out. This could then potentially crash the K8s API server due to requests spam.

I absolutely understand that we need have a problem and it must be solved, I am just saying that we might want to have a better alternative to what we have here.

> We should also keep in mind, that on OCP the end-users usually don't interact with namespaces directly but with projects. We should also evaluate how OCP handles the deletion of a project if the namespace can't be deleted. In the worst case, we end up with projects in an infinite termination state and users ending up removing the finalizers on the projects, leaving orphan namespaces behind.

You raised a good point with OCP projects. Whoever is gonna work on that ticket, if we decide to go with webhooks then this has to be checked thoroughly.

Kidswiss:
> Having an ADR for that is an overkill in my opinion. We probably already have a consensus here, and it is a loose webhook. We just need an implementation ticket and somebody to refine it after a short talk with the colleagues.

I disagree, ADRs are also intended to document the AppCat and how we arrive at decisions. So this absolutely has to be one. Even if it's just for documentation.

> We should also keep in mind, that on OCP the end-users usually don't interact with namespaces directly but with projects. We should also evaluate how OCP handles the deletion of a project if the namespace can't be deleted. In the worst case, we end up with projects in an infinite termination state and users ending up removing the finalizers on the projects, leaving orphan namespaces behind.

We might want to extend the protection to projects then as well. So that we don't even end up with projects in a potentially broken state.

PR #366, opened 2025-05-28
Permanently disable Envoy

Kidswiss:
## Summary

As of StackGres 1.16.0 it's possible to disable Envoy.

It sits in the direct data path of PostgreSQL so the performance of the database is impacted by Envoy. However the only benefit that it brings is some additional connection metrics, which are not worth the additional resource usage and the performance impact.
## Checklist

- [x] Update tests.


mdnix (approved):
LGTM

zugao (approved):
This will restart postgres

TheBigLee (approved):

PR #374, opened 2025-06-05
Revert back to try'n error approach for maintenance

Kidswiss:
## Summary

Previously the version update logic for each instances was changed to check for an empty namespace and decide that way if it has to patch the claim or the composite. This was an "improvement" over the previous try'n error approach, as it made things more explicit.

However, provider-kubernetes is currently not able to remove existing fields without switching to server-side-apply:
https://github.com/crossplane-contrib/provider-kubernetes/issues/242#issuecomment-2577579544

So this reverts the logic back to first try for a claim and the fall back to the composite.

## Checklist

- [x] Update tests.


zugao (approved):
LGTM

PR #392, opened 2025-06-20
Add/schema grants

Kidswiss:
## Summary

## Checklist

- [ ] Update tests.


pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -1447,9 +1447,6 @@ func (s *ServiceRuntime) deployConnectionDetailsToInstanceNS() error {
 
 		compName := s.desiredResources[i].Resource.GetName()
 
-		// prefix the name in the connetiondetailsref
-		// and set to crossplane namespace
-		cdRef.Name = compName + "-" + cdRef.Name
```
zugao, line 1452: happy this is not necessary anymore

apis/sql/postgresql/v1alpha1/grant_types.go
```diff
@@ -20,13 +20,19 @@ import (
 	"context"
```
zugao, line 20: This file with all its contents will be injected into provider sql and as I understand more GRANTS will be issued is that right?

pkg/comp-functions/functions/vshnpostgres/user_management.go
```diff
@@ -128,6 +128,12 @@ func addConnectionDetail(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntim
 		svc.AddResult(runtime.NewWarningResult(fmt.Sprintf("cannot get userpassword from secret: %s", err)))
 	}
 
+	// Userpass is not yet ready, so we'll check again in 30 seconds
+	if len(userpassCD) == 0 {
```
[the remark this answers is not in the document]
Kidswiss: Not necessarily, the change in the runtime solves the race. This just makes things snappier.

apis/sql/postgresql/v1alpha1/grant_types.go
Kidswiss: This is copy pasted direclty from the provider-sql. We import it to be able to create valid CRs for the provider.

Kidswiss: This is them being added :D

PR #394, opened 2025-06-23
Add `envFrom` field to VSHNKeycloak

ThisIsntTheWay:
## Summary

* Adds a new field `envFrom` with which users can add env variables stemming from ConfigMaps or Secrets to a Keycloak instance.
* The field `customEnvVariablesRef` is marked as deprecated and will emit a warning if supplied by the user, but is otherwise kept for backwards compatibility reasons.

Also fixes a bug with the `apply-cfg` job:
* On new deployments, having the job run immediately would result in parallel keycloak setups, both of which seed the DB with initial data simultaneously and causing collisions.  
The result is that both Keycloak processes abort the setup, and both pods restart just to begin the cycle anew.

## Checklist

- [X] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/781


pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -700,6 +701,40 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
...
 	}
 
+	if comp.Spec.Parameters.Service.CustomEnvVariablesRef != nil {
+		envFrom, err := toYAML(comp.Spec.Parameters.Service.CustomEnvVariablesRef)
+		if err != nil {
+			return nil, err
+		}
+
+		values["extraEnvFrom"] = envFrom
+	}
```
TheBigLee, line 711: This should probably be removed as it's not really needed and will be overridden further down

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -356,6 +365,26 @@ func buildConfigApplyJob(comp *vshnv1.VSHNKeycloak, adminSecret, jobName string)
 					RestartPolicy:    corev1.RestartPolicyOnFailure,
 					ImagePullSecrets: []corev1.LocalObjectReference{{Name: pullsecretName}},
 					Volumes:          volumes,
+					// We need this init container as the apply-config container could otherwise start a parallel keycloak setup
```
zugao, line 368: Is there a reason to create this job during provisioning? As I understood this initContainer exists only to make sure during provisioning there are no parallel keycloak setup, right?

Kidswiss (approved):

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -343,6 +389,12 @@ func buildConfigApplyJob(comp *vshnv1.VSHNKeycloak, adminSecret, jobName string)
 		})
 	}
 
+	if comp.Spec.Parameters.Service.EnvFrom != nil {
+		envFrom = append(envFrom, *comp.Spec.Parameters.Service.EnvFrom...)
+	}
+
+	KeycloakApiUrl := fmt.Sprintf("http://%s-%s.%s.svc.cluster.local", comp.GetName(), serviceSuffix, comp.GetInstanceNamespace())
```
mdnix, line 396: lowercase

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -235,12 +236,47 @@ func handleCustomConfig(ctx context.Context, comp *vshnv1.VSHNKeycloak, svc *run
 	lastConfigHash := comp.GetLastConfigHash()
 	lastEnvHash := comp.GetLastEnvHash()
 
+	// If we just created the instance, then the last hashes will be empty and thus trigger configOrEnvChanged().
+	// This will result in a race condition where two keycloak setups could run in parallel and clash with each other.
+	// Therefore, we'll check if those hashes are empty and set them to the current hashes.
+	if (lastConfigHash == "" && currentConfigHash != "") || (lastEnvHash == "" && currentEnvHash != "") {
```
mdnix, line 242: Did you test this change for the config apply job in Lab? If not I can help you out, I have all the dummy configs at my fingertips

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -719,6 +790,46 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
...
+				obj := &corev1.ConfigMap{}
+				_, err := svc.CopyKubeResource(ctx, obj, comp.GetName()+"-env-cm-"+r.ConfigMapRef.Name, r.ConfigMapRef.Name, comp.GetClaimNamespace(), comp.GetInstanceNamespace())
+				if err != nil {
+					return nil, fmt.Errorf("cannot copy Keycloak env variable ConfigMap to instance namespace: %w", err)
+				}
+			}
+		}
+	}
+
+	if len(comp.Spec.Parameters.Service.CustomMounts) > 0 {
```
mdnix, line 812: Isn't this a duplicate? We already handle copying for this in `handleCustomMounts`

pkg/comp-functions/functions/vshnkeycloak/deploy.go
zugao: That works for me.

mdnix (approved):

PR #400, opened 2025-07-01
Release appcat

zugao:
## Summary

* This is a release PR

Component PR: https://github.com/vshn/component-appcat/pull/801


Kidswiss:
For the future we should probably disable the automagic creation of a PR in the component, when we do the PR to master.

Kidswiss:
@zugao we might want to merge up to a specific commit from `develop` as we've discussed last week. To avoid having something in here that's not reflected in the component.

zugao:
> @zugao we might want to merge up to a specific commit from `develop` as we've discussed last week. To avoid having something in here that's not reflected in the component.

Yes but in this case we use the latest as there are no gaps at the moment. Commit dde0d50f860bb20b1774d2fb818a5668f1bf94ed from develop has to be merged into master

Kidswiss (approved):

PR #414, opened 2025-07-09
Make deletion protection not rely on Crossplane specifics

Kidswiss:
## Summary

We used the `crossplane.io/composite` annotation previously to determine the name of the owning composite. However, about 7 months ago Crossplane changed their logic and they always remove the annotation prior to doing the actual deletion call.

This caused our deletion protection to enter a failsafe path and allow deletion.

This commit aims to make the deletion protection more robust and not rely on any Crossplane specifics.

## Checklist

- [ ] Update tests.


pkg/controller/webhooks/deletionprotection.go
```diff
@@ -50,26 +51,29 @@ func checkManagedObject(ctx context.Context, obj client.Object, c client.Client,
 
 	ownerKind, ok := obj.GetLabels()[runtime.OwnerKindAnnotation]
 	if !ok || ownerKind == "" {
-		l.Info(runtime.OwnerKindAnnotation + " label not set, skipping evaluation")
-		return compositeInfo{Exists: false}, nil
+		l.Info(runtime.OwnerKindAnnotation + " label not set, blocking deletion")
+		return compositeInfo{Exists: isDeletionProtected(obj), Reason: fmt.Sprintf("%s label not set, blocking deletion", runtime.OwnerKindAnnotation)}, nil
```
zugao, line 55: Didn't we say we block here the deletion?

Kidswiss: It gets blocked.

`isDeletionProtected(obj)` checks if the override label is set or not. This way we can get things unstuck if for any reason the labels get messed up completely.

PR #425, opened 2025-07-17
Ensure Keycloak Config Apply Job Executes on Initial Install

mdnix:
## Summary

* PR https://github.com/vshn/appcat/pull/394 introduced a change in behaviour when the Config Apply Job should run
* This change corrects the logic so that if a user defines  customConfigurationRef  and/or customEnvVariablesRef during the initial installation, the Config Apply Job will execute accordingly. 

Component PR: https://github.com/vshn/component-appcat/pull/834

### DIFF
--- a/apis/vshn/v1/dbaas_vshn_keycloak.go
+++ b/apis/vshn/v1/dbaas_vshn_keycloak.go
@@ -136,11 +136,15 @@ type VSHNKeycloakServiceSpec struct {
 	// The referenced configmap, must have the configuration in a field called `keycloak-config.json`
 	CustomConfigurationRef *string `json:"customConfigurationRef,omitempty"`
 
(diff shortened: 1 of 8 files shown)


ThisIsntTheWay:
Is this supposed to merge into `master`?  
Right now, the diff doesn't really align with the PR description.

mdnix:
> Is this supposed to merge into `master`? Right now, the diff doesn't really align with the PR description.

Yep you’re right, it is misleading. But my thought was to fix how the Job runs on first install (as we talked about yesterday), but I also need to rollout your `envFrom` feature, which will fix the other issue where the envs defined in `customEnvVariablesRef` keep getting added to the Statefulset.

Kidswiss (changes requested):

PR #431, opened 2025-07-23
Fix race conditions due to desiredComposite

Kidswiss:
## Summary

The desiredComposite is not always initialized and can sometimes be empty. So any logic that relies on it to get the state of the observed composite might actually work on wrong values.

This commit fixes this issue.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/841

### DIFF
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -701,12 +701,7 @@ func (s *ServiceRuntime) putIntoObject(o client.Object, kon, resourceName string
 
 // GetObservedComposite returns the observed composite and unmarshals it into the given object.
 func (s *ServiceRuntime) GetObservedComposite(obj client.Object) error {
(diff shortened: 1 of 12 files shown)


pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -701,12 +701,7 @@ func (s *ServiceRuntime) putIntoObject(o client.Object, kon, resourceName string
 
 // GetObservedComposite returns the observed composite and unmarshals it into the given object.
 func (s *ServiceRuntime) GetObservedComposite(obj client.Object) error {
-	comp, err := request.GetObservedCompositeResource(s.req)
```
Kidswiss, line 704: This is a small optimization. During the initialization of the runtime, the observed composite is copied over to `observedComposite`. So no need to get it from the original request every time.

PR #452, opened 2025-08-15
Removed omitempty from default=true boolean fields

mikeshootzz:
## Summary

Fixes an issue where boolean fields with omitempty and default=true could not be set to false via YAML. The YAML serializer would omit false values due to the omitempty tag, causing them to be interpreted as the default true value instead.

## Checklist

Component PR: https://github.com/vshn/component-appcat/pull/879

### DIFF
--- a/apis/vshn/v1/common_types.go
+++ b/apis/vshn/v1/common_types.go
@@ -196,7 +196,8 @@ type Security struct {
 
 	// DeletionProtection blocks the deletion of the instance if it is enabled (enabled by default)
 	// +kubebuilder:default=true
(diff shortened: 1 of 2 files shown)


mikeshootzz:
> Won't the
> 
> ```
> required:
>   - deletionProtection
> ```
> 
> cause issues, or will this automatically be set to true due to `// +kubebuilder:default=true`?

The +kubebuilder:default=true runs before the required validation, so omitted fields get defaulted to true first.

zugao:
Also you have to run `make generate`

Kidswiss (approved):

PR #456, opened 2025-08-22
Sequencing for the user management of postgresql

mikeshootzz:
## Summary

- Grants and databases now only get created once the role resource is ready. This prevents long startup times.

## Checklist

- [ ] Update tests.


pkg/comp-functions/functions/vshnpostgres/user_management.go
```diff
@@ -340,3 +353,37 @@ func addGrants(comp common.Composite, svc *runtime.ServiceRuntime, username, dbn
 		svc.Log.Error(err, "cannot apply database")
 	}
 }
+
+// This is adapted from the private method in function_mgr.go
```
Kidswiss, line 357: It'd fine if you just make the function public in the Runtime.

pkg/comp-functions/functions/vshnpostgres/user_management.go
```diff
@@ -38,17 +40,28 @@ func UserManagement(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runti
...
+		} else {
+			svc.SetDesiredResourceReadiness(dbResourceName, runtime.ResourceUnReady)
+		}
 
-		addGrants(comp, svc, dbuser, dbname, access.Privileges)
+		grantResourceName := fmt.Sprintf("%s-%s-%s-grants", comp.GetName(), dbuser, dbname)
+		if isResourceSyncedAndReady(svc, dbResourceName) {
+			addGrants(comp, svc, dbuser, dbname, access.Privileges)
+		} else {
+			svc.SetDesiredResourceReadiness(grantResourceName, runtime.ResourceUnReady)
```
Kidswiss, line 63: There might be an edgecase where this doesn't work. If the resource doesn't exist at all currently it will not be able to set it to unready.

It makes more sense to set the composite to unready.

pkg/comp-functions/functions/vshnpostgres/user_management.go
```diff
@@ -38,17 +38,28 @@ func UserManagement(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runti
...
 
-		addDatabase(comp, svc, dbuser, dbname)
+		userPasswordRef := addUser(comp, svc, *access.User)
+		roleResourceName := fmt.Sprintf("%s-%s-role", comp.GetName(), dbuser)
+
+		dbResourceName := fmt.Sprintf("%s-%s-database", comp.GetName(), dbname)
+		if svc.IsResourceSyncedAndReady(roleResourceName) {
+			addDatabase(comp, svc, dbuser, dbname)
+		} else {
+			return runtime.NewWarningResult(fmt.Sprintf("Role %s not ready, waiting before creating database %s", roleResourceName, dbResourceName))
```
Kidswiss, line 54: It's better to just log here and also add an event to the the state. Instead of returning, as it could drop stuff from the desired state.

Kidswiss, line 54: It's better to just log here and also add an event to the the state. Instead of returning, as it could drop stuff from the desired state.

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnpostgres/user_management.go
```diff
@@ -38,17 +38,32 @@ func UserManagement(ctx context.Context, comp *vshnv1.VSHNPostgreSQL, svc *runti
...
-		addDatabase(comp, svc, dbuser, dbname)
+		userPasswordRef := addUser(comp, svc, *access.User)
+		roleResourceName := fmt.Sprintf("%s-%s-role", comp.GetName(), dbuser)
+
+		dbResourceName := fmt.Sprintf("%s-%s-database", comp.GetName(), dbname)
+		if svc.IsResourceSyncedAndReady(roleResourceName) {
+			addDatabase(comp, svc, dbuser, dbname)
+		} else {
+			svc.Log.Info(fmt.Sprintf("Role %s not ready, waiting before creating database %s", roleResourceName, dbResourceName))
+			comp.SetConditions(xpv1.Unavailable().WithMessage(fmt.Sprintf("Role %s not ready", roleResourceName)))
```
Kidswiss, line 55: you can use `svc.AddResult` crossplane will then do the rest.

Kidswiss (changes requested):

PR #469, opened 2025-09-01
Make hotfixer more generic

Kidswiss:
## Summary

With this commit the hotfixer will now container following functionaltiy:

- Ability to specify the ServiceID label
- Ability to only update a specific ServiceID

## Checklist

- [ ] Update tests.


mdnix (approved):

cmd/hotfixreleaser.go
```diff
@@ -71,11 +94,16 @@ func (h *hotfixer) runHotfixer(cmd *cobra.Command, _ []string) error {
 			continue
 		}
 
+		// let's skip if the the serviceID is not empty and matches the service
+		if serviceID != "" && xrd.GetLabels()[serviceIDLabel] != serviceID {
```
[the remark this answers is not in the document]
Kidswiss: I used it before I had the filter. I'll remove it again.

cmd/hotfixreleaser.go
```diff
@@ -71,11 +94,16 @@ func (h *hotfixer) runHotfixer(cmd *cobra.Command, _ []string) error {
...
+			continue
+		}
+
 		p := fieldpath.Pave(xrd.Object)
 
 		XRDLabels := xrd.GetLabels()
 
-		if XRDLabels[release.ServiceIDLabel] == "" {
+		if XRDLabels[serviceIDLabel] == "" {
```
mdnix, line 106: Should we really abort here when there is one not labeled?

mdnix (approved):
Overall looking good. Remove the condition Gabi mentioned

cmd/hotfixreleaser.go
Kidswiss: I'd argue for that yes. It's something we have to wire-up via the component when we generate the XRDs. So if we deploy a release that has a broken XRD label, something might be very not okay.

PR #474, opened 2025-09-05
Add nextcloud scc

zugao:
## Summary

* Create service account via nextcloud release custom resource
* Create cluster role binding that binds the above service account to a cluster role from component which is linked to an scc
* SCC and cluster role are defined in component.

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/907


zugao:
> Please test on lab and maybe on a cluster where we know that the SCCs are missing.

Tested on lab successfully. There are no clusters where SCC is missing. After the release all nextcloud instances will be restarted and will use a new SCC defined in the [component](https://github.com/vshn/component-appcat/pull/907).

PR #496, opened 2025-10-01
Label for removing the version pinning

mikeshootzz:
## Summary

## Checklist

- [ ] Update tests.


apis/vshn/v1/dbaas_vshn_forgejo.go
```diff
@@ -165,6 +165,8 @@ type VSHNForgejoStatus struct {
 	// Schedules keeps track of random generated schedules, is overwriten by
 	// schedules set in the service's spec.
 	Schedules VSHNScheduleStatus `json:"schedules,omitempty"`
+	// InitialMaintenanceRan tracks if the initial maintenance job has been triggered
```
Kidswiss: Just an idea: instead of a bool we could make it a `*string` or `*time.Time` where we record the time it ran. Gives a bit more context than a simple bool. And it being a pointer will make it easy to check if it's set or not.

pkg/comp-functions/functions/common/maintenance/maintenance_test.go
```diff
@@ -177,6 +177,68 @@ func TestAddMaintenanceJob(t *testing.T) {
...
+			fileName:                  "vshn-postgres/maintenance/01-GivenSchedule.yaml",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.TODO()
+
+			svc := commontest.LoadRuntimeFromFile(t, tt.fileName)
+
+			comp := &vshnv1.VSHNPostgreSQL{}
```
Kidswiss: PostgreSQL is the only non-helm based service. We should probably also test with a helm based one.

Kidswiss (changes requested):

apis/vshn/v1/dbaas_vshn_forgejo.go
mikeshootzz: good point, implemented this

zugao:
Please add in the description the problem you are facing. Why do we need a status for the first maintenance?

zugao (approved):
I looks good overall but make sure you do a test that current deployed instances are properly managed when this new feature is released. Randomly running maintenance after release will trigger restarts.

And on another note. As already discussed in a meeting. Probably this whole feature might have been better implemented in a separate controller (besides billing) that would exclusively update the claim/composite on demand. This is not the first time we are blocked by the fact that we cannot update the claim/composite in the comp-function. And last we agreed the maintenance is not the best way to achieve what is being proposed in this ticket, because maintenance is smth you run periodically to update a service from the initial state. The initial state should be already up to date when the service is provisioned.

P.S.: I approved this PR but maybe we could take a step back abd investigate whether the controller would be a better choice, @Kidswiss what do you think? It would also go into the direction of having provider/operator Servala that would help to consolidate the code in the controller.

PR #598, opened 2026-02-25
Add SSH support to VSHNForgejo

mdnix:
## Summary

This feature adds the option to access VSHNForgejo instances via SSH using the Kubernetes Gateway API. When enabled, users can clone and push over SSH through a shared TCP gateway.

### How it works

1. User enables SSH by setting `spec.parameters.service.ssh.enabled: true` on their VSHNForgejo claim
2. Composition function creates the required Gateway API resources:
   - `XListenerSet`: requests a  TCP listener on the Gateway (port initially set to 0)
   - `TCPRoute`: routes traffic from the Gateway to the Forgejo SSH service
   - `ReferenceGrant`: authorizes the cross namespace backend reference (scoped to the specific SSH service only)
   - `NetworkPolicy`: allows ingress from the gateway namespace to the Forgejo pods
3. Mutating webhook intercepts `XListenerSet` CREATE requests and allocates a unique port from a configurable range (default 10000-10999). Port reservations use Kubernetes `Leases` for atomic allocation, preventing races across webhook replicas. 
4. On a subsequent reconcile the comp function reads back the observed `XListenerSet` and preserves the assigned port and gateway
5. The allocated port and SSH domain are written to the connection details (`FORGEJO_SSH_HOST` and `FORGEJO_SSH_PORT`) and the Helm release values get updated so the Forgejo instances displays the correct clone URL when choosing SSH.

### Gateway sharding

When `--ssh-gateway-capacity` is set, the webhook distributes `XListenerSets` across multiple gateways. It picks the gateway with the fewest listeners that still has capacity. If all gateways are full the admission request is denied.

### Disabled by default

The feature requires explicit configuration at both the platform and user level. Cluster operators must configure gateway names and namespace per cluster before any SSH resources can be created. Individual users must then set `spec.parameters.service.ssh.enabled: true` on their Forgejo claim. Without both, no SSH  resources are created.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1081


pkg/comp-functions/functions/vshnforgejo/ssh.go
```diff
@@ -0,0 +1,373 @@
...
+	xls := &unstructured.Unstructured{
+		Object: map[string]any{},
+	}
+	xls.SetAPIVersion("gateway.networking.x-k8s.io/v1alpha1")
+	xls.SetKind("XListenerSet")
+	xls.SetName(name)
+	xls.SetNamespace(namespace)
+
+	xls.Object["spec"] = map[string]any{
+		"parentRef": map[string]any{
```
Kidswiss: Why not use the same approach as in the webhook?

pkg/controller/webhooks/sshgateway/allocator.go
```diff
@@ -0,0 +1,212 @@
...
+	ns, _, _ := unstructured.NestedString(obj.Object, "spec", "parentRef", "namespace")
+	name, _, _ := unstructured.NestedString(obj.Object, "spec", "parentRef", "name")
+
+	return GatewayKey{
+		Namespace: ns,
+		Name:      name,
+	}
+}
+
+func toInt32(v any) int32 {
```
Kidswiss: Deja Vu

Kidswiss (changes requested):

pkg/controller/webhooks/sshgateway/allocator.go
mdnix: moved it to utils

pkg/comp-functions/functions/vshnforgejo/ssh.go
mdnix: yep you are right, changed it to also use unstructured

mikeshootzz:
### Code review

Found 2 issues:

1. `tryReclaimStaleLease` treats transient API errors as "holder exists", permanently leaking ports

When the XListenerSet lookup at line 124 returns a non-404 error (e.g. RBAC denial, API timeout), the function returns `false` at line 130, leaving the Lease permanently reserved. There is no retry mechanism and no background cleanup, so a transient error during reclaim can exhaust the entire port range without manual intervention.

https://github.com/vshn/appcat/blob/2138899a232a0cf044507b243795baeaec6a0cc6/pkg/controller/webhooks/sshgateway/allocator.go#L122-L135

2. `mutating-webhooks.yaml` uses kubebuilder scaffold placeholder name `mutating-webhook-configuration`, but `make webhook-debug` expects `appcat-sshgateway`

The Makefile `webhook-debug` target at line 232 explicitly looks up `appcat-sshgateway` to patch the mutating webhook for local development. The generated file uses the kubebuilder default name `mutating-webhook-configuration`, causing the debug target to silently skip patching the mutating webhook (`|| echo "... skipping"`). The validating webhook has the same pattern (`validating-webhook-configuration` vs `appcat-validation`), so both are affected if the two names are supposed to match.

https://github.com/vshn/appcat/blob/2138899a232a0cf044507b243795baeaec6a0cc6/config/controller/mutating-webhooks.yaml#L3-L13

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mikeshootzz (approved):

pkg/comp-functions/functions/vshnforgejo/ssh.go
```diff
@@ -0,0 +1,390 @@
...
+
+	mapping := map[string]string{}
+	if err := json.Unmarshal([]byte(raw), &mapping); err != nil {
+		svc.Log.Error(err, "failed to parse sshGateways")
+		return ""
+	}
+
+	for name := range mapping {
+		return name
+	}
```
TheBigLee, line 400: This loop is non-deterministic, because the map iteration order in Go is random. So this might change on every reconcile

pkg/comp-functions/functions/vshnforgejo/ssh.go
```diff
@@ -0,0 +1,390 @@
...
+
+	common.SetNestedObjectValue(values, []string{"gitea", "config", "server", "DISABLE_SSH"}, false)
+	common.SetNestedObjectValue(values, []string{"gitea", "config", "server", "START_SSH_SERVER"}, true)
+
+	if sshDomain != "" {
+		common.SetNestedObjectValue(values, []string{"gitea", "config", "server", "SSH_DOMAIN"}, sshDomain)
+	}
+	if allocatedPort > 0 {
+		common.SetNestedObjectValue(values, []string{"gitea", "config", "server", "SSH_PORT"}, int(allocatedPort))
+	}
```
TheBigLee, line 307: If I understand this correctly, the `allocatedPort` is not known in the first reconcile. So it will enable ssh and start the ssh server but without setting the `SSH_PORT`.
This if statement should therefore be set further up and return early if no `allocatedPort` is defined yet to avoid forgejo from starting the SSH server without the correct port.

pkg/comp-functions/functions/vshnforgejo/ssh.go
```diff
@@ -0,0 +1,390 @@
...
+		"name":      gatewayName,
+	}, "spec", "parentRef")
+
+	if err != nil {
+		return fmt.Errorf("setting parentRef: %w", err)
+	}
+
+	listener := map[string]any{
+		"name":     sshListenerName,
+		"port":     int64(port),
```
TheBigLee, line 124: Why are we mixing int64 and int32 with the ports?

pkg/comp-functions/functions/vshnforgejo/ssh.go
mdnix: good catch!

pkg/comp-functions/functions/vshnforgejo/ssh.go
mdnix: Yep, you're absolutely right! I moved the part where I enable the SSH parts in the helm release

pkg/comp-functions/functions/vshnforgejo/ssh.go
mdnix: I know it's confusing but it comes from three layers each picking a different numeric type: kubernetes typed API uses `int32` for all port fields, the unstructured API requires `int64` (passing `int32` actually panics in `runtime.DeepCopyJSONValue`), and `encoding/json` unmarshals numbers into `interface{}` as `float64` after any json round trip. Since `XListenerSet` and `TCPRoute` are constructed using `unstructured` (importing the typed Gateway API types would break the controller due to version incompatibilities), we inevitably cross all three type boundaries. `ToInt32` handles the conversion.

pkg/controller/webhooks/sshgateway/sharding.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// SelectGateway determines which gateway should host the new XListenerSet.
+// If the current gateway has room, it returns unchanged.
+// Otherwise it picks the gateway with the fewest listeners that still has capacity.
+// Returns an error if all gateways are full.
+func (gs *GatewaySharding) SelectGateway(currentRef GatewayKey, newListenerCount int, listenerCounts map[GatewayKey]int) (GatewayKey, bool, error) {
+	if listenerCounts[currentRef]+newListenerCount <= gs.capacity {
+		return currentRef, false, nil
+	}
```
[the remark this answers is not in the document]
mdnix: checking the existence of the ref now

cmd/controller.go
```diff
@@ -143,6 +156,15 @@ func (c *controller) executeController(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	if c.sshGateways != "" && c.sshGatewayNamespace != "" {
+		gatewayNames := strings.Split(c.sshGateways, ",")
+
```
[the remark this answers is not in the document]
mdnix: no need to do that, in component-appcat its passed properly as flag

Makefile
```diff
@@ -215,16 +218,26 @@ $(webhook_cert): $(webhook_key)
...
 	kubectl replace -f - && \
 	kubectl annotate validatingwebhookconfigurations.admissionregistration.k8s.io appcat-validation kubectl.kubernetes.io/last-applied-configuration-
+	@if kubectl get mutatingwebhookconfigurations.admissionregistration.k8s.io appcat-sshgateway >/dev/null 2>&1; then \
+		cabundle=$$(cat .work/webhook/tls.crt | base64) && \
+		kubectl annotate mutatingwebhookconfigurations.admissionregistration.k8s.io appcat-sshgateway cert-manager.io/inject-ca-from- 2>/dev/null || true && \
+		kubectl get mutatingwebhookconfigurations.admissionregistration.k8s.io appcat-sshgateway -oyaml | \
+		yq e "with(.webhooks[]; .clientConfig.caBundle = \"$$cabundle\") | with(.webhooks[]; .clientConfig.url = \"https://$(webhook_service_name):$(webhook_port)\" + .clientConfig.service.path) | with(.webhooks[]; del(.clientConfig.service))" | \
+		kubectl replace -f - ; \
+	else \
+		echo "appcat-sshgateway MutatingWebhookConfiguration not found, skipping"; \
```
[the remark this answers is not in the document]
mdnix: slop

pkg/comp-functions/functions/vshnforgejo/ssh.go
```diff
@@ -0,0 +1,390 @@
...
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									"kubernetes.io/metadata.name": gatewayNamespace,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
mdnix: fair, narrowed it down

Kidswiss (approved):

TheBigLee (approved):

PR #612, opened 2026-03-12
Adjust billing service to servala

zugao:
## Summary

This PR introduces the following changes:

* Removed unit_id field from the CR. as it is not needed anymore in Odoo.

* Event based billing was not clearly enabled when metered billing is involved. Now, when event based billing is enabled then metered billing is disabled. This will have to be revised once we define how we migrate AppCat to event based billing.

* Configured AppCat to get the products from the claim annotations for servala clusters. It has been decided that the following annotations will be populated inside the CR for servala. The portal will take care the product items are properly passed via json format, example:

```
    billing.servala.com/items: '{"items": [{"itemDescription": "desc main", "itemGroupDescription": "desc main",
      "productID": "codey-mini", "value": "1"}, {"itemDescription": "desc storage", "itemGroupDescription": "desc storage", "productID": "cloudscale-ssd", "value": "10"}]}'
    billing.servala.com/salesOrderID: SSS
```

* Since the tests were configured for Forgejo, I had to add billing to this service as it was missing.

* There are some issues with initial maintenance in Forgejo that has to be resolved as it’s directly impacting the billing.

* We just realised that we are sending multiple products for the same instance-id. From Odoo perspective this is wrong, it accepts only unique ids across the whole Odoo instance. It has been decided to use <instance-id>-<short-sha(product-id)> format.

* When creating a new events use creation timestamp of the claim instead of the billing service custom resource so that if we enable billing later we Odoo should consider when the service was created and not when billing was enabled.

* Fixed a race condition with the point 7 having data annotation. 

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1100

### DIFF
--- a/pkg/comp-functions/functions/common/billing_service.go
+++ b/pkg/comp-functions/functions/common/billing_service.go
@@ -109,23 +109,46 @@ func CreateOrUpdateBillingServiceWithOptions(ctx context.Context, svc *runtime.S
 		}
 	}
 
+	// Pre-fetch the existing billing service kube object before annotation checks.
+	// This allows us to preserve it in the desired state even when annotation validation fails,
+	// preventing Crossplane from garbage-collecting it.
+	observedResourceName := comp.GetName() + opts.ResourceNameSuffix
+	kubeObj := &xkube.Object{}
+	observedErr := svc.GetObservedComposedResource(kubeObj, observedResourceName)
+	if observedErr != nil && observedErr != runtime.ErrNotFound {
+		log.Error(observedErr, "cannot get billing service kube object", "service", comp.GetName())
+		return runtime.NewWarningResult(fmt.Sprintf("cannot add billing to service %s", comp.GetName()))
+	}
+
+	// preserveExistingAndWarn re-adds the existing billing service to the desired state (if it
+	// exists) before returning a warning. This prevents Crossplane from garbage-collecting the
+	// billing service when annotation validation fails.
+	preserveExistingAndWarn := func(msg string) *xfnproto.Result {
+		if observedErr == nil {
+			if err := svc.SetDesiredComposedResourceWithName(kubeObj, observedResourceName); err != nil {
+				log.Error(err, "cannot preserve existing billing service", "service", comp.GetName())
+			}
+		}
+		return runtime.NewWarningResult(msg)
+	}
+
 	if prefix != "" {
 		raw := comp.GetAnnotations()[prefix+"/items"]
 		if raw == "" {
-			return runtime.NewWarningResult(fmt.Sprintf("%s/items annotation missing on composite %s", prefix, comp.GetName()))
+			return preserveExistingAndWarn(fmt.Sprintf("%s/items annotation missing on composite %s", prefix, comp.GetName()))
 		}
 		var payload struct {
 			Items []vshnv1.ItemSpec `json:"items"`
 		}
 		if err := json.Unmarshal([]byte(raw), &payload); err != nil {
-			return runtime.NewWarningResult(fmt.Sprintf("cannot parse %s/items on %s: %v", prefix, comp.GetName(), err))
+			return preserveExistingAndWarn(fmt.Sprintf("cannot parse %s/items on %s: %v", prefix, comp.GetName(), err))
 		}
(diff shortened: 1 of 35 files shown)


pkg/comp-functions/functions/common/billing_service.go
```diff
@@ -90,32 +87,97 @@ func CreateOrUpdateBillingServiceWithOptions(ctx context.Context, svc *runtime.S
...
+	observedErr := svc.GetObservedComposedResource(kubeObj, observedResourceName)
+	if observedErr != nil && observedErr != runtime.ErrNotFound {
+		log.Error(observedErr, "cannot get billing service kube object, treating as not found", "service", comp.GetName())
+		return runtime.NewWarningResult(fmt.Sprintf("cannot add billing to service %s", comp.GetName()))
+	}
+
+	// preserveExistingAndWarn re-adds the existing billing service to the desired state (if it
+	// exists) before returning a warning. This prevents Crossplane from garbage-collecting the
+	// billing service when annotation validation fails.
+	preserveExistingAndWarn := func(msg string) *xfnproto.Result {
```
Kidswiss, line 136: Nice idea to prevent accidental deletion on early return.

pkg/comp-functions/functions/common/billing_service.go
```diff
@@ -90,32 +87,97 @@ func CreateOrUpdateBillingServiceWithOptions(ctx context.Context, svc *runtime.S
...
-		}
-		if items[i].ItemGroupDescription == "" {
-			items[i].ItemGroupDescription = itemGroupDescription
+	// For Servala deployments (prefix set), the annotation is the sole source of truth
+	// for items — opts.Items are skipped to avoid duplicating addon items already
+	// declared in the annotation. For non-Servala, include caller-supplied items.
+	var items []vshnv1.ItemSpec
+	if prefix == "" {
+		items = make([]vshnv1.ItemSpec, len(opts.Items))
+		copy(items, opts.Items)
```
Kidswiss: What's the reason behind this copy? Loop pointer bugs should not happen anymore with recent go versions.

pkg/comp-functions/functions/vshnforgejo/maintenance.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // AddMaintenanceJob will add a job to do the maintenance for the instance
 func AddMaintenanceJob(ctx context.Context, comp *vshnv1.VSHNForgejo, svc *runtime.ServiceRuntime) *xfnproto.Result {
-	if err := svc.GetDesiredComposite(comp); err != nil {
+	if err := svc.GetDesiredComposite(comp); err != nil || comp.GetName() == "" {
```
Kidswiss: What's the reason behind the `comp.GetName() == ""` check?

Kidswiss (changes requested):

pkg/comp-functions/functions/common/billing_service.go
zugao: Well it happened during tests so had to find a solution :)

pkg/comp-functions/functions/vshnforgejo/maintenance.go
zugao: I forgot to clean up here, I was trying to solve a non existent problem. Had issues getting the latest version of appcat inside the cluster.

pkg/comp-functions/functions/common/billing_service.go
zugao: The code modifies the slice elements in the loop below (lines 111–119), setting `ItemDescription`,`ItemGroupDescription`, and `InstanceID` on each item. Without copy, modifying items[i] would mutate the original `opts.Items` slice elements. We are managing both servala and appuio and they both have their own way in this function.

Update: I notice we actually don't reuse the items anywhere else so we could just skip the copy, it's just a defensive safeguard.

Kidswiss: > I notice we actually don't reuse the items anywhere else

Yeah that's why I was confused :D

Kidswiss (approved):

mikeshootzz (approved):

PR #621, opened 2026-03-20
Add ability to bring your own bucket

Kidswiss:
## Summary

This commit adds the ability to specify an unmanaged bucket to use for backups.

If specified all internal bucket creation logic is skipped and the credentials provided are used instead.
## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1117


apis/vshn/v1/common_types.go
```diff
@@ -65,6 +71,28 @@ type K8upRestoreSpec struct {
...
 }
 
+type UnmanagedBucket struct {
+	// Endpoint is the unmanaged buckets endpoint without the bucket name
+	// +kubebuilder:validation:Required
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// Bucket is the name of the bucket to be used
+	// +kubebuilder:validation:Required
+	Bucket string `json:"bucket,omitempt"`
```
mikeshootzz: I think you should omit empty here and leave mister empt alone :D

pkg/comp-functions/functions/common/interfaces.go
```diff
@@ -24,6 +24,7 @@ type InfoGetter interface {
 	GetClaimName() string
 	GetSLA() string
 	GetBillingName() string
+	UnamangedBucketProvider
```
mikeshootzz:
```suggestion
	UnmanagedBucketProvider
```

pkg/comp-functions/functions/common/interfaces.go
```diff
@@ -68,3 +69,8 @@ type Alerter interface {
 	GetVSHNMonitoring() vshnv1.VSHNMonitoring
 	GetInstanceNamespace() string
 }
+
+// UnamangedBucketProvider will return an unmanaged bucket or nil if it's not defined
+type UnamangedBucketProvider interface {
```
mikeshootzz:
```suggestion
// UnmanagedBucketProvider will return an unmanaged bucket or nil if it's not defined
type UnmanagedBucketProvider interface {
```

pkg/comp-functions/functions/common/backup/backup.go
```diff
@@ -427,3 +428,52 @@ func PatchConnectionSecretWithAllowDeletion(ctx context.Context, comp common.Inf
...
+	if ub := comp.GetUnmanagedBucket(); ub != nil {
+		secret := &corev1.Secret{}
+
+		_, err := svc.CopyKubeResource(ctx, secret, comp.GetName()+"-bucket-credentials", ub.AccessKey.Name, comp.GetClaimNamespace(), comp.GetInstanceNamespace())
+
+		return BucketInfo{
+			Bucket:   ub.Bucket,
+			Endpoint: ub.Endpoint,
+			SecretID: &ub.SecretKey,
+			KeyID:    &ub.AccessKey,
```
mikeshootzz, line 446: You don't return the buckets region here, don't think this is intentional

mikeshootzz (changes requested):

pkg/comp-functions/functions/common/backup/backup.go
Kidswiss: It is, k8up doesn't care about the region. I thought I made a comment about this though...

apis/vshn/v1/common_types.go
Kidswiss: Poor mr. empt is retired now.

mikeshootzz (approved):
Not sure if GitHub isn't displaying the diff correctly, but I still see "Unamanaged" instead of "Unmanaged". Quickly check that please, but the rest looks good.

Kidswiss:
> Not sure if GitHub isn't displaying the diff correctly, but I still see "Unamanaged" instead of "Unmanaged". Quickly check that please, but the rest looks good.

I don't see it anymore in the diff.

However, grepping over the code I found another `Unamanaged` unrelated to this change 😂 fixed it as well.

pkg/comp-functions/functions/common/interfaces.go
zugao: I see `UnamangedBucketProvider` ...

zugao (approved):
I still see `UnamangedBucketProvider` please make sure we don't have these typos in the api, rest is fine

PR #622, opened 2026-03-23
Fix keycloak restore copy job

mikeshootzz:
## Summary

This fixes the copy job used when restoring a vshnkeycloak instance. By using a staging secret we can ensure that provider kubernetes doesn't overwrite the secret again on a reconcile.

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1119

### DIFF
--- a/pkg/comp-functions/functions/common/password.go
+++ b/pkg/comp-functions/functions/common/password.go
@@ -95,6 +95,60 @@ func AddGenericSecret(comp InfoGetter, svc *runtime.ServiceRuntime, suffix strin
 
 }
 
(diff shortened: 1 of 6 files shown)


zugao (approved):
Please test against an existing keycloak instance and check that password remains the same.

mikeshootzz:
> Please test against an existing keycloak instance and check that password remains the same.

@zugao I've deployed an instance on lab, switched to the feature branch and checked the credentials before and after. They are not affected.

Kidswiss (approved):

PR #623, opened 2026-03-24
Allow major version upgrades for cnpg

mikeshootzz:
## Summary

This extends the webhook for PostgreSQL to allow for major version upgrades with CNPG. Major updates with CNPG "just work" thus only needing adjustments to the webhook. 

Since downgrading is not supported by CNPG this is explicitly enforced via a webhook. Nextcloud uses the same logic, so this is extracted into a helper function.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1121

### DIFF
--- a/pkg/comp-functions/functions/vshnpostgrescnpg/backup.go
+++ b/pkg/comp-functions/functions/vshnpostgrescnpg/backup.go
@@ -60,12 +60,16 @@ func insertBackupValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *
 	}
 
 	// Enable the barman-cloud plugin in the cluster configuration
+	// serverName determines the path within the object store: {destinationPath}/{serverName}/
+	// By including the major version, each major version upgrade writes to a separate path,
+	// preventing WAL timeline conflicts after pg_upgrade resets the timeline to 1.
 	clusterPlugins := []map[string]any{{
 		"name":          "barman-cloud.cloudnative-pg.io",
 		"enabled":       true,
 		"isWALArchiver": true,
 		"parameters": map[string]any{
 			"barmanObjectName": "postgresql-object-store",
+			"serverName":       "postgresql-" + comp.Spec.Parameters.Service.MajorVersion,
 		},
 	}}
 
--- a/pkg/controller/webhooks/default_handler.go
+++ b/pkg/controller/webhooks/default_handler.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


pkg/controller/webhooks/default_handler.go
```diff
@@ -634,3 +635,24 @@ func checkManualVersionManagementWarnings(maintenance vshnv1.VSHNDBaaSMaintenanc
 	}
 	return warnings
 }
+
+// validateNoDowngrade returns an error if newVersion is lower than oldVersion.
+// Both versions are parsed tolerantly, so plain major versions ("15"), semver ("15.9"), and full versions ("15.9.1") are all accepted.
+// If either version is empty or unparseable as an old version, the check is skipped.
+func validateNoDowngrade(oldVersion, newVersion string, path *field.Path) *field.Error {
```
Kidswiss: Just a nit: putting this into the `default_handler.go` file is a misnomer, since this is not a default behavior for all services.

pkg/comp-functions/functions/vshnpostgrescnpg/backup.go
```diff
@@ -58,12 +58,16 @@ func insertBackupValues(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNPostgreSQL
 	}
 
 	// Enable the barman-cloud plugin in the cluster configuration
+	// serverName determines the path within the object store: {destinationPath}/{serverName}/
```
Kidswiss, line 63: How does this influence the retention handler? Will it still be able to cleanup the old major version WALs and backups?

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnpostgrescnpg/backup.go
mikeshootzz: That's a good point. I don't believe that it will be able to clean those up then... since we can't downgrade anyway we have to find a way to prune these after the upgrade

pkg/controller/webhooks/default_handler.go
mikeshootzz: Good point, I'll fix this

pkg/comp-functions/functions/vshnpostgrescnpg/backup.go
mikeshootzz: As discussed there will be a follow-up created to address this.

Kidswiss (approved):

zugao (approved):
LGTM

PR #628, opened 2026-03-26
Add missing sidecar and init container to config

Kidswiss:
## Summary

- dbchecker for Nextcloud, now uses the same image as for Keycloak
- redis-export for redis is now configurable, but we will need to build it from source ourselves, as the upstream image is not compatible with whatever bitnami does

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1126

### DIFF
--- a/pkg/comp-functions/functions/vshnredis/redis_deploy.go
+++ b/pkg/comp-functions/functions/vshnredis/redis_deploy.go
@@ -304,9 +304,14 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 			return nil, err
 		}
 
+		exporterTag := svc.Config.Data["exporter_tag"]
+		if exporterTag == "" {
+			exporterTag = "1.76.0-debian-12-r0"
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/comp-functions/functions/vshnredis/redis_deploy.go
```diff
@@ -304,9 +304,14 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 			return nil, err
 		}
 
+		exporterTag := svc.Config.Data["exporter_tag"]
+		if exporterTag == "" {
+			exporterTag = "1.76.0-debian-12-r0"
```
zugao, line 309: I guess it's worse to return an error here than just default to a tag...

zugao (approved):

mikeshootzz (approved):

pkg/comp-functions/functions/vshnredis/redis_deploy.go
Kidswiss: Yeah, AFAIK the default won't match our custom build image. So we have to set it to something valid.

PR #630, opened 2026-03-31
Initialize OpenBao Service 

norbertgruszka:
## Summary

Initialize OpenBao Service. The goal is to provide users with an option to setup HA OpenBao Cluster using Appcat.

Initial implementation contains:
* Auto generated Kubernetes API 
* Deployment step
* HCL generator 
* Self-signed TLS certificates

The service is not yet ready for production use. Further development will follow.

## How to test?

1. Create kindev environment **with at least 3 nodes**. For now, the OpenBao Cluster starts in HA.
2. Create new`vshnopenbaos.vshn.appcat.vshn.io` resource.
```yaml
apiVersion: vshn.appcat.vshn.io/v1
kind: VSHNOpenBao
metadata:
  name: openbao-test
spec:
  parameters:
    security:
      deletionProtection: false
  writeConnectionSecretToRef:
    name: "openbao-test"
```
3. Search for newly created namespace with `openbao` name included. Use `bao operator init -tls-skip-verify` to initialize the cluster manually with `kubectl exec` command.
4. The cluster should be operational.

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1128


apis/vshn/v1/vshn_openbao.go
```diff
@@ -0,0 +1,320 @@
...
+	Monitoring VSHNMonitoring `json:"monitoring,omitempty"`
+
+	// Instances defines the number of instances to run.
+	Instances int `json:"instances,omitempty"`
+}
+
+// VSHNOpenBaoServiceSpec contains OpenBao DBaaS specific properties
+type VSHNOpenBaoServiceSpec struct {
+	// +kubebuilder:validation:Enum=<TBD>
+	// +kubebuilder:default=<TBD>
```
Kidswiss: Should probably be defined before release.

apis/vshn/v1/vshn_openbao.go
```diff
@@ -0,0 +1,320 @@
...
+
+	// Version contains supported version of OpenBao.
+	// Multiple versions are supported. The latest version <TBD> is the default version.
+	Version string `json:"version,omitempty"`
+
+	// Openbaosettings contains additional OpenBao settings.
+	Openbaosettings string `json:"openbaosettings,omitempty"`
```
Kidswiss: Is this really just a string? Or only a placeholder? Usually settings are modelled with a `map[string]any`

pkg/comp-functions/functions/vshnopenbao/tls.go
```diff
@@ -0,0 +1,250 @@
...
+
+	serviceName := comp.GetName()
+	ns := comp.GetInstanceNamespace()
+
+	selfSignedIssuerOpts := &common.TLSOptions{
+		IssuerOptions: []common.IssuerOption{
+			withIssuerOfTypeSelfSigned(),
+		},
+	}
+	selfSignedIssuer := createIssuer(ns, serviceName+tlsSelfSignedIssuerSuffix, selfSignedIssuerOpts)
```
Kidswiss: I guess `common.CreateCertificate()` wasn't flexible enough to handle this?

I'm wondering if it would make sense to refactor it make it more flexible.

Kidswiss (changes requested):
The code looks good. Just a few remarks.

I haven't tested it yet, but will do some tests later.

> Search for newly created namespace with openbao name included. Use bao operator init -tls-skip-verify to initialize the cluster manually with kubectl exec command.

One idea: this should probably be a post-install hook in the helm chart, since it's still part of the provisioning.

pkg/comp-functions/functions/vshnopenbao/tls.go
norbertgruszka: I didn't know `common.CreateCertificate()` exists. I think we were looking at Redis and certificates are created differently there. 

I update the code, now it's using `common.CreateCertificate()`

apis/vshn/v1/vshn_openbao.go
norbertgruszka: The Version field was just a placeholder. The version is set in Components Appcat.

apis/vshn/v1/vshn_openbao.go
norbertgruszka: That was a placeholder too, so I removed it for now. We'll add it back later probably with better defined type.

pkg/comp-functions/functions/vshnopenbao/tls.go
Kidswiss: Yeah, some older services might not use the common library because they were migrated from legacy crossplane compositions and are thus stuck in a kind of limbo.

apis/vshn/v1/vshn_openbao.go
```diff
@@ -0,0 +1,309 @@
...
+type VSHNOpenBaoServiceSpec struct {
+	// +kubebuilder:validation:Enum="besteffort";"guaranteed"
+	// +kubebuilder:default="besteffort"
+	// ServiceLevel defines the service level of this service. Either Best Effort or Guaranteed Availability is allowed.
+	ServiceLevel VSHNDBaaSServiceLevel `json:"serviceLevel,omitempty"`
+}
+
+// VSHNOpenBaoSizeSpec contains settings to control the sizing of a service.
+type VSHNOpenBaoSizeSpec struct {
+
```
Kidswiss, line 96: This one seems unused. At least according to the linter.

apis/vshn/v1/vshn_openbao.go
```diff
@@ -0,0 +1,309 @@
...
+	v.Status.Schedules.Backup = schedule
+} // GetServiceName returns the name of this service
+func (v *VSHNOpenBao) GetServiceName() string {
+	return "openbao"
+}
+
+// GetPDBLabels returns the labels to be used for the PodDisruptionBudget
+// it should match one unique label od pod running in instanceNamespace
+// without this, the PDB will match all pods
+func (v *VSHNOpenBao) GetPDBLabels() map[string]string {
```
Kidswiss, line 257: Does the helm chart you're using already deploy PDBs? If not setting the labels to match the pods and then calling `common.AddPDBSettings()` will automagically deploy them. But I think the `instances` parameter needs to be set for them to work...

pkg/comp-functions/functions/vshnopenbao/deploy.go
```diff
@@ -0,0 +1,159 @@
...
+		"global": map[string]interface{}{
+			"tlsDisable": false,
+		},
+		"injector": map[string]any{
+			"enabled": false,
+		},
+		"server": map[string]any{
+			"ha": map[string]any{
+				"enabled": true,
+				"config":  "# Config provided via external file\n",
```
Kidswiss, line 59: I'm wondering why the config is provided via an external file?

Kidswiss (commented):
I tried to create an instance, but I struggle to get the cluster initialized properly.

My steps:
```
k -n vshn-openbao-openbao-test-v54vf exec -it openbao-test-v54vf-0 -- sh

bao operator init -tls-skip-verify
bao operator unseal -tls-skip-verify

k -n vshn-openbao-openbao-test-v54vf exec -it openbao-test-v54vf-1 -- sh

bao operator unseal -tls-skip-verify
Unseal Key (will be hidden):
Error unsealing: Error making API request.

URL: PUT https://127.0.0.1:8200/v1/sys/unseal
Code: 400. Errors:

* Vault is not initialized
```

The first pod looks fine, the others throw this error. According to the openbao docs `bao operator init` has to be run only once.

Checking the config within the pod I see this:
```
cat userconfig/openbao-hcl-config/config.hcl
ui           = true
log_level    = "info"
log_format   = "json"
cluster_name = "openbao-test-v54vf"
api_addr     = "https://openbao-test-v54vf:8200"
cluster_addr = "https://openbao-test-v54vf:8201"

listener "tcp" {
  address         = "[::]:8200"
  cluster_address = "[::]:8201"
  tls_disable     = false
  tls_cert_file   = "/openbao/userconfig/openbao-tls/tls.crt"
  tls_key_file    = "/openbao/userconfig/openbao-tls/tls.key"
}

storage "raft" {
  path = "/openbao/data"
}
```

Isn't there supposed to be something to tell the instance where its peers are?

pkg/comp-functions/functions/vshnopenbao/deploy.go
norbertgruszka: I wanted to make troubleshooting slightly easier by providing HCL configuration as a config map. Although, we can also just pass it as a string in Helm Values.

Is that something that we should do? To me, it's a preference choice and I'm fine with both.

apis/vshn/v1/vshn_openbao.go
norbertgruszka: The OpenBao Helm chart can create PDBs [here](https://github.com/openbao/openbao-helm/blob/main/charts/openbao/templates/injector-disruptionbudget.yaml). At the moment we don't generate PDBs, but we'll enable them in the future.

What should we do in that situation? Can we just leave `GetPDBLabels()` as it is? At the moment it returns an empty map.

apis/vshn/v1/vshn_openbao.go
norbertgruszka: That's true, the generic `VSHNSizeSpec` struct was used. The OpenBao Service is using `VSHNOpenBaoSizeSpec` now.

ioboi (approved):

pkg/comp-functions/functions/vshnopenbao/deploy.go
Kidswiss: I'm also fine with both, I was just curious about the reason, since this is more complex than the other way.

apis/vshn/v1/vshn_openbao.go
Kidswiss: For the moment it's fine. But once you get around adding the PDB, it makes sense to use our logic, so it's uniform between all services.

pkg/comp-functions/functions/vshnopenbao/config.go
```diff
@@ -95,7 +95,7 @@ func CreateHCLConfigMap(ctx context.Context, comp *vshnv1.VSHNOpenBao, svc *runt
 					{
 						AutoJoin:            fmt.Sprintf(`provider=k8s namespace=%s label_selector="app.kubernetes.io/name=openbao,component=server"`, ns),
 						AutoJoinScheme:      "https",
-						AutoJoinPort:        hclConfigApiPort,
+						AutoJoinPort:        8200,
```
ioboi, line 98: I think this port depends on `const hclConfigApiPort = "8200"`. Maybe it make sense to change this constant to `uint` as well.

Kidswiss (approved):

PR #633, opened 2026-04-07
Add self-service restore for CNPG

mikeshootzz:
## Summary

* Add self-service restore for CNPG PostgreSQL, similar to the existing StackGres restore flow
* Add barman cloud API package (`apis/barmancloud/v1/`) for typed access to backup object store resources
* Include composition function, backup credential copy script, and tests

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1134


pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -141,6 +141,12 @@ func deployPostgresSQLUsingCNPG(ctx context.Context, comp *vshnv1.VSHNPostgreSQL
 		return runtime.NewFatalResult(fmt.Errorf("cannot create helm values: %w", err))
 	}
 
+	// Handle restore before backup setup — may skip Helm release creation in Phase 1
```
zugao: What is phase 1?

zugao (approved):
Looks good overall. I suppose you have tested the restore with databases with data. Is the new secret propagated to the claim namespace with the same crendetials?

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
mikeshootzz: This is the first reconciliation loop or "phase"

mikeshootzz:
> Looks good overall. I suppose you have tested the restore with databases with data. Is the new secret propagated to the claim namespace with the same crendetials?

@zugao No the credentials are not the same, but this is also the case for StackGres and explicitly documented already: https://docs.appcat.ch/vshn-managed/postgresql/restore.html

And yes I have tested it with data and also dropped tables to check that PITR works

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
zugao: I would update the comment and use reconciliation loop instead of phase

mikeshootzz: done.

Kidswiss (approved):

PR #634, opened 2026-04-08
Add support for walStorage size

TheBigLee:
## Summary

* This PR adds support for setting the walStorage for CNPG clusters
## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1135


Kidswiss (approved):

apis/vshn/v1/dbaas_vshn_postgresql.go
```diff
@@ -53,6 +53,9 @@ type VSHNPostgreSQLParameters struct {
 	// Size contains settings to control the sizing of a service.
 	Size VSHNSizeSpec `json:"size,omitempty"`
 
+	// WalStorage contains settings to control the walStorage for a Service
```
Kidswiss: NIT: we should mention that this only applies to CNPG

mikeshootzz (approved):

zugao (approved):
LGTM

apis/vshn/v1/dbaas_vshn_postgresql.go
TheBigLee: Fixed

TheBigLee:
I've also added the feature to disable walStorage completely. Please check again.

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -299,8 +299,8 @@ func createCnpgHelmValues(ctx context.Context, svc *runtime.ServiceRuntime, comp
 	}
 
 	svc.Log.Info("Set WalStorage")
-	if comp.Spec.Parameters.WallStorage.Size != "" {
```
BacLuc: The WallStorage 🤔

mikeshootzz (changes requested):
Whether the WAL storage is disabled or not is immutable, could you please enforce this via our webhook. I would do the same for disallowing downsizing

apis/vshn/v1/dbaas_vshn_postgresql.go
zugao: why this field is under `parameter`? It should be at least under `service` but even then I would disagree

zugao (changes requested):
I would brainstorm first whether we need this parameter exposed at all and not calculated on the go. We might want to put it under size but then it would be a specific cnpg implementation. I would think carefully before updating the API with this field

apis/vshn/v1/dbaas_vshn_postgresql.go
TheBigLee: moved

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
TheBigLee: Fixed

PR #635, opened 2026-04-09
Fix memory leaks

zugao:
## Summary

* Fix 1 - probers map never shrinks. StopProbe called cancel() to stop a probe goroutine but never removed the key from the map. Every deleted or suspended service instance left a dead entry permanently.
* FIX 2 - Unbounded goroutine accumulation. `runProbe` spawned `go m.sendProbe()` every second unconditionally. When a database endpoint is unreachable, each goroutine blocks for the full 5-second timeout. With a 1-second tick and 5-second drain rate, 5 goroutines pile up per instance at steady state. This was the main issue that triggered the OOM that I noticed.
* Fix 3 — HTTP response body never closed. The Go HTTP transport cannot reuse a TCP connection unless the body is drained and closed, so a new TCP connection was opened on every probe call.
* Fix 4 - HTTP probe ignores context. Using now `http.NewRequestWithContext(ctx, ...) `+ `h.client.Do(req)`.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1137

### DIFF
--- a/pkg/sliexporter/probes/http.go
+++ b/pkg/sliexporter/probes/http.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"io"
 	"net/http"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -19,7 +20,7 @@ type HTTP struct {
 }
 
 func NewHTTP(url string, tlsEnabled bool, cacert *x509.Certificate, service, name, claimNamespace, instanceNamespace, organization, servicelevel, compositionName string, ha bool) *HTTP {
-	transport := http.DefaultTransport
+	transport := http.DefaultTransport.(*http.Transport).Clone()
 	if tlsEnabled {
 		caPool := x509.NewCertPool()
 		caPool.AddCert(cacert)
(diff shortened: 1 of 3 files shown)


pkg/sliexporter/probes/http.go
```diff
@@ -19,7 +20,7 @@ type HTTP struct {
 }
 
 func NewHTTP(url string, tlsEnabled bool, cacert *x509.Certificate, service, name, claimNamespace, instanceNamespace, organization, servicelevel, compositionName string, ha bool) *HTTP {
-	transport := http.DefaultTransport
+	transport := &http.Transport{}
```
mikeshootzz:
```suggestion
	transport := http.DefaultTransport.(*http.Transport).Clone()
```
`&http.Transport{}` is a zero-value transport with no timeouts, no connection limits, and no idle connection expiry. Cloning `DefaultTransport` gives you an isolated transport (so `CloseIdleConnections` only affects this prober) while keeping all the defaults like dial timeouts and idle connection cleanup

mikeshootzz (changes requested):

pkg/sliexporter/probes/http.go
zugao: Yeah we need to clone the DefaultTransport otherwise we loose what we get from the default implementation.

Kidswiss (approved):

mikeshootzz (approved):

PR #636, opened 2026-04-09
Expose CNPG PostgreSQL via loadbalancer

Kidswiss:
## Summary
This commit adds support to expose CNPG PostgreSQL instances via LoadBalancer.

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1138

### DIFF
--- a/pkg/comp-functions/functions/common/release.go
+++ b/pkg/comp-functions/functions/common/release.go
@@ -12,6 +12,7 @@ import (
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
 )
 
 // GetReleaseValues returns the parsed values from the given release.
@@ -95,6 +96,10 @@ func NewRelease(ctx context.Context, svc *runtime.ServiceRuntime, comp InfoGette
 			Name: comp.GetName(),
 		},
 		Spec: xhelmv1.ReleaseSpec{
+			// This will make provider-helm retry up to 10 times
+			// to install a release. Otherwise a release could
+			// get stuck indefinitely due to race conditions.
+			RollbackRetriesLimit: ptr.To[int32](10),
 			ForProvider: xhelmv1.ReleaseParameters{
 				Chart: xhelmv1.ChartSpec{
 					Repository: svc.Config.Data["chartRepository"],
--- a/pkg/comp-functions/functions/vshnpostgrescnpg/connection_details.go
+++ b/pkg/comp-functions/functions/vshnpostgrescnpg/connection_details.go
@@ -116,6 +116,10 @@ func generateConnectionDetailInfoForRelease(comp *vshnv1.VSHNPostgreSQL, svc *ru
 		ToConnectionSecretKey: ClusterInstanceCdField,
 	})
 
+	if comp.Spec.Parameters.Network.ServiceType == string(corev1.ServiceTypeLoadBalancer) && externalAccessEnabled(svc) {
+		connectionDetails = append(connectionDetails, getLoadBalancerConnectionDetail(comp, svc))
+	}
+
 	return connectionDetails
 }
 
@@ -162,4 +166,18 @@ func getConnectionDetails(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNPostgreS
 	}
 
 	return cd, nil
+}
+
+func getLoadBalancerConnectionDetail(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) v1beta1.ConnectionDetail {
+	return v1beta1.ConnectionDetail{
+		ObjectReference: corev1.ObjectReference{
+			APIVersion: "v1",
+			Kind:       "Service",
+			Name:       "primary",
+			Namespace:  comp.GetInstanceNamespace(),
+			FieldPath:  "status.loadBalancer.ingress[0].ip",
+		},
+		ToConnectionSecretKey:  "LOADBALANCER_IP",
+		SkipPartOfReleaseCheck: true,
+	}
 }
--- a/pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
+++ b/pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
@@ -2,8 +2,10 @@ package vshnpostgrescnpg
 
 import (
 	"context"
+	"crypto/x509"
 	_ "embed"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"strings"
@@ -12,13 +14,16 @@ import (
 	cmv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
 	xfnproto "github.com/crossplane/function-sdk-go/proto/v1"
+
 	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
 
 	"github.com/vshn/appcat/v4/pkg/common/utils"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/functions/common"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	"github.com/vshn/appcat/v4/pkg/controller/webhooks"
+	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -88,6 +93,14 @@ func createCerts(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error
 	}
 
 	svcName := "postgresql-rw"
+
+	// Once the loadbalancer ip is in the connection details, we can use it to fix the tls cert
+	ipAddresses := []string{}
+	cd := svc.GetObservedConnectionDetails()
+	if v, ok := cd["LOADBALANCER_IP"]; ok && comp.Spec.Parameters.Network.ServiceType == string(corev1.ServiceTypeLoadBalancer) && externalAccessEnabled(svc) {
+		ipAddresses = append(ipAddresses, string(v))
+	}
+
 	certificate := &cmv1.Certificate{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      comp.GetName(),
@@ -117,6 +130,7 @@ func createCerts(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error
 				svcName + "." + comp.GetInstanceNamespace() + ".svc.cluster.local",
 				svcName + "." + comp.GetInstanceNamespace() + ".svc",
 			},
+			IPAddresses: ipAddresses,
 			IssuerRef: certmgrv1.ObjectReference{
 				Name:  comp.GetName(),
 				Kind:  selfSignedIssuer.GetObjectKind().GroupVersionKind().Kind,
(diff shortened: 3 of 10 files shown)


pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -475,3 +494,44 @@ func createCnpgSCCRoleBinding(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceR
...
+		}
+
+		services := map[string]any{
+			"additional": []map[string]any{
+				{
+					"selectorType": "rw",
+					"serviceTemplate": map[string]any{
+						"metadata": map[string]any{
+							"name":         "primary",
+							"annotations:": annotations,
```
mdnix: typo
```suggestion
							"annotations": annotations,
```

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -12,13 +12,15 @@ import (
 	cmv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
 	xfnproto "github.com/crossplane/function-sdk-go/proto/v1"
+	"github.com/stretchr/testify/assert/yaml"
```
mdnix: wrong yaml pkg?

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -88,6 +90,20 @@ func createCerts(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error
...
 
 	svcName := "postgresql-rw"
+	dnsNames := []string{
+		svcName + "." + comp.GetInstanceNamespace() + ".svc.cluster.local",
+		svcName + "." + comp.GetInstanceNamespace() + ".svc",
+	}
+
+	// Once the loadbalancer ip is in the connection details, we can use it to fix the tls cert
+	cd := svc.GetObservedConnectionDetails()
+	if err != nil {
```
mdnix: Checking a stale err here

pkg/comp-functions/functions/vshnpostgrescnpg/connection_details.go
```diff
@@ -12,6 +12,7 @@ import (
 	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
 	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
mdnix: why re-import it again as `v1`?

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -303,6 +317,11 @@ func createCnpgHelmValues(ctx context.Context, svc *runtime.ServiceRuntime, comp
 		return map[string]any{}, fmt.Errorf("cannot add user management values: %w", err)
 	}
 
+	// LoadBalancer: Inject loadbalancer config if enabled
+	if err := addLoadbalancerConfig(svc, comp, values); err != nil {
+		return map[string]any{}, fmt.Errorf("cannot add service values: %w", err)
```
mdnix: This error should be a bit more specific

mdnix (changes requested):
Regarding the race that's happening... What if we added an  observe only kube object for the tls-certificate secret as a readiness gate? We could then parse the observed tls.crt and keep it ResourceUnReady until the bytes actually contain the LB IP in the IPAddresses SAN.

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
Kidswiss: Stupid autoimporter...

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
Kidswiss: Leftover of my race condition hunting. Good catch.

pkg/comp-functions/functions/vshnpostgrescnpg/connection_details.go
Kidswiss: See above :D

Kidswiss:
> Regarding the race that's happening... What if we added an observe only kube object for the tls-certificate secret as a readiness gate? We could then parse the observed tls.crt and keep it ResourceUnReady until the bytes actually contain the LB IP in the IPAddresses SAN.

I kinda went with this.

So the logic is that if LoadBalancer is enabled we will always mark the certificate as unready until we can actually find the IP address in the certificate which will get exposed as the connection detail. I think the logic for this is bigger than the rest of the feature 🙈

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -475,3 +551,44 @@ func createCnpgSCCRoleBinding(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceR
 	}
 	return svc.SetDesiredKubeObject(rb, comp.GetName()+"-scc-rb")
 }
+
+func addLoadbalancerConfig(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNPostgreSQL, values map[string]any) error {
```
mikeshootzz, line 562: You're missing the `externalDatabaseConnectionsEnabled` config check that StackGres has in load_balancer.go:29 and load_balancer.go:62, allowing users to bypass the operator's kill-switch for external database connections.

mikeshootzz (changes requested):

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
Kidswiss: I'm considered adding it. We usually forget to enable it for customers anyway. So I'm not sure if it really adds much benefit to the table.

Kidswiss: I thought a bit more about this and added it.

pkg/comp-functions/functions/vshnpostgrescnpg/deploy_test.go
```diff
@@ -236,6 +245,149 @@ func Test_pinImageTag(t *testing.T) {
...
+
+		annotations, found, err := unstructured.NestedStringMap(entry, "serviceTemplate", "metadata", "annotations")
+		require.NoError(t, err)
+		require.True(t, found)
+		assert.Equal(t, "nlb", annotations["service.beta.kubernetes.io/load-balancer-type"])
+	})
+}
+
+func Test_createCerts_LoadBalancer(t *testing.T) {
+	t.Run("cert marked ready when LB IP is in certificate DNSNames", func(t *testing.T) {
```
zugao:
```suggestion
	t.Run("cert marked ready when LB IP is in certificate IPAddresses", func(t *testing.T) {
```
Small nits I think we are checking for IpAddresses and not DNSNames

pkg/comp-functions/functions/vshnpostgrescnpg/deploy_test.go
```diff
@@ -236,6 +245,149 @@ func Test_pinImageTag(t *testing.T) {
...
+		err := createCerts(comp, svc)
+		require.NoError(t, err)
+
+		desired := svc.GetAllDesired()
+		certRes, ok := desired[resource.Name("certificate")]
+		require.True(t, ok, "certificate resource should exist in desired state")
+		assert.Equal(t, resource.ReadyTrue, certRes.Ready, "certificate should be marked ready when LB IP is in cert DNSNames")
+	})
+
+	t.Run("cert marked unready when LB IP is not in certificate DNSNames", func(t *testing.T) {
```
zugao:
```suggestion
	t.Run("cert marked unready when LB IP is not in certificate IPAddresses", func(t *testing.T) {
```
Here as well

pkg/comp-functions/functions/vshnpostgrescnpg/deploy_test.go
```diff
@@ -236,6 +245,149 @@ func Test_pinImageTag(t *testing.T) {
...
+		err := createCerts(comp, svc)
+		require.NoError(t, err)
+
+		desired := svc.GetAllDesired()
+		certRes, ok := desired[resource.Name("certificate")]
+		require.True(t, ok, "certificate resource should exist in desired state")
+		assert.Equal(t, resource.ReadyFalse, certRes.Ready, "certificate should be marked unready when LB IP is not in cert DNSNames")
+	})
+
+	t.Run("LB IP is added to certificate DNSNames", func(t *testing.T) {
```
zugao:
```suggestion
	t.Run("LB IP is added to certificate IPAddresses", func(t *testing.T) {
```
Here as well

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -88,6 +93,14 @@ func createCerts(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error
 	}
 
 	svcName := "postgresql-rw"
+
+	// Once the loadbalancer ip is in the connection details, we can use it to fix the tls cert
+	ipAddresses := []string{}
+	cd := svc.GetObservedConnectionDetails()
+	if v, ok := cd["LOADBALANCER_IP"]; ok && comp.Spec.Parameters.Network.ServiceType == string(corev1.ServiceTypeLoadBalancer) {
```
zugao: Nit: I would add here also `externalAccessEnabled(svc)` just to be consistent. I know in practice this should be no problem but you never know what edge cases can do.

zugao (approved):
I approved this PR, cool that now we are dedicating more time to intermediate state of our services so that customer can know what is happening with the instance (or us). I added some small nits.

pkg/comp-functions/functions/vshnpostgrescnpg/deploy_test.go
Kidswiss: Good catch, at first I added it as a DNS name which was of course wrong.

Kidswiss:
> I approved this PR, cool that now we are dedicating more time to intermediate state of our services so that customer can know what is happening with the instance (or us). I added some small nits.

Well it's kinda necessary here, as things would just be plainly wrong. The thing with exposing the intermediate state is that it's really hard to do with Crossplane, as we can't control the reconcile loop directly. Figuring out what to actually observe here took most of the development time of this feature...

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
Kidswiss: Satisfied the edgelords.

mikeshootzz (approved):

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
```diff
@@ -131,6 +145,62 @@ func createCerts(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) error
...
+				svc.Log.Info("cannot decode tls certificate")
+				svc.AddResult(runtime.NewWarningResult("cannot decode tls certificate"))
+			} else {
+				cert, err := x509.ParseCertificate(block.Bytes)
+				if err != nil {
+					svc.Log.Error(err, "cannot parse tls certificate")
+					svc.AddResult(runtime.NewWarningResult(fmt.Sprintf("cannot parse tls certificate: %s", err)))
+				}
+
+				for _, ip := range cert.IPAddresses {
```
mdnix, line 202: I see a panic risk here, if above `cert` is nil or has an err we still continue going down that branch.

pkg/comp-functions/functions/vshnpostgrescnpg/connection_details.go
```diff
@@ -163,3 +167,17 @@ func getConnectionDetails(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNPostgreS
...
 }
+
+func getLoadBalancerConnectionDetail(comp *vshnv1.VSHNPostgreSQL, svc *runtime.ServiceRuntime) v1beta1.ConnectionDetail {
+	return v1beta1.ConnectionDetail{
+		ObjectReference: corev1.ObjectReference{
+			APIVersion: "v1",
+			Kind:       "Service",
+			Name:       "primary",
+			Namespace:  comp.GetInstanceNamespace(),
+			FieldPath:  "status.loadBalancer.ingress[0].ip",
```
mdnix, line 178: Ok for now, but we should think about how we would handle adding a proper hostname

pkg/comp-functions/functions/common/release.go
```diff
@@ -95,6 +96,10 @@ func NewRelease(ctx context.Context, svc *runtime.ServiceRuntime, comp InfoGette
 			Name: comp.GetName(),
 		},
 		Spec: xhelmv1.ReleaseSpec{
+			// This will make provider-helm retry up to 10 times
+			// to install a release. Otherwise a release could
+			// get stuck indefinitely due to race conditions.
+			RollbackRetriesLimit: ptr.To[int32](10),
```
mdnix, line 102: Worth nothing that this will apply to all services

pkg/comp-functions/functions/vshnpostgrescnpg/connection_details.go
Kidswiss: Yes, hostnames will be a completely different beast as it involves external DNS.

pkg/comp-functions/functions/vshnpostgrescnpg/deploy.go
Kidswiss: As discussed: I will add a nil check and do a fatal if it's nil, so we don't mess up the state.

Kidswiss: Satisfied more edgelords.

mdnix (approved):

PR #639, opened 2026-04-10
Add controller for garage bucket cleanup

TheBigLee:
## Summary

* This PR adds a new logic to the controller, to delete all the data from garage buckets, that are in the deletion state.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1141

### DIFF
--- a/pkg/controller/garagebucket/cleanup_controller.go
+++ b/pkg/controller/garagebucket/cleanup_controller.go
@@ -1,0 +1,189 @@
+package garagebucket
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+
+	"github.com/go-logr/logr"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+//+kubebuilder:rbac:groups=garage.rajsingh.info,resources=garagebuckets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=vshn.appcat.vshn.io,resources=vshngarages,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+
+const adminCredentialsSecret = "admin-s3-credentials"
+
+var garageBucketGVK = schema.GroupVersionKind{
+	Group:   "garage.rajsingh.info",
+	Version: "v1alpha1",
+	Kind:    "GarageBucket",
+}
+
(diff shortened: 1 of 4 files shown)


pkg/controller/garagebucket/cleanup_controller.go
```diff
@@ -0,0 +1,233 @@
...
+
+	bucketName := xObjectBucket.Status.BucketName
+	if bucketName == "" {
+		bucketName = xObjectBucket.GetName()
+	}
+
+	// Get the Helm Release to find the garageClaimNamespace.
+	release := &xhelmv1.Release{}
+	if err := r.Get(ctx, types.NamespacedName{Name: xObjectBucket.GetName()}, release); err != nil {
+		if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
TheBigLee: Might have to look into that

pkg/controller/garagebucket/cleanup_controller.go
```diff
@@ -0,0 +1,224 @@
...
+			return compositionRefName(e.Object) == garageCompositionName
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return compositionRefName(e.ObjectNew) == garageCompositionName
+		},
+		DeleteFunc:  func(event.DeleteEvent) bool { return false },
+		GenericFunc: func(event.GenericEvent) bool { return false },
+	}
+
+	objBucket := &unstructured.Unstructured{}
```
Kidswiss: Why not use a v1.XObjectBucket directly? Then you won't need the GVK.

Something like here: https://github.com/vshn/appcat/blob/develop/pkg/controller/billing/billing_handler.go#L84

pkg/controller/garagebucket/cleanup_controller.go
```diff
@@ -0,0 +1,224 @@
...
+	r := &CleanupReconciler{
+		Client: c,
+		log:    log.WithName("controller").WithName("garagebucket-cleanup"),
+	}
+	r.emptyBucketFn = r.emptyBucket
+	return r
+}
+
+// SetupWithManager registers the controller with the manager.
+func (r *CleanupReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
Kidswiss, line 55: Wouldn't it be easier to just reconcile the `GarageBuckets` directly? You could have a predicate to look only at those that are stuck in `Deletion` state.

Going through the helm release and fiddle with finalizers seems a bit brittle, and as copilot pointed out, could create raceconditions. Keeping the release around seems a bit pointless since we're just interested in the dangling `GarageBuckets`.

Kidswiss (changes requested):
Going through the helm release to find the `GarageBuckets` seems a bit convoluted. It might be simpler to just reconcile on `GarageBuckets` directly.

pkg/controller/garagebucket/cleanup_controller.go
TheBigLee: Refactored the whole logic completely

pkg/controller/garagebucket/cleanup_controller.go
```diff
@@ -0,0 +1,142 @@
...
+
+	garageBucket := &unstructured.Unstructured{}
+	garageBucket.SetGroupVersionKind(garageBucketGVK)
+	if err := r.Get(ctx, req.NamespacedName, garageBucket); err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
```
zugao, line 83: exit here if deletion timestamp is not set. It's a good safeguard.

zugao (commented):
Some things to check

zugao (approved):

Kidswiss (approved):
I like, it's much simpler than the previous solution

mikeshootzz (approved):

PR #647, opened 2026-04-20
Duplicate billing events sent to Odoo on status update failure

zugao:
## Summary

  - Replace productID-keyed event lookups with instanceID across all billing handlers (`create`, `delete`, `remove`, `scale`) — fixes silent skipping of second items sharing the same productID on Servala
  - Retry `persistEventSent` on transient k8s API errors (Conflict, ServerTimeout, etc.) — prevents re-delivery of already-sent events when status update fails
  - Block finalizer removal while there is an undelivered event backlog
  - Remove deprecated productID-keyed helpers from `events.go`

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1149

### DIFF
--- a/pkg/controller/billing/delete.go
+++ b/pkg/controller/billing/delete.go
@@ -5,40 +5,76 @@ import (
 	"time"
 
 	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+// handleDeletion enqueues delete events for every item that was ever sent to Odoo.
+//
+// It processes two sources to ensure no instance is missed:
+//  1. Current spec items — the normal case where all items are present.
+//  2. Sent created/scaled events in status — catches items transiently absent from spec.
+//     Crossplane applies the composed BillingService in two steps: first sets DeletionTimestamp,
+//     then may still update spec via provider-kubernetes. If the spec momentarily has N-1 items
+//     while status shows N sent events, relying on spec alone permanently misses the Nth delete.
+//     Scaled events are included to handle cases where the original create event was pruned.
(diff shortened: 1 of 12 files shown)


pkg/controller/billing/delete.go
```diff
@@ -8,35 +8,70 @@ import (
...
 
 	delTime := *billingService.DeletionTimestamp
+	seen := make(map[string]bool)
 
-	// Delete all items/products currently in spec
+	enqueue := func(it vshnv1.ItemSpec) error {
+		if seen[it.InstanceID] || hasEventByInstanceID(billingService, BillingEventTypeDeleted, it.InstanceID) {
+			return nil
+		}
+		seen[it.InstanceID] = true
```
Kidswiss: NIT: can't this be just a normal function where we pass the `seen` map?

Kidswiss (approved):

pkg/controller/billing/delete.go
zugao: I refactored it to use a map, it's easier to read indeed.

mdnix (approved):

pkg/controller/billing/remove.go
```diff
@@ -9,57 +9,40 @@ import (
...
 
-	for productID := range createdProducts {
-		if currentProducts[productID] || hasEvent(billingService, BillingEventTypeDeleted, productID) {
-			continue
-		}
-
-		info, ok := lastSent[productID]
-		if !ok {
-			// Never sent to Odoo — nothing to delete
-			b.log.Info("Skipping delete event for product never sent to Odoo",
```
TheBigLee, line 51: Nit: I'd argue, that we should keep this log output for observability and not silently continue

TheBigLee (approved):

pkg/controller/billing/remove.go
zugao: Good thing you noticed, I will add it back but in a different way as the code has changed.

mikeshootzz (approved):

PR #659, opened 2026-04-29
Add TCP Gateway to CNPG PostgreSQL

Kidswiss:
## Summary

* Adds an additional TCP Gateway implementation to the service type for CNPG

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1163

### DIFF
--- a/apis/vshn/v1/common_types.go
+++ b/apis/vshn/v1/common_types.go
@@ -211,8 +211,9 @@ type VSHNDBaaSNetworkSpec struct {
 	// Possible enum values:
 	//   - `"ClusterIP"` indicates that the service is only reachable from within the cluster.
 	//   - `"LoadBalancer"` indicates that the service is reachable from the public internet via dedicated Ipv4 address.
+	//   - `"TCPGateway"` indicates that the service is reachable via a TCP gateway (Gateway API).
 	// +kubebuilder:default="ClusterIP"
-	// +kubebuilder:validation:Enum="ClusterIP";"LoadBalancer"
+	// +kubebuilder:validation:Enum="ClusterIP";"LoadBalancer";"TCPGateway"
 	ServiceType string `json:"serviceType,omitempty"`
 }
 
--- a/crds/vshn.appcat.vshn.io_vshnkeycloaks.yaml
+++ b/crds/vshn.appcat.vshn.io_vshnkeycloaks.yaml
@@ -9449,9 +9449,11 @@ spec:
                                     Possible enum values:
                                       - `"ClusterIP"` indicates that the service is only reachable from within the cluster.
                                       - `"LoadBalancer"` indicates that the service is reachable from the public internet via dedicated Ipv4 address.
+                                      - `"TCPGateway"` indicates that the service is reachable via a TCP gateway (Gateway API).
                                   enum:
                                     - ClusterIP
                                     - LoadBalancer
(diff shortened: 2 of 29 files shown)


apis/vshn/v1/common_types.go
```diff
@@ -211,8 +211,9 @@ type VSHNDBaaSNetworkSpec struct {
 	// Possible enum values:
 	//   - `"ClusterIP"` indicates that the service is only reachable from within the cluster.
 	//   - `"LoadBalancer"` indicates that the service is reachable from the public internet via dedicated Ipv4 address.
+	//   - `"TCPGateway"` indicates that the service is reachable via a TCP gateway (Gateway API).
 	// +kubebuilder:default="ClusterIP"
-	// +kubebuilder:validation:Enum="ClusterIP";"LoadBalancer"
+	// +kubebuilder:validation:Enum="ClusterIP";"LoadBalancer";"TCPGateway"
```
mikeshootzz, line 216: Can we add a validating webhook to not allow this on clusters without a gateway implementation?

mikeshootzz (changes requested):
Was a cool review process with the plan steps :D

apis/vshn/v1/common_types.go
```diff
@@ -211,8 +211,9 @@ type VSHNDBaaSNetworkSpec struct {
...
 	ServiceType string `json:"serviceType,omitempty"`
```
mdnix, line 217: `TCPGateway` will be essentially available for all DBaaS services eventough we just wanted to have it for `VSHNPostgreSQL` for now?

apis/vshn/v1/common_types.go
mdnix: should be already guarded by `externalAccessEnabled()`

mdnix (commented):
Overall looking good, and I have tested it on a Servala cluster. 
My only concern is with the scope of the new value `TCPGateway`.

apis/vshn/v1/common_types.go
Kidswiss: > should be already guarded by `externalAccessEnabled()`

This, but I'm thinking, if we should patch this enum conditionally in the component?

externalAccessEnabled() -> we allow LoadBalancer
Gateways configured -> we allow TCPGateway

mdnix: As discussed during the call, we can leave it as is, but update documentation

zugao: If we can handle this better at component level I would go with that

Kidswiss: > If we can handle this better at component level I would go with that

I'd suggest to put that in a different work item though. We have multiple such cases where our default CRD isn't implemented everywhere. We have to think about how we want to handle this uniformly so we don't end up with multiple implementations of the same thing again.

zugao: > > If we can handle this better at component level I would go with that
> 
> I'd suggest to put that in a different work item though. We have multiple such cases where our default CRD isn't implemented everywhere. We have to think about how we want to handle this uniformly so we don't end up with multiple implementations of the same thing again.

that works for me

zugao (approved):

mdnix (approved):

mikeshootzz (approved):

PR #660, opened 2026-04-29
Allow disabling Keycloak admin interface

mikeshootzz:
## Summary

* This allows disabling the admin interface for Keycloak and setting a custom FQDN for separation from the main ingress
* Documentation: https://github.com/vshn/appcat-user-docs/pull/239

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1164


pkg/comp-functions/functions/vshnkeycloak/ingress.go
```diff
@@ -86,6 +88,64 @@ func AddIngress(_ context.Context, comp *vshnv1.VSHNKeycloak, svc *runtime.Servi
...
+// DisableAdminAccess is false.
+func buildKeycloakIngress(comp *vshnv1.VSHNKeycloak, svc *runtime.ServiceRuntime, fqdn string) (*netv1.Ingress, error) {
+	base := strings.TrimSuffix(comp.Spec.Parameters.Service.RelativePath, "/")
+
+	allowedPaths := []string{
+		base + "/realms/",
+		base + "/resources/",
+		base + "/.well-known/",
+	}
+	if !comp.Spec.Parameters.Service.DisableAdminAccess {
```
Kidswiss: Do I see this right, if the `AdminFQDN` is set, the default ingress still exposes the admin interface? So it would actually be accessible on both ingresses? Checking the work item this does not look like what the AC describes.

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnkeycloak/ingress.go
mikeshootzz: As described in the documentation, if the FQDN is set and the admin interface is not disabled, it will redirect to the ingress on which the admin interface is exposed. But yeah checking the AC, this is wrong :D I think Friday Mike got a bit confused here

zugao (approved):

Kidswiss (approved):

PR #663, opened 2026-04-30
Adjust runtime to properly support provider-kubernetes SSA

TheBigLee:
## Summary

* This PR adjust the runtime as follows:
  * Remove empty fields from objects before putting them into the desired state
  * Adjust sts-observer object to not create it with the unnecessary `spec` fields

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1168

### DIFF
--- a/pkg/comp-functions/runtime/function_mgr.go
+++ b/pkg/comp-functions/runtime/function_mgr.go
@@ -723,9 +723,43 @@ func (s *ServiceRuntime) putIntoObject(o client.Object, kon, resourceName string
 		ko.Spec.References = refs
 	}
 
-	ko.Spec.ForProvider.Manifest = runtime.RawExtension{Object: o}
+	rawBytes, err := json.Marshal(o)
+	if err != nil {
+		return nil, fmt.Errorf("cannot marshal manifest: %w", err)
+	}
+	var manifestMap map[string]any
+
+	if err := json.Unmarshal(rawBytes, &manifestMap); err != nil {
+		return nil, fmt.Errorf("cannot unmarshal manifest: %w", err)
+	}
+	pruneNilFields(manifestMap)
+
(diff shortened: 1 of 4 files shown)


mikeshootzz (approved):

pkg/comp-functions/runtime/function_mgr.go
```diff
@@ -572,6 +572,19 @@ func KubeOptionObserve(obj *xkube.Object) {
 	obj.Spec.ManagementPolicies = append(obj.Spec.ManagementPolicies, xpv1.ManagementActionObserve)
 }
 
+// KubeOptionObserveMinimalManifest removes the spec from the inner manifest, keeping only
+// the fields needed to identify the object (apiVersion, kind, metadata).
+// Use this for observer objects whose types contain non-omitempty fields (e.g. StatefulSet.spec.serviceName)
+// that would otherwise be sent via server-side apply and fail immutability validation.
+func KubeOptionObserveMinimalManifest(obj *xkube.Object) {
```
Kidswiss: I would probably add this to the normal observe. Observe only needs the metadata anyway. Any observed state lands in the status.

Kidswiss (changes requested):

zugao (approved):
LGTM

pkg/comp-functions/runtime/function_mgr.go
TheBigLee: Fixed. Please check again

Kidswiss (approved):
LGTM

I'd advise to do one last round of testing though, since it's now used more broadly.

mdnix (approved):

TheBigLee:
Tested it on lab and works without issues

PR #665, opened 2026-05-05
Also pass admin credentials to apply job

Kidswiss:
## Summary

The apply job applies ALL configuration again. Including the default one. The default one however will fallback to a default password if a specific env variable wasn't set. Those env vars were not set in the apply job, after the job ran it created the default user.

This fix adds the missing env vars to the job so it will not create the default user.

## Checklist

- [ ] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1171


zugao (approved):
We need to test this in lab

TheBigLee (approved):

mikeshootzz (approved):

PR #675, opened 2026-05-11
Allow adding additional resources to the instance namespace

mikeshootzz:
## Summary

* This allows managing additional Kubernetes resources within the instance namespace
* The feature is disabled by default for security concerns
* Extra resources have to be explicitly allowed via the component

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1184


pkg/comp-functions/functions/common/additional_resources.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}
+
+	log := controllerruntime.LoggerFrom(ctx)
+
+	if err := svc.GetObservedComposite(obj); err != nil {
+		return runtime.NewFatalResult(fmt.Errorf("cannot get composite: %w", err))
+	}
+
+	getter, ok := any(obj).(AdditionalResourcesGetter)
```
Kidswiss, line 50: I don't like this, but I don't have a better idea either...

pkg/comp-functions/functions/common/additional_resources.go
```diff
@@ -0,0 +1,134 @@
...
+		log.Info("Deployed additional resource", "key", key, "kind", u.GetKind(), "name", u.GetName())
+	}
+
+	return nil
+}
+
+// blockedAPIGroups lists API groups whose resources may not be deployed as additional resources.
+// These groups grant cluster-wide privileges, intercept API traffic, or extend the cluster API
+// in ways that would let a user escape the instance namespace or escalate privileges.
+var blockedAPIGroups = map[string]bool{
```
Kidswiss: Should we make this configurable in the component? It's just a matter of time until someone wants to deploy some RBAC to the instance namespace.

Actually thinking about this, that's already a use-case we use espejote for...

Kidswiss (changes requested):

pkg/comp-functions/functions/common/additional_resources.go
mikeshootzz: That's definitely an option. Since this utilizes provider Kubernetes, we're already limited to the RBAC it has...

zugao: For security reasons, I would do the other way around, what is allowed to install not what should be blocked.

Update: if we allow this in framework v1 what's the point of having a claim controller... this has to be very very controlled.

zugao (commented):
The idea is cool but this feature goes against the proncipal of having instance namespace as isolated space for our services. I would make it restrictive and present it as an exception not a mainstream feature.

Kidswiss (approved):

zugao (approved):
I guess you already tested it in lab.

TheBigLee (approved):

PR #682, opened 2026-05-20
Add allowed namespaces to vshngarage

mikeshootzz:
## Summary

* This uses the GarageReferenceGrant to restrict where GarageKeys and GarageBuckets can be provisioned.
* `syn-appcat` is included by default so AppCat can provision the buckets as normal

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1193

### DIFF
--- a/pkg/comp-functions/functions/vshngarage/deploy.go
+++ b/pkg/comp-functions/functions/vshngarage/deploy.go
@@ -21,25 +21,6 @@ const (
 	GarageHost = "GARAGE_URL"
 )
 
-// applyAllowedNamespaces decodes the comma-coupled JSON list provided through
-// the comp-function xfn-config (key: garageAllowedNamespaces) and injects it
-// into the vshngaragecluster chart values. Empty input leaves values
-// untouched so older component-appcat releases that don't ship the key keep
-// working — the chart simply skips the GarageReferenceGrant template.
-func applyAllowedNamespaces(values map[string]any, raw string) error {
-	if raw == "" {
-		return nil
-	}
-	var allowed []string
-	if err := json.Unmarshal([]byte(raw), &allowed); err != nil {
-		return fmt.Errorf("cannot parse garageAllowedNamespaces %q: %w", raw, err)
-	}
-	if len(allowed) > 0 {
-		values["allowedNamespaces"] = allowed
-	}
-	return nil
-}
-
 func DeployGarage(ctx context.Context, comp *vshnv1.VSHNGarage, svc *runtime.ServiceRuntime) *xfnproto.Result {
 	l := svc.Log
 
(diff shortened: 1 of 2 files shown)


pkg/comp-functions/functions/vshngarage/deploy.go
```diff
@@ -20,6 +21,25 @@ const (
 	GarageHost = "GARAGE_URL"
 )
 
+// applyAllowedNamespaces decodes the comma-coupled JSON list provided through
+// the comp-function xfn-config (key: garageAllowedNamespaces) and injects it
+// into the vshngaragecluster chart values. Empty input leaves values
+// untouched so older component-appcat releases that don't ship the key keep
+// working — the chart simply skips the GarageReferenceGrant template.
```
TheBigLee:
```suggestion
// applyAllowedNamespaces decodes the JSON array through
// the comp-functions config key (`garageAllowedNamespaces`) and injects it
// into the vshngaragecluster chart values.
// An empty input leaves values untouched, for backwards compatibility.
// the chart simply skips the GarageReferenceGrant template.
```

TheBigLee (approved):
Nit in the comment

pkg/comp-functions/functions/vshngarage/deploy.go
```diff
@@ -20,6 +21,25 @@ const (
...
+func applyAllowedNamespaces(values map[string]any, raw string) error {
```
zugao: nit: for reading purposes I would move this function at the bottom

zugao (approved):

pkg/comp-functions/functions/vshngarage/deploy.go
```diff
@@ -71,6 +91,10 @@ func DeployGarage(ctx context.Context, comp *vshnv1.VSHNGarage, svc *runtime.Ser
 		"storageMetadataSpace": comp.Spec.Parameters.Service.MetadataStorage,
 	}
 
+	if err := applyAllowedNamespaces(values, svc.Config.Data["garageAllowedNamespaces"]); err != nil {
+		return runtime.NewFatalResult(err)
```
zugao, line 76: Can this fatal error return too early and try to delete managed resources?

mikeshootzz: It doesn't, no. if we'd throw an error instead of a fatal we would have no allowed namespaces, which means no reference grant would be created. Throwing a fatal doesn't return early, it just halts the reconciliation, which in this case is what we want.

pkg/comp-functions/functions/vshngarage/deploy.go
mikeshootzz: Moved this to the bottom

PR #691, opened 2026-06-08
Make luks key secret object name deterministic

mikeshootzz:
## Summary

The provider Kubernetes object creating the Luks secret for CNPG encryption was not deterministic, thus it collided when more than one instance was created.

I've checked existing clusters and there are currently no instances that use this feature, so this won't break anything.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1204


mikeshootzz:
Tested on lab as well

TheBigLee (approved):

zugao (approved):
This will recreate the object and if there are production instances a adaptation/migration will be necessary

mikeshootzz:
> This will recreate the object and if there are production instances a adaptation/migration will be necessary

That is correct, but there are currently no instances that use this feature except for the one that broke on a test cluster

PR #698, opened 2026-06-17
Add support for forgejo runners

TheBigLee:
## Summary

* This PR adds the possibility to add runners to forgejo.
* The `runner` fields is an array that accepts a `name`, `replicas`, `size` and `labels` fields.
* When adding a runner, the runner will be automatically registered in Forgejo and the runner deploymnent will be created.
* When removing a runner, the runner will be automatically deregistered in Forgejo and the runner deployment will be removed

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1215


apis/vshn/v1/dbaas_vshn_forgejo.go
```diff
@@ -83,6 +87,34 @@ type VSHNForgejoParameters struct {
...
+	// Size controls the compute sizing of the runner pods. CPU, Memory and
+	// Plan are used; Disk is unused for runners.
+	Size VSHNSizeSpec `json:"size,omitempty"`
+
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=0
+
+	// Replicas is the fixed number of runner pods. Runners do not dynamically
+	// spawn job pods; concurrency is bounded by pre-provisioned capacity.
+	Replicas int `json:"replicas,omitempty"`
```
Kidswiss, line 109: Does setting the replicas work? AFAIK the wrenix chart mounts the `.runner` once per pod. Which might cause issue if two runner container use it at the same time.

pkg/comp-functions/functions/vshnforgejo/runners.go
```diff
@@ -0,0 +1,377 @@
...
+
+	// No runners configured: compose nothing. Previously-composed runner resources
+	// are garbage-collected; each Request's REMOVE mapping deregisters its runner.
+	if len(comp.Spec.Parameters.Runners) == 0 {
+		svc.Log.Info("No runners configured, skipping (resources will be garbage-collected)")
+		return nil
+	}
+
+	// Wait for the Forgejo instance release to be ready before registering.
+	if !forgejoReleaseReady(svc, comp) {
```
Kidswiss, line 65: The release can be ready before any pods have started... does provider-http do re-tries?

Also, if some helm upgrade fails and the release becomes unready, we'll have an early return here. I'd probably just check if the release is in the observed state.

pkg/comp-functions/functions/vshnforgejo/runners.go
```diff
@@ -0,0 +1,377 @@
...
+		// The registration Request needs the auth secret and ProviderConfig during
+		// its own deletion (the REMOVE mapping deregisters the runner). Protect them
+		// by the Request so Crossplane deletes the Request first and only then the
+		// credentials.
+		if err := svc.SetDesiredKubeObject(newRunnerAuthSecret(comp, runner, authHeader), base+runnerAuthSuffix,
+			runtime.KubeOptionAllowDeletion, runtime.KubeOptionProtectedBy(base+runnerReqSuffix)); err != nil {
+			return runtime.NewWarningResult(fmt.Sprintf("cannot add auth secret for runner %q: %s", runner.Name, err))
+		}
+
+		if err := svc.SetDesiredKubeObject(newRunnerProviderConfig(comp, runner), base+runnerPCSuffix,
```
Kidswiss, line 86: I really hate, that providerconfigs are no valid managed resource...

Kidswiss (changes requested):

apis/vshn/v1/dbaas_vshn_forgejo.go
TheBigLee: Yes it works. Both are up and running just fine as far as I could see. I haven't tested running any actual jobs though.

TheBigLee:
Closing this PR for now due to security concerns of having privileged runners on a shared infrastructure. See https://vshnticket.atlassian.net/browse/APPCAT-1373 for more information

PR #700, opened 2026-06-17
Fix probe path when relativePath is overwritten

mdnix:
## Summary

Keycloak health probes (liveness/readiness/startup) hardcoded paths to `/health/live`, `/health/ready`, and `/health`. When a custom `service.relativePath` is set (e.g. `/auth`), Keycloak serves its health endpoints under `KC_HTTP_RELATIVE_PATH`, so the probes pointed at the wrong path and failed leaving pods unready.

While testing I also found that a non-root `relativePath` is mutually exclusive with `keycloak_images_optimized`. KC_HTTP_RELATIVE_PATH is a build-time option, so under `--optimized` the path is ignored and the pod crash-loops. We now drop `--optimized` for such instances and a webhook warns about the combination.

## Checklist

- [x] Update tests.

Component PR: https://github.com/vshn/component-appcat/pull/1217

### DIFF
--- a/pkg/comp-functions/functions/vshnkeycloak/deploy.go
+++ b/pkg/comp-functions/functions/vshnkeycloak/deploy.go
@@ -1084,8 +1084,23 @@ func buildKeycloakCommand(comp *vshnv1.VSHNKeycloak, svc *runtime.ServiceRuntime
 		"--spi-events-listener-jboss-logging-success-level=info",
 		"--spi-events-listener-jboss-logging-error-level=warn",
 	}
-	if comp.Spec.Parameters.Service.CustomizationImage.Image == "" && svc.Config.Data["keycloak_images_optimized"] == "true" {
-		cmd = append(cmd, "--optimized")
+	wantOptimized := comp.Spec.Parameters.Service.CustomizationImage.Image == "" && svc.Config.Data["keycloak_images_optimized"] == "true"
+
+	if wantOptimized {
+		// http-relative-path is a Keycloak build-time option. With --optimized, Keycloak uses the
+		// image's pre-baked build (relative-path "/") and skips the startup build, so a non-root
+		// relativePath would be silently ignored and Keycloak crash-loops (the setup waits for the
+		// health endpoint under the new path, which is never served). Fall back to a runtime build
+		// (plain start) for non-root relativePath, which rebuilds and honours the value.
+		if strings.TrimSuffix(comp.Spec.Parameters.Service.RelativePath, "/") == "" {
+			cmd = append(cmd, "--optimized")
+		} else {
+			svc.Log.Info("Disabling --optimized: a non-root service.relativePath requires a runtime Keycloak build", "relativePath", comp.Spec.Parameters.Service.RelativePath)
+			svc.AddResult(runtime.NewWarningResult(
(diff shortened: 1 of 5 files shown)


pkg/comp-functions/functions/vshnkeycloak/deploy_test.go
```diff
@@ -332,6 +332,61 @@ func Test_dbcheckerImage(t *testing.T) {
...
+			wantReady:    "/health/ready",
+			wantStartup:  "/health",
+		},
+		"custom relative path": {
+			relativePath: "/auth",
+			wantLiveness: "/auth/health/live",
+			wantReady:    "/auth/health/ready",
+			wantStartup:  "/auth/health",
+		},
+		"custom relative path with trailing slash": {
```
BacLuc, line 355: thanks for that.

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -824,9 +826,10 @@ func newValues(ctx context.Context, svc *runtime.ServiceRuntime, comp *vshnv1.VS
 			},
 		},
 		// Workaround until https://github.com/codecentric/helm-charts/pull/784 is merged
```
Kidswiss, line 828: Merged dec 10, 2025 :D

Kidswiss (approved):
maybe an e2e test for this?

zugao (approved):

pkg/comp-functions/functions/vshnkeycloak/deploy.go
```diff
@@ -1081,8 +1084,23 @@ func buildKeycloakCommand(comp *vshnv1.VSHNKeycloak, svc *runtime.ServiceRuntime
 		"--spi-events-listener-jboss-logging-success-level=info",
 		"--spi-events-listener-jboss-logging-error-level=warn",
 	}
-	if comp.Spec.Parameters.Service.CustomizationImage.Image == "" && svc.Config.Data["keycloak_images_optimized"] == "true" {
-		cmd = append(cmd, "--optimized")
+	wantOptimized := comp.Spec.Parameters.Service.CustomizationImage.Image == "" && svc.Config.Data["keycloak_images_optimized"] == "true"
+
+	if wantOptimized {
+		// http-relative-path is a Keycloak build-time option. With --optimized, Keycloak uses the
```
Kidswiss, line 1090: This will make things even more complicated in the future 🙈

Kidswiss (approved):
Why is keycloak like this...?

pkg/controller/webhooks/keycloak.go
```diff
@@ -205,6 +214,66 @@ func warnPinImageTagIgnoredForCustomImage(keycloak *vshnv1.VSHNKeycloak) admissi
...
+		if step.Input == nil {
+			continue
+		}
+		input := &corev1.ConfigMap{}
+		if err := json.Unmarshal(step.Input.Raw, input); err != nil {
+			continue
+		}
+		if input.Data[keycloakImagesOptimizedInput] == "true" {
+			return true
+		}
```
Kidswiss:
```suggestion
		return input.Data[keycloakImagesOptimizedInput] == "true"
```

Kidswiss (approved):
Just one nit

mikeshootzz (approved):

PR #707, opened 2026-06-23
Allow exposing Redis via LoadBalancer or TCPGateway

mdnix:
## Summary

* Adds the ability to expose a `VSHNRedis` instance via `parameters.network.serviceType` (`LoadBalancer|TCPGateway`)
* provider-kubernetes does SSA by default in v1.2.x and owns `spec.listeners[].port`, reverting anything the webhook writes there. So instead of mutating the port, the webhook now runs the same lease-based allocation but writes the result to an annotation (`appcat.vshn.io/allocated-port`), which survives SSA. The composition function reads it back and sets the real port in the desired manifest, which provider-kubernetes then applies and owns.

## Checklist

- [x] Update tests.
- [ ] Follow [deprecation guidelines](https://github.com/vshn/appcat#deprecation-guidelines) where applicable.

Component PR: https://github.com/vshn/component-appcat/pull/1229


mikeshootzz (approved):

pkg/comp-functions/functions/vshnredis/loadbalancer.go
```diff
@@ -0,0 +1,103 @@
...
+
+	mergeChartServiceConfig(values, loadBalancerServiceValuePath(comp), serviceConfig)
+	return nil
+}
+
+func loadBalancerServiceValuePath(comp *vshnv1.VSHNRedis) []string {
+	if comp.GetInstances() > 1 {
+		return []string{"sentinel", "masterService"}
+	}
+	return []string{"sentinel", "service"}
```
zugao, line 61: isn't sentinel disabled when number of instances equals to 1? It might be under master.service.

pkg/comp-functions/functions/vshnredis/loadbalancer.go
```diff
@@ -0,0 +1,103 @@
...
+	vshnv1 "github.com/vshn/appcat/v4/apis/vshn/v1"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/functions/common"
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const loadBalancerIPConnectionDetailsField = "LOADBALANCER_IP"
+
+func externalAccessEnabled(svc *runtime.ServiceRuntime) bool {
```
zugao: Nit here: I think this was already defined in postgrescnpg package. Might be a good idea to reuse the same function.

zugao (commented):
Not sure if the thing with sentinel in the helm values is a problem, should be checked.

pkg/comp-functions/functions/vshnredis/loadbalancer.go
mdnix: fair. moving it to common

pkg/comp-functions/functions/vshnredis/loadbalancer.go
mdnix: Sentinel is actually always enabled if you look in `redis_deploy.go`. That's why for single instance I check that values path and for HA the `masterService`

pkg/comp-functions/functions/vshnredis/cert_readiness.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/vshn/appcat/v4/pkg/comp-functions/runtime"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// gateExternalCertReadiness keeps the server certificate (and thus the
+// composite) unready until the external endpoint (gateway domain or
+// loadbalancer IP) is present in the connection details and covered by the
+// server certificate's SANs. This forces Crossplane to keep reconciling so the
+// external connection details get published promptly.
+func gateExternalCertReadiness(svc *runtime.ServiceRuntime, comp *vshnv1.VSHNRedis, gatewayHost, loadbalancerIP string) error {
```
Kidswiss: I've seen this function for MariaDB as well. Might be worth putting into common as well.

Kidswiss (changes requested):

pkg/comp-functions/functions/vshnredis/cert_readiness.go
mdnix: done

Kidswiss (approved):

